PR #5, opened 2020-06-29
Improve cromwell monitoring script

TedBrookings:
* Added new cromwell monitoring script:
  /scripts/cromwell/cromwell_monitoring_script2.sh
  Outputs resource log in tabular format that 
  * uses less disk space
  * (is thus hopefully more likely to de-localize after job completion)
  * is easier to visually scan
  * is easier to ingest and process via script
* Added new script to extract summary info from new logs:
  /scripts/cromwell/get_cromwell_resource_usage2.sh 
  This script can process local or remote logs, and can select logs via
  cloud path, local path, or cromwell workflow ID

### DIFF
--- a/scripts/cromwell/get_cromwell_resource_usage2.sh
+++ b/scripts/cromwell/get_cromwell_resource_usage2.sh
@@ -1,0 +1,354 @@
+#!/bin/bash
+
+function show_help() {
+    cat <<-END
+USAGE: get_cromwell_memory_usage2.sh [OPTIONS] WORKFLOW_INFO
+Displays #tasks x #fields table of resource usage info with two
+header lines, and additional column of task descriptions.
+    WORKFLOW_INFO: specifies workflow identity. Can be
+                   a) a cromwell workflow ID
+                   b) a path to workflow output in google cloud (starting with gs://)
+                   c) a local path to workflow output
+    OPTIONS:
+        -r --raw-output
+            If set, output data as tab-separated table. Otherwise pass data through column -t for
+            easier reading.
+        -u --no-units
+            If set, don't show second header line with units.
+        -o --output-file OUTPUT_FILE_NAME
+            If specified, write output to file instead of stdout. Note that all non-tabular output is
+            sent to stderr, so this is just syntactic sugar for file redirection.
+
+-This script works by finding all the logs, sorting them into sensible
+ order, and chopping up their paths to make a description column. If
+ any jobs have not completed they will simply be omitted. The script
+ makes no attempt to figure out what tasks *should* have run. However,
+ the description field should make any such omissions discoverable.
+-If you run on a local path, the log file names must still be
+ "monitoring.log", and the local folder structure must be the same as
+ in the original cloud bucket (other non-log files are not required
+ though)
+-If you pass a workflow id, cromshell will be used to find the
+ appropriate workflow folder in google cloud storage.
+-If cromshell is not located on your PATH, then define the ENV
+ variable CROMSHELL with the appropriate path.
+-Since there is significant start-up time to running gsutil functions,
+ running the inner loop of this script in parallel results in
+ signficant speed-up. Installing gnu parallel (available on osx and
+ linux) triggers this automatically.
+END
+}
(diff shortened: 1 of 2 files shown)


scripts/cromwell/get_cromwell_resource_usage2.sh
```diff
@@ -0,0 +1,339 @@
...
+        | tail -n1 \
+    )
+fi
+1>&2 echo "WORKFLOW_DIR=$WORKFLOW_DIR"
+
+function get_monitor_logs() {
+    TOP_DIR=$1
+    REMOTE=$2
+    if $REMOTE; then
+      gsutil -m ls "$TOP_DIR/**monitoring.log" 2>/dev/null || echo ""
```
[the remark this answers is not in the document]
TedBrookings: I added code to remove trailing slashes from the input WORKFLOW_INFO when it's a cloud or local path. I tested it both with and without extra slashes and it appeared to work.

The other error you mentioned was: if you supplied options that bypasses the (no input arguments) check at the beginning of the script. If you forgot to supply WORKFLOW_INFO, it default to empty, isn't a path, and gets interpreted as an empty WORKFLOW_ID. Cromshell gets the last workflow when you pass it an empty workflow ID. I addressed this by checking for an empty WORKFLOW_INFO and throwing an error. I also moved around some (trivial) output to stderr to make errors more clear.

TedBrookings:
@epiercehoffman, following our offline conversation, I added an option to omit the second "units" header line, and I altered the monitoring script to make the header column names unique. I'm happy to wait until you have a chance to test these changes (I did but your test routine is doubtlessly different from mine). If you want me to just go ahead and merge, I can do that too.

scripts/cromwell/get_cromwell_resource_usage2.sh
```diff
@@ -50,6 +53,8 @@ for ((i=1; i<=$#; ++i)); do
         exit 0
     elif [[ ${!i} =~ ^-+(r|raw-output) ]]; then
         RAW_OUTPUT=true
+    elif [[ ${!i} =~ ^-+(u|hide-units) ]]; then
```
epiercehoffman, line 56:
```suggestion
    elif [[ ${!i} =~ ^-+(u|no-units) ]]; then
```

scripts/cromwell/get_cromwell_resource_usage2.sh
```diff
@@ -5,17 +5,19 @@ function show_help() {
...
-      sent to stderr, so this is just syntactic sugar for file redirection.
+    WORKFLOW_INFO: specifies workflow identity. Can be
+                   a) a cromwell workflow ID
+                   b) a path to workflow output in google cloud (starting with gs://)
+                   c) a local path to workflow output
+    OPTIONS:
+        -r --raw-output
+            If set, output data as tab-separated table. Otherwise pass data through column -t for
+            easier reading.
+        -u --hide-units
```
epiercehoffman, line 16:
```suggestion
        -u --no-units
```

epiercehoffman (changes requested):
I am so sorry it took me this long to review your changes again, especially since you responded to my last review so quickly! Thanks so much for adding the option to hide the second units header line and for making the first header line have unique column names - that will make parsing the output of this script in python simpler for sure. 

I tested a whole bunch of combinations of arguments and had no trouble until `--hide-units`, which gets matched by the `^-+(h|help)` regex before it gets to `^-+(u|hide-units)`. This is certainly not a big deal as `-u` works perfectly, but for the aesthetic I have suggested changing the longer-form option to hide the units to `--no-units`, as that seemed to be the easiest way to deal with this problem. If you are satisfied with that change, then I think this is all set to merge - I tested the `--no-units` change already locally so I don't feel the need to do another round of testing.

scripts/cromwell/get_cromwell_resource_usage2.sh
TedBrookings: Good catch, and that's a good fix.

epiercehoffman (approved):

PR #6, opened 2020-06-30
Fix running without any MELT inputs

cwhelan:
This fixes some points in the WDL which were failing when run with `use_melt` set to false and no melt inputs provided for the reference panel.

Also added an input JSON file to demonstrate how to run without MELT, which will likely be the most frequent use case when the pipeline is deployed to Terra.


wdl/GATKSVPipelineSingleSample.wdl
```diff
@@ -628,30 +631,47 @@ workflow GATKSVPipelineSingleSample {
...
+    call SingleSampleFiltering.FilterVcfBySampleGenotypeAndAddEvidenceAnnotation as FilterMelt {
+        input :
+            vcf_gz=select_first([Module01.melt_vcf]),
+            sample_id=sample_id,
+            evidence="RD,PE,SR",
+            sv_base_mini_docker=sv_base_mini_docker,
+            runtime_attr_override=runtime_attr_filter_vcf_by_id
+    }
+  }
+  if (use_melt) {
```
TedBrookings, line 664:
```suggestion
  if (use_delly) {
```
Typo?

cwhelan: Whoops, thanks for pointing that out!

cwhelan:
At this point I don't think we envision turning off manta or wham inputs, so I'm inclined not to add extra complexity around handling that, until it becomes a thing.

PR #14, opened 2020-07-01
Remove sample_list output from Module00c

cwhelan:
The cromwell servers used by Terra don't have local filesystems enabled, so they error out if `write_lines` is called outside of a task context. This changes Module00c to spin up a task to write out the sample list, and writes out `all_samples` instead of `samples`, which would include the case sample in the single sample pipeline.


cwhelan:
After discussing with @mwalker174 It sounds like the sample_list output wasn't really needed anyway, so I've revised this to just remove it as an output.

PR #20, opened 2020-07-08
Allow docker build with deprecated XML package

TedBrookings:
* Added install_deprected_R_package.sh to download and install R packages
  that are no longer supported in online repos.
* Updated sv-base/Dockerfile to install this script and include curl
  dependency.
* Updated sv-pipeline-base/Dockerfile to use this script to install XML
  R module

### DIFF
--- a/dockerfiles/sv-base/Dockerfile
+++ b/dockerfiles/sv-base/Dockerfile
@@ -13,7 +13,7 @@ ENV HOME=/root
 # G-Cloud
 ARG CLOUD_SDK_VERSION=235.0.0
 ARG GCLOUD_SDK_EXTRA_COMPONENTS
-ARG GCLOUD_SDK_TRANSIENT_DEP="curl gpg-agent gnupg python-pip python-setuptools "
+ARG GCLOUD_SDK_TRANSIENT_DEP="gpg-agent gnupg python-pip python-setuptools "
 ARG GCLOUD_SDK_DEPENDENCIES="lsb-release ${GCLOUD_SDK_TRANSIENT_DEP}"
 ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get -qqy update --fix-missing && \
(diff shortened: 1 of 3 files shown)


dockerfiles/sv-base/Dockerfile
```diff
@@ -53,8 +53,12 @@ RUN apt-get -qqy update --fix-missing && \
 #  4. final clean up
 COPY install_R_packages.R /opt/
 COPY install_bioconductor_packages.R /opt/
+COPY install_deprecated_R_package.sh /opt/
 ARG R_DEP_TRANSIENT="make gpg-agent gnupg"
-ARG R_DEPENDENCIES="software-properties-common build-essential ${R_DEP_TRANSIENT} libz-dev libncurses5-dev libbz2-dev liblzma-dev libcurl4-openssl-dev libssl-dev libxml2-dev ca-certificates apt-transport-https openssh-client"
+ARG R_DEPENDENCIES="software-properties-common build-essential \
+    ${R_DEP_TRANSIENT} libz-dev libncurses5-dev libbz2-dev liblzma-dev \
+    libcurl4-openssl-dev libssl-dev libxml2-dev ca-certificates \
+    apt-transport-https openssh-client curl"
```
cwhelan, line 61: It looks like you added `curl` here, right? It looks like curl is already installed in `sv-base-mini`. However, it looks like it's getting listed as part of `GCLOUD_SDK_TRANSIENT_DEP` on line 16 of this dockerfile. I think that if you just removed it from that list, it would not get purged on line 36, and then would be available in this docker image (although I haven't tested that). That would avoid downloading and installing it twice.

dockerfiles/sv-base/install_deprecated_R_package.sh
```diff
@@ -0,0 +1,11 @@
...
+set -eu -o pipefail
+
+MODULE_ARCHIVE_URL=$1
+
+ARCHIVE_DIR=$(mktemp -d "${TMPDIR:-/tmp/}$(basename $0).XXXXXXXXXXXX")
+trap "rm -rf $ARCHIVE_DIR" EXIT
+
+MODULE_ARCHIVE_DEST="$ARCHIVE_DIR/$(basename "$MODULE_ARCHIVE_URL")"
+curl "$MODULE_ARCHIVE_URL" --output "$MODULE_ARCHIVE_DEST"
+Rscript -e "install.packages('$MODULE_ARCHIVE_URL', repos = NULL, quiet = TRUE)"
```
cwhelan, line 11: Do you want the first parameter to be `$MODULE_ARCHIVE_DEST` instead of `$MODULE_ARCHIVE_URL`? I think this will make R re-download it from the URL, won't it?

TedBrookings: Chris, I made more typos in this script than it has lines, by like a factor of two. And yet, here's another...
Good catch, I fixed it.

dockerfiles/sv-base/Dockerfile
TedBrookings: I did add curl.
I don't see the curl package in sv-base-mini (it looks like it install libcurl but not curl itself).
The other curl like you said is transient because it's needed temporarily for gcloud but not long-term. I was reluctant to make that a permanent dependency there because it might not be clear later on why it's permanent there. And curl is small. However it is a wasteful extra download, so if you think it's best I can make that change. Let me know what you think.

cwhelan: I think curl is installed in the sv-base-mini dockerfile on line 61. It's available if you run an sv-base-mini image.

To my eye sv-base-mini is trying to install it permanently and then sv-base comes along and wipes it out unintentionally, which is why i suggested just removing it from the transient list here.

TedBrookings: You're right. My editor makes that block of the sv-base-mini dockerfile look like a bunch of comments, so I didn't check it carefully. I installed the dockerfile plugin and made the fix you suggested.

PR #22, opened 2020-07-10
Update WHAM TAGS INFO field with renamed sample ID

cwhelan:
Recent changes to re-write sample IDs missed the fact that WHAM encodes the sample name in the TAGS INFO field. This caused the VCF standardization script for WHAM files to exclude all variants, leaving the files empty. This change ensures that the TAGS field is re-written with the specified sample ID in addition to the VCF header.

Tested by running the single sample pipeline on NA12878 with different sample identifier.

### DIFF
--- a/wdl/Whamg.wdl
+++ b/wdl/Whamg.wdl
@@ -173,16 +173,25 @@ task RunWhamg {
         ln -s "~{reference_index_file}" "~{reference_fasta}.fai"
     fi
         
-    echo "Invoking whamg"
+    # We need to update both the VCF sample ID and the TAGS INFO field in the WHAM output VCFs.
+    # WHAM uses both to store the sample identifier, and by default uses the SM identifier from the BAM file.
+    # We need to update both to reflect the potentially-renamed sample identifier used by the pipeline (sample_id) --
+    # svtk standardize_vcf uses the TAGS field to identify the sample for WHAM VCFs.
+
     VCF_FILE_FIXED_HEADER="~{sample_id}.wham.vcf.gz"
     VCF_FILE_BAD_TAGS="~{sample_id}.wham_bad_header_bad_tags.vcf.gz"
     VCF_FILE_BAD_HEADER="~{sample_id}.wham_bad_header.vcf.gz"
+
+    echo "Invoking whamg"
+
     whamg -c "~{sep=","  chr_list}" -x ~{num_cpu} -a ~{reference_fasta} -f ~{bam_file} \
       | bgzip -c > "$VCF_FILE_BAD_TAGS"
-
     tabix -p vcf "$VCF_FILE_BAD_TAGS"
+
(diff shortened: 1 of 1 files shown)


LNGDingj:
@cwhelan  What should to do with results generated by the Whamg.wdl in gatk-sv-0.8.2-beta? Thanks a lot!

cwhelan:
Hi @LNGDingj -- have you been trying to run the single sample pipeline (`GATKSVPipelineSingleSample.wdl`)?

First I'd verify that this bug was affecting you. I believe that it would only affect you if you were running the pipeline with a different sample identifier input (`sample_id`) than the sample ID that is present in the bam or cram file you are using in the `@SM` header. The issue that this PR was trying to fix resulted in standardized wham VCF files that are returned from Module00c in the `std_wham_vcf` output but contain no records. This was also causing downstream fatal errors in the `SRTest` subworkflow in some samples.

If you think you might be affected, I'd probably recommend re-running the workflow with the `v0.8.3-beta` release. If you've already run on lots of data, I could try to walk through other solutions for recovering your data.

LNGDingj:
Hi, @cwhelan -- Thanks so much for your quick reply. I already used Whamg.wdl through Module00aBatch, and generated results of over 8000 samples at GCP using same sample IDs as those presenting in .cram and .bam files. And plan to continue to run Module00b and following Modules in the gatk-sv pipeline. Thanks again!

cwhelan:
@LNGDingj  If you'd like to double-check that your files are OK, you can make sure that the TAGS info field within your wham output VCFs matches the sample ID listed on the header line of the VCF file and the expected sample_id that you gave as an input for the workflow. If you do run into issues let me know (file a ticket on this repo, probably) and we could help you figure out how to fix the raw wham output files if necessary.

LNGDingj:
@cwhelan I did not change sample IDs (same as in .cram and .bam), when I was using cohort mode to run Module00aBatch.wdl, I noticed that in README.md it states  "Currently, sample IDs can be replaced again in Module 00c". May I change sample IDs for those with "-" in Module 00c? Thanks so much!

mwalker174:
@LNGDingj You should be able to proceed to 00b. However, I would wait on 00c until we resolve this issue: https://github.com/broadinstitute/gatk-sv/issues/23

LNGDingj:
@mwalker174 Sure, thanks a lot!

LNGDingj:
@mwalker174 
Hello, 
Following your reply on Jul 10, I noticed that issues #26 (Replace remote tabix with GATK) was closed yesterday, so it is ready to proceed to Module00cBatch with all scripts currently deposited at broadinstitute/gatk-sv on github?
Regarding to "Currently, sample IDs can be replaced again in Module00c", in addition to replacing all "-" in sample IDs with "_" for "Module00c.samples" in input json file, what else should I do? 
Thanks a lot!

Issue #23, opened 2020-07-10
Tabix localization fails silently in some regions


mwalker174:
There is currently an issue where retrieval of PE/SR/RD records from GCS using `tabix` can fail silently in some regions, resulting in loss of evidence for some variants. This affects modules 00c (CNMOPs), 02, 04, and 05/06.

We have been [corresponding with the developer](https://github.com/samtools/htslib/issues/1037) but not all issues could be resolved. We are working on an alternative solution.

mwalker174:
Fixed in #26

PR #26, opened 2020-07-16
Replace remote tabix with GATK

mwalker174:
Addresses issue #23 (mostly) using a new GATK tool `LocalizeSVEvidence`, which is currently in a dev branch in the GATK repo. 

Although we haven't observed record loss for retrieving VCF records with tabix, it will be safer to use GATK `SelectVariants` with streaming from GCS. This is implemented (but untested) for module07 annotation. I attempted to use this tool in Module0506, but GATK stripped out empty fields in all records, which caused major issues in the CleanVcf subworkflow. I fixed one of the bugs but ultimately decided to revert to using tabix, as the bugs further occurred (also related to GATK stripping empty FORMAT fields) and this PR is critical for current gnomAD processing.

A current GATK jar (from a branch off ~v1.8.0) is also now available in most pipeline dockers (except mini).

Summary:

- Use GATK for PE/SR/RD localization instead of tabix
- Add GATK to sv-base docker and update dockers
- Fix fragile cleanvcf1 code block that assumed that the `EV` format field came last


src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part1.sh
```diff
@@ -27,18 +27,27 @@ zcat $vcf \
...
+cut -f1-4 ev.tab > ev.sites.tab
+cut -f5- ev.tab \
+  | sed -e 's/7/RD,PE,SR/g' -e 's/6/PE,SR/g' -e 's/5/RD,SR/g' -e 's/3/RD,PE/g' -e 's/2/PE/g' -e 's/4/SR/g' -e 's/1/RD/g' \
+  > ev.replaced.list
+paste ev.sites.tab ev.replaced.list | bgzip -c > ev.replaced.tab.gz
+tabix -s1 -b2 -e2 ev.replaced.tab.gz
+
+echo '##FORMAT=<ID=EV,Number=1,Type=String,Description="Classes of evidence supporting final genotype">' > ev_header.txt
+tabix $vcf
+bcftools annotate -x FORMAT/EV $vcf | bgzip -c > no_ev.vcf.gz
```
cwhelan, line 40: Do you need to remove EV here first? I thought the second annotate command would just replace the value.

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part1.sh
```diff
@@ -27,18 +27,27 @@ zcat $vcf \
...
+rm $vcf $vcf.tbi
+
+bcftools annotate \
+  -a ev.replaced.tab.gz \
+  -c CHROM,POS,REF,ALT,FORMAT/EV \
+  -h ev_header.txt \
+  no_ev.vcf.gz \
+  | bgzip -c > EV.update.vcf.gz
+
+rm ev.tab no_ev.vcf.gz ev.replaced.tab.gz ev.replaced.tab.gz.tbi ev.replaced.list
```
cwhelan, line 50: Is there a reason to remove these files? Just curious on whether you ran into some issue based on leaving them around.

wdl/AnnotateVcf.wdl
```diff
@@ -97,36 +97,52 @@ task SubsetVcf {
...
+  command <<<
+
+    set -euo pipefail
+
+    java -Xmx~{java_mem_mb}M -jar ${GATK_JAR} SelectVariants \
+      -V "~{vcf}" \
+      -L "~{contig}" \
+      -O ~{prefix}.~{contig}.vcf
+
+    # GATK does not block compress
```
cwhelan, line 137: It should be block compressing -- is there a bug? I thought it would be setting up the output stream at `VariantContextWriterBuilder.determineOutputTypeFromFile`

wdl/BAFTestChromosome.wdl
```diff
@@ -114,28 +117,22 @@ task BAFTest {
...
-    cmp --silent local_baf_1.bed.gz local_baf_2.bed.gz || exit 1
-
-    mv local_baf_1.bed.gz local_baf.bed.gz
+    java -Xmx~{java_mem_mb}M -jar ${GATK_JAR} LocalizeSVEvidence \
+      --sequence-dictionary ~{ref_dict} \
+      --evidence-file ~{baf_metrics} \
+      -L "${chrom}:${start}-${end}" \
+      -O local_baf.bed
+
+    # GATK does not block compress
```
cwhelan, line 126: Again, I think it should be able to -- you might want to let the engine team know if you're having trouble with that.

wdl/MatrixQC.wdl
```diff
@@ -147,16 +153,27 @@ task PESRBAF_QC {
...
+
+    java -Xmx~{java_mem_mb}M -jar ${GATK_JAR} LocalizeSVEvidence \
+      --sequence-dictionary ~{ref_dict} \
+      --evidence-file ~{matrix_file} \
+      -L regions.bed \
+      -O local_matrix.txt
+
+    # GATK does not block compress
+    bgzip local_matrix.txt
+    tabix -s 1 -b 2 -e 2 local_matrix.txt.gz
```
cwhelan, line 175: Is this a bed file? Should it be `-e 3` (or `-p bed`)?

wdl/ResolveCpxSv.wdl
```diff
@@ -329,30 +336,36 @@ task ResolvePrep {
...
       while read GS_PATH_TO_DISC_FILE; do
         ((++DISC_FILE_NUM))
         SLICE="disc"$DISC_FILE_NUM"shard"
 
-        tabix -R regions_to_tabix.bed "$GS_PATH_TO_DISC_FILE" \
+        java -Xmx~{java_mem_mb}M -jar ${GATK_JAR} LocalizeSVEvidence \
+              --sequence-dictionary ~{ref_dict} \
+              --evidence-file $GS_PATH_TO_DISC_FILE \
+              -L regions.bed \
+              -O ${SLICE}.txt
```
cwhelan, line 362: If you wanted to, could you change this to `-O /dev/stdout` and keep the pipe as it was?

wdl/Tasks04.wdl
```diff
@@ -31,14 +31,14 @@ task SplitVariants {
...
+      | split --additional-suffix ".bed" -l ~{n_per_split} -a 6 - gt5kb.
     svtk vcf2bed ~{vcf} stdout \
       | awk -v OFS="\t" '(($5=="DEL" || $5=="DUP") && $3-$2<5000) {print $1, $2, $3, $4, $6, $5}' \
-      | split -l ~{n_per_split} -a 6 - lt5kb.
+      | split --additional-suffix ".bed" -l ~{n_per_split} -a 6 - lt5kb.
     if [ ~{generate_bca} == "true" ]; then
       svtk vcf2bed ~{vcf} stdout \
-        | awk -v OFS="\t" '($5!="DEL" && $5!="DUP") {print $1, $2, $3, $4, $6, $5}' \
-        | split -l ~{n_per_split} -a 6 - bca.
+        | awk -v OFS="\t" '($5!="DEL" && $5!="DUP") {print $1, $2RDTestGenotype, $3, $4, $6, $5}' \
```
cwhelan, line 40: Is the `$2RDTestGenotype` supposed to be there? Looks like it could be a typo?

wdl/Tasks04.wdl
```diff
@@ -394,20 +414,28 @@ task CountPE {
 
     set -euo pipefail
     svtk vcf2bed --split-bnd --no-header ~{vcf} test.bed
-    awk -v OFS="\t" -v window=5000 '{if ($2-window>0){print $1,$2-window,$2+window}else{print $1,0,$2+window}}' test.bed  >> region.bed
-    awk -v OFS="\t" -v window=5000 '{if ($3-window>0){print $1,$3-window,$3+window}else{print $1,0,$3+window}}' test.bed  >> region.bed
+    awk -v OFS="\t" -v window=500 '{if ($2-window>0){print $1,$2-window,$2+window}else{print $1,0,$2+window}}' test.bed  > region.bed
```
cwhelan, line 417: Why change this from 5000 to 500? Should it be a parameter for the task?

wdl/VcfClusterSingleChromsome.wdl
```diff
@@ -164,10 +168,37 @@ task JoinContigFromRemoteVcfs {
   
   command <<<
     set -eu -o pipefail
-    
+
+    ## This doesn't quite work (but almost!) as a replacement for tabix below
```
cwhelan, line 172: It might be good to make an ticket to track these issues. Not sure if you think they should be fixed in GATK or not, but if so it might be good to make an issue there as well.

cwhelan: Or if there's nothing to be fixed I might replace this chuck of commented code with a shorter comment explaining that you tried to do this with GATK and what the issues were.

dockerfiles/sv-base/Dockerfile
```diff
@@ -7,6 +7,31 @@ FROM ${MINIBASE_IMAGE}
...
 ENV HOME=/root
 
+####################################################################################################
+# GATK - only used for LocalizeSVEvidence
+ARG GATK_TRANSIENT_DEP="git git-lfs"
+ARG GATK_DEP="openjdk-8-jdk ${GATK_TRANSIENT_DEP}"
+ARG GATK_COMMIT="e7e761358fc06880729e9a9ed1826c787a6efc19"
+ENV GATK_JAR="/opt/gatk.jar"
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get -qqy update --fix-missing && \
```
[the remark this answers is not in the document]
mwalker174: I meant to do this originally, added `rm` for the source dir

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part1.sh
mwalker174: No just cleanup. I expect this could be worthwhile for gnomad.

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part1.sh
mwalker174: Good question, I had to go back and double-check this. The problem is that in the original vcf, EV is type Integer, but this seems to fail below since we are assigning Strings. The only solution was to delete EV entirely.

wdl/AnnotateVcf.wdl
mwalker174: Apparently it depends on the way the output stream is written. I've fixed the tool so that it writes block compressed files now.

wdl/MatrixQC.wdl
mwalker174: Only bincov is bed type, the others do not have a proper END column

wdl/ResolveCpxSv.wdl
mwalker174: Hm that makes me nervous in case a log message or something gets sent to stdout. I'd rather keep it this way to be safe.

wdl/Tasks04.wdl
mwalker174: Yes must have accidentally pasted that in somehow

wdl/VcfClusterSingleChromsome.wdl
mwalker174: Thanks I had forgotten this was here. I've replaced it with a reference to #52.

mwalker174:
Incorporated into #51

PR #30, opened 2020-07-20
Fixes for Module03Qc.wdl: optional inputs, no .tbi required, no GQ plots when GQ missing, etc

epiercehoffman:
Various edits to make Module03Qc.wdl work and to make sure those edits don't break Module05_06.wdl, which calls the same MasterVcfQc.wdl workflow:
* Make input VCFs to Module03Qc optional, to be able to mix and match callers and run with regular testing inputs, which don't include all callers
* Make inputs to sub-workflows called by Module03Qc top-level inputs to Module03Qc, for ease & convenience of providing inputs
* Make it possible to do Tasks0506.SplitVcf without a tabix index: make tabix index optional input to MasterVcfQc.wdl & pass through ShardedQcCollection.wdl to Tasks0506.wdl (so that Module03Qc can forego index and Module05_06 can use index), specify index in Module05_06, fix outputted file in case that index is not provided & contig is provided
* Only make per-family QC plots with GQ if GQ data exists, so that plotting functions don't crash during Module03Qc when there is no GQ (`analyze_fams.R`)
* Update `sv-pipeline-qc` docker image in all test input jsons to image with new version of `analyze_fams.R`
* Add `/test/Module03Qc.test_large.json` with test inputs to Module03Qc.wdl

Important files edited (aka anything but the many `sv-pipeline-qc` docker image updates in the test input jsons):
* /wdl/Module03Qc.wdl
* /wdl/Module05_06.wdl
* /wdl/MasterVcfQc.wdl
* /wdl/ShardedQcCollection.wdl
* /wdl/Tasks0506.wdl
* /src/sv-pipeline/scripts/vcf_qc/analyze_fams.R
* /test/Module03Qc.test_large.json

Testing performed:
* Module03Qc ran successfully & output plots looked good, with no GQ plots (cromshell ID: `17da0053-4a3c-443b-9dc1-c571e293ca52`)
* Module05_06 ran successfully (cromshell ID: `50cf309a-8137-409d-b4c5-9eac25060e26`)
* All input jsons & corresponding wdls passed input validation with `/scripts/test/validate.sh` 
* Note that a few other workflows besides the two tested call Tasks0506.SplitVcf, but without a tabix index (which was already an optional input), so it is assumed that no changes to these wdls are necessary and that if Module03Qc works, these tasks will as well. If anything, they should be more efficient because of the changes made. Please advise if additional testing of these workflows is recommended. They are: CleanVcf, ScatterCpxGenotyping

### DIFF
--- a/src/sv-pipeline/scripts/vcf_qc/analyze_fams.R
+++ b/src/sv-pipeline/scripts/vcf_qc/analyze_fams.R
@@ -1034,7 +1034,7 @@ wrapperInheritancePlots <- function(fam.dat.list,fam.type,count="variants"){
   dev.off()
 }
 #Wrapper for de novo rate lineplots
-wrapperDeNovoRateLines <- function(fam.dat.list,fam.type,count="variants"){
+wrapperDeNovoRateLines <- function(fam.dat.list,fam.type,count="variants",gq=F){
   #Set title prefix
   if(count=="variants"){
     title.prefix <- "SV Site "
@@ -1061,14 +1061,15 @@ wrapperDeNovoRateLines <- function(fam.dat.list,fam.type,count="variants"){
   dev.off()
   
   #DNR by Proband GQ
-  GQ.dat <- deNovoRateByProGQ(trio.dat.list=fam.dat.list,GQ.bins=40,count=count)
-  pdf(paste(OUTDIR,"/supporting_plots/sv_inheritance_plots/sv_de_novo_rate.",fam.type,"s.",count,".by_proband_GQ.pdf",sep=""),
-      height=4,width=5)
-  plotDNRvsGQ(DNRs=GQ.dat$DNRs,bins=GQ.dat$bins,k=4,nfams=length(fam.dat.list),
-              title=paste(title.prefix,"De Novo Rate by Min. Proband GQ",sep=""),
-              count=count,fam.type=fam.type,legend=T,xlab="Min. Proband GQ")
-  dev.off()
-  
+  if(gq) {
+    GQ.dat <- deNovoRateByProGQ(trio.dat.list=fam.dat.list,GQ.bins=40,count=count)
+    pdf(paste(OUTDIR,"/supporting_plots/sv_inheritance_plots/sv_de_novo_rate.",fam.type,"s.",count,".by_proband_GQ.pdf",sep=""),
+        height=4,width=5)
+    plotDNRvsGQ(DNRs=GQ.dat$DNRs,bins=GQ.dat$bins,k=4,nfams=length(fam.dat.list),
+                title=paste(title.prefix,"De Novo Rate by Min. Proband GQ",sep=""),
+                count=count,fam.type=fam.type,legend=T,xlab="Min. Proband GQ")
+    dev.off()
+  }
 }
 #Wrapper for de novo rate heatmaps
 wrapperDeNovoRateHeats <- function(fam.dat.list,fam.type,count="variants"){
@@ -1105,7 +1106,7 @@ wrapperDeNovoRateHeats <- function(fam.dat.list,fam.type,count="variants"){
   })
 }
 #Wrapper for master summary panel
-masterInhWrapper <- function(fam.dat.list,fam.type){
+masterInhWrapper <- function(fam.dat.list,fam.type, gq=T){
   #Set title suffix
   if(fam.type=="trio"){
     title.suffix <- paste("(Trios; n=",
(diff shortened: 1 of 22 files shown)


wdl/MasterVcfQc.wdl
```diff
@@ -34,6 +34,7 @@ workflow MasterVcfQc {
     Array[File]? custom_comparison_sets
     Array[String] contigs
     Int? random_seed
+    File? vcf_idx
```
cwhelan, line 37: I'm not sure if you discussed this with @mwalker174 or not, but we recently removed vcf indices as explicit inputs to subworkflows in the pipeline, see https://github.com/talkowski-lab/gatk-sv-v1/pull/252.

Is the reason for adding it back the fact that this module is sometimes called without an index file having been created? Is there any chance we could just remedy that situation rather than reverting to passing the index in this particular case? In my opinion we should be creating indices alongside vcf files consistently through the pipeline.

src/sv-pipeline/scripts/vcf_qc/analyze_fams.R
```diff
@@ -1331,8 +1344,16 @@ if(nrow(trios)>0){
   })
   names(trio.dat) <- trios[,1]
   
+  # if there are no GQ values, do not make GQ plots
+  gq <- F
```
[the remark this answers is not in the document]
epiercehoffman: You interpreted correctly and that one-liner does look nicer!

wdl/MasterVcfQc.wdl
epiercehoffman: The vcf cannot be tabix-indexed at this point in the pipeline because it is out of spec - the END field is used anomalously for BNDs so END < START sometimes, and that is not fixed until the very end of the pipeline, and Mark felt it was easier to do a workaround here rather than change everything in between to get it to comply with vcf spec. That means MasterVcfQc.wdl needs to be called with an index for Module05_06 and without for Module03Qc, which motivated this change. There is still no need to provide a VCF index file path on the top-level WDL though. But also, if you have suggestions for a better way to accomplish this I would welcome that - I originally wanted to just check if the expected .tbi file existed but could not figure out how to do that in WDL.

cwhelan: Oh, sorry, you mentioned the END thing before and I forgot.

One potential option to check if the tbi file exists might be to use the WDL `size()` function on it. It looks like if you pass it a `File?` and the file doesn't exist it should return 0:

https://github.com/openwdl/wdl/blob/main/versions/1.0/SPEC.md#float-sizefile-string

I haven't tested that myself, though, and it's a little messy looking.

epiercehoffman: It looks to me like the behavior if the file is provided as an argument to `size()` but does not exist is not specified in the docs - I read it as returning 0 when no File argument is provided. Since it's a bit unclear and the current implementation still avoids having to provide `vcf_idx` on the top level WDL, I'm inclined to leave this as-is - but could also test out `size()` and see what happens if you think that's a better way

wdl/Module03Qc.wdl
```diff
@@ -25,89 +25,115 @@ workflow Module03Qc {
...
-      prefix="${batch}.depth_03_filtered_vcf",
-      sv_per_shard=10000,
-      samples_per_shard=100,
-      thousand_genomes_tarballs=thousand_genomes_tarballs,
-      hgsv_tarballs=hgsv_tarballs,
-      asc_tarballs=asc_tarballs,
-      sanders_2015_tarball=sanders_2015_tarball,
-      collins_2017_tarball=collins_2017_tarball,
-      werling_2018_tarball=werling_2018_tarball
+  Array[String] algorithms = ["manta", "delly", "melt", "depth", "pesr"]
```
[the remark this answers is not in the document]
epiercehoffman: Good catch! I will add wham. I had taken it for granted that it was written without wham for a reason but Mark is not aware of one. For sake of transparency, though, there is no Module03 wham output saved in the test outputs directory, so I won't be able to test it for now. That also applies to everything except pesr and depth though - I just used what was available for testing.

test/module03/Module03Qc.test_large.json
```diff
@@ -0,0 +1,43 @@
+{
+  "Module03Qc.contigs" : [
+    "chr1","chr2","chr3","chr4","chr5","chr6","chr7","chr8","chr9","chr10","chr11","chr12","chr13","chr14","chr15","chr16","chr17","chr18","chr19","chr20","chr21","chr22","chrX","chrY"
```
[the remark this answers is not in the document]
epiercehoffman: Very true, will modify to follow the other WDLs & take a contigs file

wdl/MasterVcfQc.wdl
cwhelan: That's fine, it's not a big deal either way.

epiercehoffman:
Made requested changes (convert to one-liner in analyze_fams.R, add wham, make Module03Qc.wdl take contigs list file), tested Module03Qc.wdl again (cromshell ID: 507093a1-9fc5-4644-a48c-9a73510f314b), and re-validated WDLs & input JSONs. 

I updated the docker image again because of the change to analyze_fams.R, so there are a lot of files in the most recent commit, but the most important changes are:
* analyze_fams.R
* Module03Qc.wdl
* Module03Qc.test_large.json

PR #40, opened 2020-08-21
Annotate single sample output with gnomAD allele frequencies and add BED output

cwhelan:
This PR enables annotation of the single sample pipeline's output with gnomAD allele frequencies.

It modifies`module07.wdl` to optionally annotate the vcf with allele frequencies from an external resource if the right parameters are supplied. 

In addition, a version of the single sample pipeline's output is supplied in BED format in addition to VCF.


cwhelan:
I didn't update the `sv-pipeline` docker for this one yet, either -- figured that could be batched with https://github.com/broadinstitute/gatk-sv/pull/36 (and https://github.com/broadinstitute/gatk-sv/pull/26 if it's going to be ready shortly)

mwalker174 (approved):
Sorry I could have sworn I had approved this a while back. Looks good to me, as long as the Docker images get updated.

Issue #48, opened 2020-09-10
CollectCounts task ignores bam index parameter


cwhelan:
If the bam index name can't be inferred from the bam name, CollectCounts fails with the following error:

`A USER ERROR has occurred: Traversal by intervals was requested but some input files are not indexed.`

I believe that we need to add the `--read-index` parameter to our invocation of GATK CollectReadCounts.

This issue may affect other tasks.

cwhelan:
Closed via https://github.com/broadinstitute/gatk-sv/pull/54

Issue #11, opened 2020-06-30
Manta WDL can break with unexpected cram/bam index extensions


mwalker174:
For example, Manta's configuration script always expects the cram index to be `<cram_path>.crai`, ie ending in `.cram.crai`. Currently, one can pass non-standard index paths, such as in a different directory or lacking `.cram` in the file name. This causes Manta to crash.

This bug may apply to the other callers as well. A fix could use `mv` to ensure the index has a proper path.

cwhelan:
This applies to `.bam` and `.bai` files in Manta as well. It's odd because Manta claimed to have fixed that issue long ago: https://github.com/Illumina/manta/issues/14. It might be worth investigating whether this is something particular to our setup somehow.

cwhelan:
Closed via https://github.com/broadinstitute/gatk-sv/pull/54

PR #54, opened 2020-09-20
Single sample pipeline bug fixes

cwhelan:
This fixes several bugs encountered by users of the single sample pipeline on different inputs that would result in the pipeline failing to complete with an error message:

* The `RunManta` task failed when run with a bam file input and an index file that did not conform to the naming scheme `$BAMNAME.bam.bai` https://github.com/broadinstitute/gatk-sv/issues/11
* The `CollectCoverage` task also failed to run when the index did not follow the expected naming convention. https://github.com/broadinstitute/gatk-sv/issues/48
* `PETest` and `SRTest` could fail if shards of the input VCF were empty (for example, if they had no calls on chrY). 
* Module0506 could fail in the `MasterVcfQc` workflow if shards of the VCF were empty (this was also encountered when samples were missing any calls on chrY.

### DIFF
--- a/wdl/Manta.wdl
+++ b/wdl/Manta.wdl
@@ -28,7 +28,7 @@ import "Structs.wdl"
 workflow Manta {
   input {
     File bam_or_cram_file
-    File? bam_or_cram_index
+    File bam_or_cram_index
     String sample_id
     File reference_fasta
     File? reference_index
@@ -81,7 +81,7 @@ workflow Manta {
 task RunManta {
   input {
     File bam_or_cram_file
-    File? bam_or_cram_index
+    File bam_or_cram_index
     String sample_id
     File reference_fasta
     File? reference_index
@@ -94,8 +94,8 @@ task RunManta {
   }
 
   Boolean is_bam = basename(bam_or_cram_file, ".bam") + ".bam" == basename(bam_or_cram_file)
+  String bam_ext = if is_bam then ".bam" else ".cram"
   String index_ext = if is_bam then ".bai" else ".crai"
-  File bam_or_cram_index_file = select_first([bam_or_cram_index, bam_or_cram_file + index_ext])
   File ref_index = select_first([reference_index, reference_fasta + ".fai"])
   File bed_index = select_first([region_bed_index, region_bed + ".tbi"])
 
@@ -118,7 +118,7 @@ task RunManta {
   # ensure there's sufficient disk space
   Float disk_overhead = 10.0
   Float bam_or_cram_size = size(bam_or_cram_file, "GiB")
-  Float bam_or_cram_index_size = size(bam_or_cram_index_file, "GiB")
+  Float bam_or_cram_index_size = size(bam_or_cram_index, "GiB")
   Float ref_size = size(reference_fasta, "GiB")
   Float ref_index_size = size(ref_index, "GiB")
   Float region_bed_size = size(region_bed, "GiB")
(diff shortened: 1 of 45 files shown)


wdl/Manta.wdl
```diff
@@ -148,6 +151,12 @@ task RunManta {
       rm ./runWorkflow.py
     fi
 
+    if [ ! -z ~{select_first([bam_or_cram_index, ""])} ]; then
+      if [ ! -f `dirname ~{bam_or_cram_file}`/~{expected_index_name} ]; then
+        ln -s ~{bam_or_cram_index} `dirname ~{bam_or_cram_file}`/~{expected_index_name} || true
+      fi
+    fi
+
```
mwalker174, line 159: Just a suggestion, I think this would be more readable:
```
String bam_ext = if is_bam then ".bam" else ".cram"
...
ln -s ~{bam_or_cram} sample.~{bam_ext}
ln -s ~{bam_or_cram_index} sample.~{bam_ext}.~{index_ext}
...
/usr/local/bin/manta/bin/configManta.py \
  --bam sample.~{bam_ext}
```

wdl/PETestChromosome.wdl
```diff
@@ -201,22 +201,28 @@ task PETest {
...
-      # Download twice
-      GCS_OAUTH_TOKEN=`gcloud auth application-default print-access-token` tabix -R region.merged.bed ~{discfile} | bgzip -c > PE_1.txt.gz
-      GCS_OAUTH_TOKEN=`gcloud auth application-default print-access-token` tabix -R region.merged.bed ~{discfile} | bgzip -c > PE_2.txt.gz
-      # Done if the downloads were identical, otherwise retry
-      cmp --silent PE_1.txt.gz PE_2.txt.gz && break       
-      x=$(( $x + 1))
-    done
-    echo "PE-tabix retry:" $x
-    cmp --silent PE_1.txt.gz PE_2.txt.gz || exit 1
+    if [ -s region.merged.bed ]; then
```
mwalker174, line 204: For future reference, which part of this was failing in empty shards?

mwalker174 (approved):
Thanks this looks great. I had a minor question and suggestion but otherwise it's good to merge.

wdl/PETestChromosome.wdl
cwhelan: The thing I'm trying to fix with this change cropped up after I made the change to the `SplitVCF` task in `Tasks02.wdl`. That task can now produce an empty VCF file (with header, but no records in it). When there aren't any records in the input VCF to this task, the awk and bedtools commands above end up producing an empty `region.merged.bed` file. When you give tabix an empty bed file, it interprets that as a lack of intervals, so the remote tabix command I wrapped here would try to localize the entire PE or SR file. This avoids that; afterwards `svtk pe-test` operates gracefully on the empty VCF shard, returning an empty stats table.

wdl/Manta.wdl
cwhelan: One thing that I found tricky in this fix is that `bam_or_cram_index` is an optional input to this task. Do you know if there's some way to run it successfully without passing an index in that I don't understand?

mwalker174: I see that it's always passed a value from 00a, so we can make it non-optional in the task and subworkflow.

wdl/PETestChromosome.wdl
mwalker174: Ok thanks. We did run into a problem like this, but it was only for the common regions, which motivated the changes to PE/SRTest in the other PR. The new GATK tool also will localize the whole remote file if the intervals list is empty, so I'll just merge these changes manually to include your empty file check.

wdl/Manta.wdl
cwhelan: I liked your suggestion better than how I had it originally; changed to implement it.

PR #62, opened 2020-10-07
Fix additional failures on empty shards in annotation module

cwhelan:
This fixes another error that cropped up in the annotation module with empty shards (ie chromsome-shard vcfs with no variants). The only change to logic is in `AnnotateChromosome.wdl`. Although `svtk annotate` had been modified to not fail in this case, the subsequent `grep` commands that counted variants could return with a non-zero exit code and cause the entire command block to fail.

The other changed files are updates to make WDLs and inputs validate with cromwell/womtool v 51+.


mwalker174 (approved):
Looks good to me

PR #63, opened 2020-10-14
Fix one more empty-file bug in Module07

cwhelan:
This fixes one more error that can occur in Module07 when a sample has no variants called on a contig. Modified the `ChromosomeAlleleFrequencies` workflow's `ShardVcf` task to always create at least one VCF shard (an empty one), even if there are no variants in the input VCF. 

I was able to verify that the original input VCF to Module07 that caused this error is now able to run through to completion of Module07 successfully.


mwalker174 (approved):
👍

PR #65, opened 2020-10-20
Xz fixes

xuefzhao:
this branch incorporates multiple recent fixes


mwalker174:
We should hold off on this until #51 goes in.

mwalker174:
Moved to #90

Issue #27, opened 2020-07-16
Too many json files


mwalker174:
Maintaining them all is cumbersome and error-prone. We do have docker updates automated, but it's vandalizing event the smallest of PRs with tons of unnecessary edits.

We need scripts to build jsons locally. This way fixed resources, used by many workflows (e.g. dockers, reference files, etc.), can be located in a central file. If we use consistent variable names across all wdls (see #16), a script would be able to use womtool to retrieve workflow inputs, and then reference both common and workflow-specific resources automatically.

mwalker174:
This was addressed in #84

Issue #16, opened 2020-07-01
Formatting improvements


mwalker174:
To-be-addressed from https://github.com/talkowski-lab/gatk-sv-v1/pull/252:

> * Some workflows currently require index files as inputs (e.g. module00c/CNMOPS). They should either be given default values or removed as inputs altogether.
> * Some tasks require index files as inputs. This matters in some cases because tasks are called from other workflows, requiring passing the input files around (e.g. ClinicalFiltering.wdl::resetFilter line 392, called from GATKSVPipelineClinical).
> * Because of the previous points, some workflows are still passing around index files. Note some index files have _idx in the file name, and some have _index
> * Camel case convention: I think in most projects acronyms are treated as "words", hence CNMOPS -> Cnmops, GATKSVPipelinePhase1 -> GatkSvPipelinePhase1. We've just been treating them as a grey area. I don't think this is a big deal, but since you're going through and making a camel case fix, I wanted to take the time to get an explicit statement on how we treat acronyms.
> * There are a few remaining references to "fam_file" instead of "ped_file"
>   -- "runtime_override_merge_fam_file_list" in GATKSVPipelineClinical.wdl
>   -- "cleaned_fam_file" in master_vcf_qc.wdl
>   -- "fam_file" in several of the mosaic WDLs

epiercehoffman:
- Module00a outputs pesr_disc and pesr_split files, and Module00c takes them as inputs but calls them PE_files and SR_files respectively
- Many inputs/outputs are named manta_vcf (or wham_vcf, melt_vcf, etc.) but do not always refer to the same file/type of file. Some are raw VCFs from the callers, some are standardized, and some are clustered. It would alleviate confusion to prefix them with the type, ie. call the Module01 input std_manta_vcf to match the name of the corresponding Module00c output.

mwalker174:
Echoing @ekarlins from #345 here:
`There are a lot of output files on Terra for each run of GATK-SV. The output files populate the data tables and are sorted in alphabetical order by output name. It's sometimes difficult to tell which step produced each output. Is it possible to add the module number to the beginning of the output name to make this more obvious?`

Thank you, great idea! We are generally trying to cut down on the number of data table columns needed, but given the number of modules this is likely to stay relevant for a while.

ekarlins:
Thanks @mwalker174!

Issue #10, opened 2020-06-30
Migrate to 1KGP testing


mwalker174:
Testing data currently contain sensitive data that cannot be released publicly. We should migrate over to a 2-batch test dataset built using the 1000 Genome Project (1KGP) high-coverage data that we are using for the [cohort mode Terra workspace](https://app.terra.bio/#workspaces/broad-firecloud-dsde-methods/gatk-sv-pipeline-1kgp) (not to be confused with the [single-sample mode 1KGP reference panel](https://github.com/broadinstitute/gatk-sv/blob/master/input_values/ref_panel_1kg_v2.json) which is a single batch composed of a different set of samples from 1KGP).

Ideally, all our testing should be self-contained, meaning that prerequisite cohort-dependent inputs for all modules (e.g. vcfs, metrics files, etc.) can be generated from the tests of earlier modules. Therefore, we will need separate tests for `batch1` and `batch2` starting at `GenerateSampleMetricsBatch` through `FilterBatch` and `GenotypeBatch`. Other downstream modules are run on the whole cohort (`batch1` and `batch2` together).

We will replace `small`/`large` test set designations. In the future, we can think about options to run on a subset of chromosomes to speed up testing. The one exception would be `GenerateSampleMetrics` - currently we test the batch version of this (`GenerateSampleMetricsBatch`). We should add another template for `GenerateSampleMetrics` itself to run on one sample, since this workflow is quite expensive.

A few technical notes:

- New input values need to be defined for `batch1` and `batch2` in `/input_values`. For cohort-level steps (mentioned above), let's define a third inputs file for a `1kgp_test` cohort (i.e. `1kgp_test.json`).
- Input data and configurations can be found in `inputs/terra_workspaces/cohort_mode` (after running `scripts/inputs/build_default_inputs.sh`). This includes CRAM and gVCF paths, batch membership assignments, and cohort-specific resource files (e.g. ped file).
- Copy and organize workflow inputs/outputs in `gs://gatk-sv-resources-public/test`, including metrics generated by enabling `run_module_metrics`.

epiercehoffman:
As part of the migration to 1KGP testing, we should test with multiple batches for increased robustness & specifically to test MergeCohortVcfs.wdl. This will also be beneficial for testing a future batch-combine workflow needed for Terra.

PR #67, opened 2020-10-21
Generate input files from templates [WIP]

cwhelan:
This is an initial demonstration of a system for dynamically generating input files from a set of templates and specified input values.

To generate input files, run `scripts/inputs/build_inputs.py`. This script will currently populate the `test/` and `inputs/` directories with json files that replicate what's currently in the files in those directories in master.

Input value files are located in the `input_values` directory. Any json files in the input_values directory are read into maps corresponding to the file name.

The script then finds any template files in the 'input_templates` and `test_input_templates` directories and renders them into output files using the jinja2 templating library.

Values can be referred to by their input file name + "." + input file key. For example, if the input_values directory contains a file called dockers.json containing the map `{ "sv_pipeline_docker" : "gatksv/sv-pipeline:tag" }`, the string `{{ dockers.sv_pipeline_docker }}` will be replaced with the string `gatksv/sv-pipeline:tag`.

Jinja2 filters can be applied. For example, to ensure that string values are quoted in json files, use the tojson filter: `{{ dockers.sv_pipeline_docker | tojson }}`

I've created input value maps `dockers`, `ref_panel_1kg`, `ref_panel_v1b`, and `resources_hg38` to hold related groups of input values. This way our most frequent code change, updating the sv-pipeline docker image tag, will only affect one line of the small `dockers.json` file.

I haven't fully templated all of the input files to the fullest extent. For example, there still seems to be a lot of overlap in values between files referred to as part of ref_panel_v1b and the test_large inputs, but copies of the files live in different directories and probably need some due diligence to be substituted for one another. Doing so might achieve a large reduction in the number of times listings of these files duplicated across the input files templates, however.

Git (or GitHub) seems to have munged up the renaming operations that I used to move input files into the template directories, so I'd recommend checking this branch out and browsing locally rather than trying to read the files on GitHub.


mwalker174 (commented):
This looks awesome! Noting that this will resolve #27. There is still some cleanup to do as described in #16 and #10 that will help with some of the redundant inputs.

I have a few comments:

1) The templates all have extension `.json` but aren't actually json files. It would be clearer to have an extension like `.json_templ` or similar.
2) Is there a way to make it easier to swap in different values files? For example, if one wanted to add support for hg19 they would create the new values file `resources_hg19.json` but then have to replace all occurrences of `resources_hg38.` with `resources_hg19.` in the templates. Is it possible to dereference value files somehow? For example, have a `resources.json` values file that redirects to either the hg38 or hg19 file? That way switching would be a 1-line change. This would be extremely useful for using different reference panels.
3) To build on (2), it would be great to eliminate the "test_small" and "test_large" redundancies that we now have in the template files. That is, we shouldn't really need `ModuleX.test_small.json` and `ModuleX.test_large.json`. It would be ideal to only have a `ModuleX.json` and then specify the reference panel values to use at build time.
4) Is it possible to have optionally-specified values? To support building jsons for different cohorts/reference panels, one would need to template the following inputs as well:
```
"Module01.del_bed": "gs://gatk-sv-resources/test/module00c/large/outputs/test_large.DEL.bed.gz"
```
But if the cohort hasn't been run yet, i.e. the file does not exist, (or in the case of test_small where the pipeline stops at 03) the builder would have to know to skip jsons where it doesn't have all the inputs.

cwhelan:
Closed in favor of https://github.com/broadinstitute/gatk-sv/pull/84

PR #68, opened 2020-10-23
Single sample pipeline CTX BND representation issues

cwhelan:
This fixes a couple of bugs that cropped up in a single sample run that called a reciprocal translocation:

I was basing the second coordinate off of the END tag for the variant but in this case it had already been reset to POS + 1, so we should use the END2 tag, if it's available, for the second breakpoint.

The metrics module doesn't like encountering PASS BND records that don't have SVLEN -- modified the scripts to not try to gather SVLEN for BND records.

### DIFF
--- a/scripts/docker/update_json_docker.sh
+++ b/scripts/docker/update_json_docker.sh
@@ -68,7 +68,7 @@ JSONS=$(printf "%s "  "${JSON_ARR[@]}")
 for name in "${DOCKER_NAME_ARR[@]}"; do
   docker_var=$(tr '-' '_' <<< "${name}")"_docker"
   docker_regex='\.'"${docker_var}"'"\s*:\s*".+\/.+:.+"'
-  cmd="perl -pi -e 's/${docker_regex}/.${docker_var}\": \"${DOCKER_ROOT}\/${name}:${DOCKER_TAG}\"/g' ${JSONS}"
+  cmd="perl -pi -e 's|${docker_regex}|.${docker_var}\": \"${DOCKER_ROOT}\/${name}:${DOCKER_TAG}\"|g' ${JSONS}"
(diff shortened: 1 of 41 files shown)


scripts/docker/update_json_docker.sh
```diff
@@ -68,7 +68,7 @@ JSONS=$(printf "%s "  "${JSON_ARR[@]}")
 for name in "${DOCKER_NAME_ARR[@]}"; do
   docker_var=$(tr '-' '_' <<< "${name}")"_docker"
   docker_regex='\.'"${docker_var}"'"\s*:\s*".+\/.+:.+"'
-  cmd="perl -pi -e 's/${docker_regex}/.${docker_var}\": \"${DOCKER_ROOT}\/${name}:${DOCKER_TAG}\"/g' ${JSONS}"
+  cmd="perl -pi -e 's|${docker_regex}|.${docker_var}\": \"${DOCKER_ROOT}\/${name}:${DOCKER_TAG}\"|g' ${JSONS}"
```
mwalker174, line 71: Just curious what this does?

mwalker174 (approved):
Looks fine, just a quick question.

scripts/docker/update_json_docker.sh
cwhelan: I just wanted to allow using `DOCKER_ROOT`s with forward slashes in their names, which we need to do to this script with gcr repo identifiers. Changing the `s///` command delimiter to `s|||` lets you put in forward slashes without escaping them, so you can invoke this script like this:

`scripts/docker/update_json_docker.sh -d . -r us.gcr.io/broad-dsde-methods/cwhelan -n sv-pipeline -t cw_ctx_fixes_d1e9818`

PR #69, opened 2020-10-23
Revise analyze monitoring logs script to skip downloading logs

epiercehoffman:
### Changes
Revise `analyze_monitoring_logs.py` to avoid having to download logs first. A new file was created so as not to deprecate the old log-analysis scripts, which are needed when the old monitoring script was used. New script is named with "2" to match the new monitoring script it goes with. 

Process to use:
* Run workflow with new monitoring script `cromwell_monitoring_script2.sh`. A copy of this script is located in the cloud at `gs://broad-dsde-methods-tbrookin/cromwell_monitoring_script2.sh` and at `gs://gatk-sv-resources/resources/cromwell_monitoring_script2.sh`
* Use `get_cromwell_resource_usage2.sh -u -r -o <out.tsv> <workflow-id>` to generate monitoring log summary table (TSV)
* Run analysis script: `python analyze_monitoring_logs2.py /path/to/log_summary_table /path/to/output_base [optional parameters]` (see comment at top of script for further documentation)

This script should produce functionally equivalent output to the previous script. There are some aesthetic differences, though, including the following (open to feedback):
* In the `all.tsv` output, tasks are named according to the format `<task>.shard-<shard#>.attempt-<attempt#>` (omitting shard/attempt where not applicable) rather than `<task_name>.<shard#>.<attempt#>.<job_id>`
* Rows are included for tasks that had a monitoring log with no usage entries (typically very short tasks). These rows have `NaN` values where resource usage was not recorded. Cost columns are calculated based on the minimum resource settings. Bars are *not* included in the cost chart for these tasks. 
* Column names have been modified to use input summary table naming style (camel-case)
* Disk usage only has one decimal point of precision, as recorded in the new monitoring script
* Percentage columns are represented as a percent rather than a fraction (ie. x100)
* Order of some columns changed, to group each type of resource together more for side-by-side comparison of resources requested vs. used

### Testing 
* Module03 was run with test_large inputs, using the new monitoring script, and analyzed as described above. Separately, the same workflow was run with the old monitoring script to provide a point of (non-identical) comparison. Each column was compared to verify that the calculations in the new script were correct (with some small deviation expected from a separate run of the workflow). An example summary log TSV can be provided upon request, as input for further testing. This test table included tasks for which some data was `NaN`.
* Tested with missing, empty, malformed (non-TSV format and missing columns) input files

### DIFF
--- a/scripts/cromwell/analyze_monitoring_logs2.py
+++ b/scripts/cromwell/analyze_monitoring_logs2.py
@@ -9,7 +9,7 @@ import logging
 
 # Synopsis:
 #  Generates summary statistics on Cromwell monitoring log summary table generated using the following command:
-#       get_cromwell_resource_usage2.sh -u -r meta.json > table.tsv
+#       get_cromwell_resource_usage2.sh -u -r workflow_id > table.tsv
 #  Cost estimates assume all machines are preemptible and have a fixed boot time. Resource
 #  costs are estimated for requesting optimal resources (equal to the max observed) uniformly across all shards
 #  ("static") and individually for each shard ("dynamic").
--- a/scripts/cromwell/analyze_monitoring_logs2.py
+++ b/scripts/cromwell/analyze_monitoring_logs2.py
@@ -1,20 +1,18 @@
 #!/bin/python
 
 import pandas as pd
-import glob
 import argparse
 import numpy as np
-from datetime import datetime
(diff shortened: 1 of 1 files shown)


mwalker174 (approved):
Looks good! Tested this on several workflows now without any issues.

PR #75, opened 2020-10-30
Move all dockers over to copies in us.gcr.io

cwhelan:
Moves all docker inputs to point to copies in GCR, in anticipation of the new DockerHub pull rate limits.


mwalker174 (approved):
Thank you, this looks good

PR #79, opened 2020-11-13
Regenotyping-related updates: filter by VF instead of count, update names

epiercehoffman:
### Updates
* In Module04.GetRegenotype, filter by variant frequency rather than count in cohort - keep variants with VF < 0.01 or sample count < 3 in cohort
* Change MergeCohortVcfs output bed file names (and related task names) to be more descriptive (and accurate… the Module04 `cohort_combined` input did not match up with the MergeCohortVcfs `cohort_combined` output and actually should have been `lookup`) - see table below detailing changes:
image

### Testing
* Womtool validation successful (v52)
* Testing completed for change to VF rather than count: ran `test_large` dataset through Module04 successfully. With this change, the number of variants to regenotype went from 688 to 406, roughly even between DELs and DUPs. The new set was verified to be a subset of the old one.
* Testing planned for name changes: just verify a successful run of `test_large` dataset through MergeCohortVcfs, Module04, and Module04b with new file names


wdl/GenotypeDepthPart2.wdl
```diff
@@ -221,9 +223,10 @@ task IntegrateDepthGq {
 task GetRegenotype{
   input{
     File depth_genotyped_vcf
-    File? cohort_combined_bed
-    File? cohort_sort
+    File? regeno_sample_counts_lookup
+    File? regeno_raw_combined_depth
     String Batch
+    Int? n_samples_cohort
```
mwalker174, line 229: Even though `n_samples_cohort` will always be defined as the workflow is written, it is good practice to make it non-optional, since the command section would not run correctly without it. In the call, you will need to force it using `n_samples_cohort=select_first([n_samples_cohort])`

wdl/GenotypeDepthPart2.wdl
```diff
@@ -254,8 +258,9 @@ task GetRegenotype{
     # VF<0.01 filter for missing sample variants
     while read chr start end variant type; do
      varid=$variant: #varid should be single variants
-     num=$(fgrep -m1 $varid ~{cohort_combined_bed}|cut -f8)
-     if ((num<100));then
+     num=$(fgrep -m1 $varid ~{regeno_sample_counts_lookup}|cut -f8)
+     thresh=$((num*100))
+     if ( ((thresh < n_samp)) || ((num <= 3)) );then
```
mwalker174, line 263: Both 100 and 3 should be parameters, with a default values set in `Module04.wdl` like this: `Float regeno_max_allele_freq = 0.01` and `Int regeno_allele_count_threshold = 3`

wdl/GenotypeDepthPart2.wdl
mwalker174: Also, I realize you found it this way, but the same applies to `regeno_sample_counts_lookup` and `regeno_raw_combined_depth`

mwalker174 (commented):
Much easier to understand with the new variable names. I do think we should parameterize the AF/AC parameters for the regenotyping task, though.

wdl/GenotypeDepthPart2.wdl
epiercehoffman: Good point. Done

wdl/GenotypeDepthPart2.wdl
epiercehoffman: Done

epiercehoffman:
### Further updates
* Tasks GetRegenotype, GetMedianSubset, and MedianIntersect were moved from Module04.GenotypeDepthPart2 to Module04b, because they are related to regenotyping so it makes more sense to organize the code that way. This meant shifting the relevant inputs from Module04 to Module04b, and the `regeno_beds` parameter became an intermediate in 04b. Also, Module04 now outputs `regeno_coverage_medians` to feed into regenotyping in 04b; a set of files were added to the Module04b test JSON to serve as test data for this argument.
* A task was added to Module04b to concatenate the batch sample lists into a cohort list and count the number of samples in the cohort, which allowed for the removal of the `cohort_samplelist` and `n_samples_cohort` inputs to Module04b
* GetRegenotype set to 3 preemptible tries by default instead of 0
* Add top-level runtime attribute override parameters for several tasks (not exhaustive - there are still some missing)
* Fix spacing and indentation in several WDLs (this resulted in many lines changed - apologies for the difficulties that may raise in reviewing this PR)

### Testing
* WDLs and JSONs validated successfully with womtool-v52
* MergeCohortVcfs, Module04, and Module04b ran in sequence successfully on test_large data (caveat: just one batch)

mwalker174 (approved):
Looks great! Merge at will.

PR #83, opened 2020-11-19
Move regenotyping tasks from MergeCohortVcfs to Module04b

epiercehoffman:
### Updates
Move all regenotyping-related tasks (making various cohort-level BED files) from MergeCohortVcfs to Module04b, in a continuation of the reorganization started in #79 to consolidate regenotyping tasks into one workflow to improve and simplify organization. This change involved moving and altering tasks, shifting some inputs, and removing other Module04b inputs that were no longer necessary.

### Testing
* Validated WDLs and JSONs with womtool-v52
* MergeCohortVcfs and Module04b were tested independently using (updated) standard inputs (`test_large` test data). They each ran successfully, and the outputs section was checked to verify that it included all (and only) the correct files. The `regeno_bed` output from `MedianIntersect` was checked against the previous version to verify that it was identical.

### DIFF
--- a/wdl/MergeCohortVcfs.wdl
+++ b/wdl/MergeCohortVcfs.wdl
@@ -15,19 +15,10 @@ workflow MergeCohortVcfs {
     Array[File] depth_vcfs    # Filtered depth VCFs across batches
     Array[File] pesr_vcfs     # Filtered PESR VCFs across batches
     String cohort             # Cohort name or project prefix for all cohort-level outputs
-    File contig_list
     String sv_pipeline_docker
-    String sv_base_mini_docker
     RuntimeAttr? runtime_attr_merge_pesr
     RuntimeAttr? runtime_attr_merge_depth
-    RuntimeAttr? runtime_attr_regeno_raw_combined_depth
-    RuntimeAttr? runtime_attr_regeno_merged_depth
-    RuntimeAttr? runtime_attr_sample_lookup
-    RuntimeAttr? runtime_attr_concat_samplecountlookup
-    RuntimeAttr? runtime_attr_concat_sampleidlookup
   }
-
-  Array[Array[String]] contigs = read_tsv(contig_list)
 
   call MergeVcfs as MergePESRVcfs {
     input:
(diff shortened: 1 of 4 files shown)


wdl/MergeCohortVcfs.wdl
```diff
@@ -161,120 +104,11 @@ task MergeDepthVcfs {
 
   output {
     File merged_vcf = "~{prefix}.vcf.gz"
-    File regeno_merged_depth_clustered = "~{cohort}.regeno.merged_depth_clustered.bed"
   }
   command <<<
-
     set -euxo pipefail
     /opt/sv-pipeline/04_variant_resolution/scripts/merge_vcfs.py ~{write_lines(vcfs)} ~{prefix}.vcf
-    rm ~{sep=' ' vcfs}
```
mwalker174, line 170: What happened to this `rm`?

wdl/Module04b.wdl
```diff
@@ -19,23 +19,84 @@ workflow Module04b{
...
     RuntimeAttr? runtime_attr_get_regeno
     RuntimeAttr? runtime_attr_get_median_subset
     RuntimeAttr? runtime_attr_median_intersect
   }
+
+  call ClusterMergedDepthBeds {
+    input:
+      cohort_depth_vcf = cohort_depth_vcf,
+      cohort = cohort,
+      prefix = cohort + ".all_batches.depth",
```
mwalker174, line 50: Let's get rid of this input - it doesn't actually change output file names.

mwalker174 (approved):
Glad to finally have 04b properly encapsulated! I have a couple of minor suggestions.

wdl/MergeCohortVcfs.wdl
epiercehoffman: Since the rest of the task got moved to 04b, I didn't think it was necessary to remove the input VCFs anymore - it was there originally to save on disk space for the rest of the task

wdl/Module04b.wdl
epiercehoffman: Good call, it was a bit lazy to leave it in when it's no longer relevant. Fixed and re-testing to be careful

Issue #77, opened 2020-11-03
Module04b fails when all regeno_bed files are empty


epiercehoffman:
### Affected module(s) or script(s)
Module04b

### Description 
When all files in the `regeno_beds` input to Module04b are empty, Module04b.Genotype_2.GetRegenotype fails because it does not produce an output file, so Cromwell fails to de-localize the expected output file.

#### Expected behavior
Module04b should handle empty inputs rather than failing.

epiercehoffman:
Pasting details here to keep track:

> For the PD cohort with 2 non-empty “median_missing_sorted.bed”, Module04b was finished with results.
> But for another cohort, LBD, with all empty “median_missing_sorted.bed”, the “master_regeno.bed” in /call-MergeList is empty and all /call-GetRegenotype scripts failed when scattering all cross 16 batches since no per batch .to_regeno.bed generated

> Use one batch as an example, the following script in call-GetRegenotype/ failed since it cannot generate output ".to_regeno.bed" because "master_regeno.bed" from call-MergeList is empty

```
svtk vcf2bed LBD_samples_2F2M_batch.2.depth.vcf.gz LBD_samples_2F2M_batch.2.bed
sample=$(fgrep -v "#" LBD_samples_2F2M_batch.2.bed|awk '{if($6!="" )print $6}' |head -n1|cut -d"," -f1) # a sample in batch is needed for regenotyping
while read chr start end variant s type; do
    printf "$chr\t$start\t$end\t$variant\t$sample\t$type\n">>LBD_samples_2F2M_batch.2.to_regeno.bed # modify this to suit intput for Rdtest
done <  /cromwell-execution/Module04b/6562d968-3734-4390-8f97-0ea097caaaa1/call-MergeList/master_regeno.bed
```

> Error message in GetRegenotype.log, it failed for attempting to copy .to_regeno.bed

```
Running command: sudo gsutil -q -m cp -L /var/log/google-genomics/out.log /mnt/local-disk/LBD_samples_2F2M_batch.2.to_regeno.bed gs://cromwell-execution/Module04b/6562d968-3734-4390-8f97-0ea097caaaa1/call-Genotype_2/shard-1/Regenotype/08570764-d41f-42bb-ae03-34d279798cb3/call-GetRegenotype/LBD_samples_2F2M_batch.2.to_regeno.bed
E: command failed: CommandException: No URLs matched: /mnt/local-disk/LBD_samples_2F2M_batch.2.to_regeno.bed
[...]

PR #86, opened 2020-12-11
Regeno coverage medians handling: streamline and bug fixes

epiercehoffman:
### Updates
* Merge `regeno_coverage_medians` files for each batch before outputting from Module04, to make inputs/outputs for large datasets easier to wrangle
* Fix bug in 04b that would have improperly handled multi-batch `regeno_coverage_medians` (without merging files from 04, would have needed Array[Array[File]] input to account for multiple batches)
* Fix bug in 04b that failed when no variants to regenotype (#77)
  * New behavior: if the file from MergeList contains no non-empty lines (no variants to regenotype), the regenotyping is skipped and the output VCFs are just copies of the input VCFs
  * Also added an output to 04b called `number_regenotyped_file` which contains the number of lines from `MergeList`, so that it's easy to tell if regenotyping occurred (number > 0) or not (number = 0) without doing a `diff` on the input/output VCFs
* Formatting, indentation, and spacing improvements

### Testing
Completed:
* Tested running with one batch (`test_large` data) through Module04 and Module04b. It previously had 2 `regeno_coverage_medians` files, so the merge task in 04 was not superfluous & the effects of that change in 04b could be tested. It had 406 variants to regenotype, so it ran the full 04b workflow. Both workflows completed successfully and output was identical to latest master code. The default `regeno_coverage_medians` input to 04b was updated in the JSON template to use the file from this test run.
* Tested Module04 and Module04b two GMKF batches to confirm that a multi-batch run of Module04b was successful. It had over 60k variants to regenotype, so the full 04b workflow ran to completion.
* Tested partial workflow with dummy empty `regeno_beds` input to `MergeList` (all other inputs used `test_large` data) and verified that the run was successful, that the number of variants counted for regenotyping was 0 and that was reflected in the output file, that the regenotyping tasks were not executed, and that the output VCF was identical to the input depth VCF. 

Caveats:
* No comparison to latest master was made for multi-batch test because it took so long to run everything
* Did not test a full workflow with a real batch that produced empty `regeno_beds`

### DIFF
--- a/test_input_templates/module04b/Module04b.test.json.tmpl
+++ b/test_input_templates/module04b/Module04b.test.json.tmpl
@@ -8,9 +8,9 @@
 
   "Module04b.cohort": {{ test_batch.batch_name | tojson }},
   "Module04b.contig_list": {{ reference_resources.primary_contigs_list | tojson }},
-  "Module04b.regeno_coverage_medians": ["gs://broad-dsde-methods-eph/Module04StressTest/Module04/2430c16a-23ee-44b4-9305-b35bcba352d9/call-GenotypeDepthPart2/GenotypeDepthPart2/f1483c4e-9e57-47ff-926f-3b354a2aa678/call-MergeRegenoCoverageMedians/test_large.regeno.coverage_medians_merged.bed"],
+  "Module04b.regeno_coverage_medians": {{ test_batch.regeno_coverage_medians | tojson }},
 
-  "Module04b.RD_depth_sepcutoffs": ["gs://talkowskicromwellresults/M04/Module04/89617399-fcaa-4aef-9759-7f3d04f328d3/call-GenotypeDepthPart1/GenotypeDepthPart1/dd257617-3e1f-47b5-b17b-2144c1931986/call-TrainRDGenotyping/TrainRDGenotyping/8b8ea223-1bc1-4e7b-a07c-ca2b47e8e699/call-UpdateCutoff/depth_sepcutoff.txt"],
+  "Module04b.RD_depth_sepcutoffs": [{{ test_batch.depth_gt_rd_sep_file | tojson }}],
 
   "Module04b.cohort_depth_vcf": {{ test_batch.cohort_depth_vcf | tojson }},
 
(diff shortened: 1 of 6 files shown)


wdl/GenotypeDepthPart2.wdl
```diff
@@ -202,3 +211,39 @@ task IntegrateDepthGq {
...
+    preemptible_tries: 3,
+    max_retries: 1
+  }
+  RuntimeAttr runtime_attr = select_first([runtime_attr_override, default_attr])
+
+  output {
+    File regeno_coverage_medians = "~{batch}.regeno.coverage_medians_merged.bed"
+  }
+  command <<<
+    cat ~{sep=' ' regeno_coverage_medians_array} | fgrep -v "start" > ~{batch}.regeno.coverage_medians_merged.bed
```
mwalker174, line 237: I'd assume `-v start` is to filter out a header line. Can we search for the full header line for the off chance that `start` appears in a record?

test_input_templates/module04b/Module04b.test.json.tmpl
```diff
@@ -8,10 +8,7 @@
 
   "Module04b.cohort": {{ test_batch.batch_name | tojson }},
   "Module04b.contig_list": {{ reference_resources.primary_contigs_list | tojson }},
-  "Module04b.regeno_coverage_medians": [
-    "gs://broad-dsde-methods-eph/Module04StressTest/Module04/6d90e729-0cd1-4752-94b3-e0ad128c6dc4/call-GenotypeDepthPart2/GenotypeDepthPart2/2b962a21-897e-49ac-8055-118ba459f7e4/call-RDTestGenotypeOver5kb/shard-0/cacheCopy/gt5kb.aaaaaa.bed.median_geno",
-    "gs://broad-dsde-methods-eph/Module04StressTest/Module04/6d90e729-0cd1-4752-94b3-e0ad128c6dc4/call-GenotypeDepthPart2/GenotypeDepthPart2/2b962a21-897e-49ac-8055-118ba459f7e4/call-RDTestGenotypeOver5kb/shard-1/cacheCopy/gt5kb.aaaaab.bed.median_geno"
-  ],
+  "Module04b.regeno_coverage_medians": ["gs://broad-dsde-methods-eph/Module04StressTest/Module04/2430c16a-23ee-44b4-9305-b35bcba352d9/call-GenotypeDepthPart2/GenotypeDepthPart2/f1483c4e-9e57-47ff-926f-3b354a2aa678/call-MergeRegenoCoverageMedians/test_large.regeno.coverage_medians_merged.bed"],
```
mwalker174, line 11: Let's take this opportunity to fix this template. Can you redefine `regeno_coverage_medians`, `RD_depth_sepcutoffs`, and `depth_vcfs` in the `test_batch`, i.e. in `input_values/test_batch_large.json`? Also let's move these inputs to `gs://gatk-sv-resources/test/module04b/input/`.

mwalker174 (approved):
Thank you @epiercehoffman, just a couple of minor comments.

test_input_templates/module04b/Module04b.test.json.tmpl
epiercehoffman: Done - reformatted template, moved regeno_coverage_medians, and it turns out the other files were already in that bucket under 04 outputs. Tested template generation

wdl/GenotypeDepthPart2.wdl
epiercehoffman: Done (grep first 4 columns of header, excluding sample IDs), and tested with test_large batch to make sure it ran properly through Module04.GenotypeDepthPart2 and the regeno_coverage_medians output was identical

PR #88, opened 2020-12-15
Break up Module0506 stages

mwalker174:
To reduce workflow size (particularly for larger cohorts), Module0506 is now composed of 4 main subworkflows:
1. `Module0506Cluster`
2. `Module0506ComplexResolve`
3. `Module0506ComplexGenotype`
4. `Module0506Clean`

The master `Module0506` workflow still exists and is used for the SingleSample, pipeline, Batch pipeline, and Tests.

Also deleted the underscore in appearances of `05_06`.


TedBrookings (approved):
Looks good to me. It took awhile to manually diff the new vs old code and had one false alarm due to legal out-of-order use of UpdateSrList in the old code.

PR #89, opened 2020-12-15
Optional case counts, PE, and SR file inputs for SingleSample workflow

mwalker174:
Lets users bypass Module00a in the SingleSample workflow.

### DIFF
--- a/wdl/GATKSVPipelineSingleSample.wdl
+++ b/wdl/GATKSVPipelineSingleSample.wdl
@@ -38,11 +38,14 @@ workflow GATKSVPipelineSingleSample {
     Boolean use_melt = true
     Boolean use_wham = true
 
-    # Used in lieu of running tool
+    # If Module00a outputs already prepared
     File? case_delly_vcf
     File? case_manta_vcf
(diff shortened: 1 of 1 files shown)


cwhelan (commented):
This looks good to me, nice addition. 

I have one tiny complaint about the preexisting but related code: one thing that always trips me up when I follow the logic around whether 00a gets run is that running the callers in 00a is triggered by the docker string inputs being defined, and that we purposefully set them to undefined by setting them to `NONE_STRING_`, which should never be defined. It makes me want to rename `NONE_STRING_` to `UNDEFINED_STRING_` to help me remember what it is. I know we follow a similar pattern in other WDL files, though, so if other people don't have the same problem remembering it that I do then feel free to ignore this.

mwalker174:
Thanks @cwhelan. I chose `NONE_STRING_` because `None` is the null type for WDL (see https://github.com/openwdl/wdl/pull/263). I'm not opposed to `UNDEFINED_STRING_` either, though it would be better if we could find a way to avoid using this hack in the first place.

Issue #87, opened 2020-12-14
Error running tasks04.CountPE


henriqueribeiro:
I'm running `GATKSVPipelineSingleSample.wdl` and I'm facing this error in Module04:

> A USER ERROR has occurred: An index is required but was not found for file /tmp/scratch/xxxxx-gwf-core/cromwell-execution/GATKSVPipelineSingleSample/4e9e57d2-48ef-4b52-9cb8-5324c64cfddc/call-Module00c/Module00c/d2cdda29-616b-44ea-8e1a-9225c05e73da/call-EvidenceMerging/EvidenceMerging/c4f03868-e71b-452b-83f2-edf54f0f580b/call-MergePEShards/test_na12878.PE.txt.gz. Support for unindexed block-compressed files has been temporarily disabled. Try running IndexFeatureFile on the input.

However, Module00c is also outputting `test_na12878.PE.txt.gz.tbi` but is not being fed to this task. Is this correct?

Thanks in advance

mwalker174:
Hi @henriqueribeiro,

On GCP, we use GATK to stream PE data directly from the bucket to avoid localizing the entire file. This won't work with any other backend (eg local or AWS). If you are running with another backend, you should just need to add the index as an input so that it localizes.

henriqueribeiro:
it makes sense. I will add it to the wdl file. After testing it, does it make sense to do a pull request?

mwalker174:
@henriqueribeiro Yes a PR would be helpful. If you could find instances where `localization_optional` is used and add index inputs?

henriqueribeiro:
Just did a pull request. Can you please review it?
Thanks a lot

henriqueribeiro:
solved

PR #91, opened 2020-12-17
Add index inputs to Tasks04

henriqueribeiro:
Fix issue https://github.com/broadinstitute/gatk-sv/issues/87


mwalker174 (approved):
Thank you @henriqueribeiro, this looks good. Note that there are other tasks in the pipeline (eg in Modules 02 and 0506) where an explicit index for evidence files also are needed for local backends.

henriqueribeiro:
I just did the modifications for module0506 and it worked fine. I will do a PR. 
However for module02 (at least the tasks on SimpleSample) it worked fine without it. Can you give an example where is it needed and I will do the change? Thanks

mwalker174 (approved):
Thanks @henriqueribeiro! This looks fine to me. No need to add index inputs for 02 for now.

PR #95, opened 2021-01-13
Remove PESR_GT_OVERDISPERSION filter from single sample pipeline

cwhelan:
Changing to median-based genotyping caused many additional variants to be filtered with `PESR_GT_OVERDISPERSION` in the single sample results. Analysis of concordance with PacBio calls showed that this filter was not particularly good at filtering out false positives in the single sample context under either the original or changed genotyping schemes. This PR turns the `PESR_GT_OVERDISPERSION` filter into an INFO flag at the end of the single sample pipeline.

Also includes an fix to the `dockers.json` specification of the location of `sv_pipeline_qc_docker` and a fix for a bug in svtest vcf analysis code that was causing an error in the single sample pipeline.

### DIFF
--- a/src/svtest/svtest/cli/vcf.py
+++ b/src/svtest/svtest/cli/vcf.py
@@ -178,6 +178,9 @@ def get_metrics(ftest, fbase_vcf, fbase_bed, contigs, variant_types, min_ro, pad
         test_tree = iu.create_trees_from_records(test_records, variant_types, contigs, padding=padding)
         base_tree = iu.create_trees_from_bed_records(base_records, variant_types, contigs, padding=padding)
         base_pass_tree = None
+    else:
+        base_tree = None
+        base_pass_tree = None
 
     if base_tree is not None:
         metrics, fp_intervals, fn_intervals = add_evaluation_metrics(metrics, test_tree, base_tree, variant_types, min_ro, metric_prefix)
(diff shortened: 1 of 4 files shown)


src/svtest/svtest/cli/vcf.py
```diff
@@ -178,6 +178,9 @@ def get_metrics(ftest, fbase_vcf, fbase_bed, contigs, variant_types, min_ro, pad
         test_tree = iu.create_trees_from_records(test_records, variant_types, contigs, padding=padding)
         base_tree = iu.create_trees_from_bed_records(base_records, variant_types, contigs, padding=padding)
         base_pass_tree = None
+    else:
+        base_tree = None
+        base_pass_tree = None
```
mwalker174, line 183: Thanks for this fix

mwalker174 (approved):
Looks good! Did we already determine that we couldn't reset these filters within 0506 so that the single sample and cohort pipelines are consistent?

cwhelan:
I've been being conservative about modifying the cohort pipeline with these changes since I've never evaluated them on a cohort callset. We should check out the effect of the PESR_GT_OVERDISPERSION filter and the HIGH_SR_BACKGROUND filter on a cohort callset, particularly one with a diverse set of samples in terms of sample quality -- it may be that the filters are still useful there. If they're not then I agree that we should probably just remove them from 05/06.

Moving BOTHSIDES_SUPPORT to an INFO flag rather than a filter does seem to be something that should be moved into the cohort pipeline as well, though.

cwhelan:
I created a new issue to track removing these filters from the cohort mode pipeline after evaluating whether they should be removed: https://github.com/broadinstitute/gatk-sv/issues/97

PR #96, opened 2021-01-16
Templates to populate single sample featured terra workspace

cwhelan:
Creates templates to populate the sample and participant data tables, the workspace data, and the method configuration input json. 

The method configuration json refers to workspace attributes for all reference resources and ref panel parameters of type File or String, but is populated directly with any inputs of type Array because Terra doesn't support uploading Array type values inside TSVs (This was a compromise suggested by the comms team). 

Created a template for the dashboard documentation as well despite the fact that it has no templated values, just for the sake of keeping all the files for the workspace together. 

Single sample pipeline case inputs are now templated as a 'test_batch' input value bundle.


.gitignore
```diff
@@ -30,5 +30,5 @@ src/svtk/svtk/utils/helpers.cpython-37m-darwin.so
 src/svtk/svtk/utils/s3bam.pyc
 src/svtk/svtk/utils/utils.pyc
 src/svtk/svtk/vcfcluster.pyc
-test/
-inputs/
+/test/
+/inputs/
```
mwalker174, line 34: Was this causing problems?

input_templates/GATKSVPipelineSingleSample.ref_panel_1kg.na12878.json.tmpl
```diff
@@ -1,8 +1,8 @@
 {
-  "GATKSVPipelineSingleSample.sample_id" : "NA12878",
-  "GATKSVPipelineSingleSample.batch" : "test_na12878",
-  "GATKSVPipelineSingleSample.bam_or_cram_file" : "gs://fc-56ac46ea-efc4-4683-b6d5-6d95bed41c5e/CCDG_13607/Project_CCDG_13607_B01_GRM_WGS.cram.2019-02-06/Sample_NA12878/analysis/NA12878.final.cram",
-  "GATKSVPipelineSingleSample.requester_pays_cram": "true",
+  "GATKSVPipelineSingleSample.sample_id" : {{ test_batch.sample_id | tojson }},
```
mwalker174, line 2: `test_batch` seems like a confusing identifier since it's only one sample. Could this be renamed to `single_sample` or something similar?

mwalker174 (approved):
Great! Creating a workspace from scratch was easy with this. Just a couple of minor questions. I would add that the root dir is getting a bit cluttered with input-related subdirs. Would it be hard to move these into `inputs`?

.gitignore
cwhelan: Not having the leading slash was making git ignore files in subdirectories named `test` or `inputs`. Specifically I ran into it ignoring changes made to the `build_default_inputs.sh` script in `scripts/inputs`.

cwhelan:
@mwalker174 made the changes you suggested -- do these look better to you now?

.gitignore
```diff
@@ -30,5 +30,5 @@ src/svtk/svtk/utils/helpers.cpython-37m-darwin.so
 src/svtk/svtk/utils/s3bam.pyc
 src/svtk/svtk/utils/utils.pyc
 src/svtk/svtk/vcfcluster.pyc
-test/
-inputs/
+/test/
```
epiercehoffman, line 33: Should this now be /test_inputs_*?

input_templates/terra_workspaces/single_sample/single_sample_workspace_dashboard.md.tmpl
```diff
@@ -0,0 +1,153 @@
+## GATK-SV Best Practices for Structural Variation Discovery on Single Samples
```
epiercehoffman, line 1: I was wondering why this was a template? Was it just so that the README ends up in the same directory that's generated for the other Terra files?

epiercehoffman (approved):
Looks good! I tried it out by cloning the latest version of the workspace that I had, deleting the existing data tables, and uploading each file. It went smoothly, and just for good measure I started a job and let it run for 5 minutes, and it didn't complain about any of the inputs so that seemed successful.

I had a few questions (as opposed to actual feedback), some of which I commented, but some more here:
1. Was the participant TSV something that was required for the featured workspace? I hadn't been using one
2. What would you think of adding instructions on how to use the workspace on your own data to the dashboard? Users should hopefully have already done a Terra tutorial previously, but I do think that it could be helpful to state things like a) delete the existing sample data & upload your own following the same TSV format, b) should you click "create participant/sample associations" when uploading sample and participant TSVs, c) no need to touch the workspace data or method configuration unless you wish to change one of the default settings or reference files
3. I assume you're planning to have the workspace already set up with workspace data & the corresponding method configuration, right? So the main users of the templated file generation will be us?

Finally, I do have some worries about template files piling up once we do a similar thing for the Terra cohort pipeline, as it looks like with this model we will need to have a separate Terra-specific template for each workflow. Of course, most of the time we will only have to actually edit the input _values_ rather than the templates, so it's still far better than the previous JSON system. But, it could be a lot of files to edit if inputs are added/names are changed. Could it work to have one template per workflow but tag template items with "this" or "workspace" if that's how they should be configured for Terra, and do away with the separate Terra template? Or is that too complicated with the current JSON-generation implementation / are there cases where it would not work to use the same variable names in Terra?

epiercehoffman:
> Great! Creating a workspace from scratch was easy with this. Just a couple of minor questions. I would add that the root dir is getting a bit cluttered with input-related subdirs. Would it be hard to move these into `inputs`?

Actually, what about also moving the `test_inputs_*` directories into a `test_inputs` directory?

input_templates/terra_workspaces/single_sample/single_sample_workspace_dashboard.md.tmpl
cwhelan: Yes, I just wanted all the files to create the workspace to be together in the same output directory for easy creation of the workspace.

cwhelan:
@epiercehoffman Thanks for the comments, I will address them in turn. To answer this question:

> I assume you're planning to have the workspace already set up with workspace data & the corresponding method  configuration, right? So the main users of the templated file generation will be us?

Yes, this is for us (the gatk-sv team). My vision for this is that whenever we  need to create a new version of a terra workspace (because we want to release a new code version), we can generate these files, delete the old version of the workspace, and create a new workspace from these files in a matter of a few clicks and TSV uploads. That way we don't have to worry about, for example, going into a preexisting workspace and deleting old parameters that we've dropped or changed the name of -- I've seen some of my workspaces get polluted with obsolete data or variables that way in the past. Plus we'll have version history on everything associated with the workspaces, tied to our github commits and release tags. If this model doesn't make sense to you, though, let me know.

cwhelan:
@epiercehoffman 

>Was the participant TSV something that was required for the featured workspace? I hadn't been using one

When I first learned Terra the docs made it sound like the participant table was required -- I'm surprised to hear it's not technically necessary! I think it may be "best practice". I could ask the comms team about whether that's true for the featured workspace or not.

.gitignore
cwhelan: Yes although following your suggestion below to consolidate the test input files under one directory I moved this to `/test_inputs`.

cwhelan:
> Actually, what about also moving the `test_inputs_*` directories into a `test_inputs` directory?

Done

cwhelan:
> Finally, I do have some worries about template files piling up once we do a similar thing for the Terra cohort pipeline, as it looks like with this model we will need to have a separate Terra-specific template for each workflow. Of course, most of the time we will only have to actually edit the input values rather than the templates, so it's still far better than the previous JSON system. But, it could be a lot of files to edit if inputs are added/names are changed. Could it work to have one template per workflow but tag template items with "this" or "workspace" if that's how they should be configured for Terra, and do away with the separate Terra template? Or is that too complicated with the current JSON-generation implementation / are there cases where it would not work to use the same variable names in Terra?

I definitely see your point. I tried experimenting with a couple of approaches but with the way that I have things set up I couldn't find an elegant way to dynamically choose whether to populate a template with a value from `input_values` or with a reference to `${this.var}` or `${workspace.var}`. I created an issue to track your suggestion: https://github.com/broadinstitute/gatk-sv/issues/103 so that we can circle back to this in the future.

cwhelan:
@epiercehoffman I've finished responding to your comments; are you OK with these changes?

scripts/inputs/clean_default_inputs.sh
```diff
@@ -33,4 +33,4 @@ fi
 rm -r ${BASE_DIR}/inputs
 rm -r ${BASE_DIR}/test_inputs_small
```
epiercehoffman, line 34:
```suggestion
rm -r ${BASE_DIR}/test_inputs
```

scripts/inputs/clean_default_inputs.sh
```diff
@@ -33,4 +33,4 @@ fi
...
 rm -r ${BASE_DIR}/test_inputs_large
-rm -r ${BASE_DIR}/test_inputs_single_sample
\ No newline at end of file
+rm -r ${BASE_DIR}/test_inputs_single_sample
```
epiercehoffman, line 36:
```suggestion
```

epiercehoffman:
> @epiercehoffman I've finished responding to your comments; are you OK with these changes?

Looks good! The only thing is that clean_default_inputs.sh needs to be updated to reflect the new test_inputs directory - I made suggestions for that above

cwhelan:
Thanks for catching that @epiercehoffman . Fixed.

Issue #66, opened 2020-10-21
Switched order of Number and Type keys in VCF END2 meta info line causes problems in htsjdk


cwhelan:
Module0506's `postCPX_cleanup.py` writes the `END2` meta info line as:

```
##INFO=<ID=END2,Type=Integer,Number=1,Description="Position of breakpoint on CHR2">
```

The switch from the usual ordering of Number,Type causes htsjdk to throw an error when trying to parse the file:

```
htsjdk.tribble.TribbleException$InvalidHeader: Your input file has a malformed header: Tag Type in wrong order (was #2, expected #3) in line <ID=END2,Type=Integer,Number=1,Description="Position of breakpoint on CHR2">
        at htsjdk.variant.vcf.VCF4Parser.parseLine(VCFHeaderLineTranslator.java:172)
        at htsjdk.variant.vcf.VCFHeaderLineTranslator.parseLine(VCFHeaderLineTranslator.java:58)
        at htsjdk.variant.vcf.VCFCompoundHeaderLine.<init>(VCFCompoundHeaderLine.java:215)
        at htsjdk.variant.vcf.VCFInfoHeaderLine.<init>(VCFInfoHeaderLine.java:56)
        at htsjdk.variant.vcf.AbstractVCFCodec.parseHeaderFromLines(AbstractVCFCodec.java:192)
        at htsjdk.variant.vcf.VCFCodec.readActualHeader(VCFCodec.java:111)
        at htsjdk.tribble.AsciiFeatureCodec.readHeader(AsciiFeatureCodec.java:79)
        at htsjdk.tribble.AsciiFeatureCodec.readHeader(AsciiFeatureCodec.java:37)
        at htsjdk.tribble.TribbleIndexedFeatureReader.readHeader(TribbleIndexedFeatureReader.java:261)
        ... 16 more

```

nh13:
@cwhelan just hit this issue, I am not 100% certain if this is a bug in htsjdk or a bug here.  

See: https://github.com/samtools/hts-specs/issues/642

cwhelan:
Yes, we changed the order of the tags a while back on our end but it looks like you're using a VCF generated before that.

[...]

PR #98, opened 2021-01-20
Fix the header line for the INFO field END2 in `postCPX_cleanup.py`

cwhelan:
One of the scripts that writes an END2 header line was writing it with the `Type` and `Number` fields in the wrong order, which was causing htsjdk-based tools to throw an error when parsing it. Fixes https://github.com/broadinstitute/gatk-sv/issues/66.


mwalker174 (approved):
👍

PR #102, opened 2021-01-26
single sample pipeline sample QC

cwhelan:
This modifies the single sample pipeline to apply a filter to all of the variants in the final VCF if the sample fails the QC check for WGD score or rm_mean_depth, or both. This is intended to make it more clear when a sample is incompatible with the reference panel and the results should be analyzed carefully before being accepted.

 Also updates the qc definitions file to add depth as a metric and relaxes the expected counts of variants to make it less likely for good samples to have failing qc metrics.

### DIFF
--- a/wdl/GATKSVPipelineSingleSample.wdl
+++ b/wdl/GATKSVPipelineSingleSample.wdl
@@ -16,6 +16,7 @@ import "GermlineCNVCase.wdl" as gcnv
 import "SingleSampleFiltering.wdl" as SingleSampleFiltering
 import "GATKSVPipelineSingleSampleMetrics.wdl" as SingleSampleMetrics
 import "Utils.wdl" as utils
+import "TestUtils.wdl" as tu
 import "Structs.wdl"
 
 # GATK SV Pipeline single sample mode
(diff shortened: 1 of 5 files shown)


wdl/GATKSVPipelineSingleSample.wdl
```diff
@@ -964,7 +965,19 @@ workflow GATKSVPipelineSingleSample {
...
-    }
+  }
+
+  call SingleSampleMetrics.SingleSampleMetrics as SampleFilterMetrics {
+    input:
+      name = batch,
+      ref_samples = ref_samples,
+      case_sample = sample_id,
+      wgd_scores = Module00b.WGD_scores,
+      sample_counts = select_first([Module00a.coverage_counts]),
```
mwalker174, line 976: `sample_counts = case_counts_file_`

mwalker174 (approved):
Looks good, just caught one thing.

wdl/GATKSVPipelineSingleSample.wdl
cwhelan: Fixed, thanks for catching that.

PR #106, opened 2021-02-01
Standardize SVTYPE nomenclature for INS variants after minGQ

RCollins13:
[A recent PR](https://github.com/broadinstitute/gatk-sv/pull/93) split out mobile element insertions (MEIs) from non-MEI insertions during minGQ filtering, as the properties of MEIs and non-MEIs appeared to be different enough to warrant separate treatment during minGQ. This was accomplished by rewriting `SVTYPE` with the `ALT` allele for MEIs such that minGQ would treat them as a separate SV type during filtering.

However, this `SVTYPE` rewrite is never corrected after the filter is applied, meaning that the current output of minGQ includes records with multipartite insertion `SVTYPE`s, such as `SVTYPE=INS:ME:SVA`. This behavior is undesirable because our other filtering, annotation, and analysis workflows downstream of minGQ assume all insertion variants should have `SVTYPE=INS`, which is no longer the case.

In this PR, I have modified `apply_minGQ_filter.py` and `minGQ.wdl` to include a `--simplify-INS-SVTYPEs` option, which simplifies all SVTYPEs containing the substring `INS` after applying the minGQ filter cutoffs.

### DIFF
--- a/src/sv-pipeline/scripts/downstream_analysis_and_filtering/apply_minGQ_filter.py
+++ b/src/sv-pipeline/scripts/downstream_analysis_and_filtering/apply_minGQ_filter.py
@@ -357,6 +357,9 @@ def main():
     parser.add_argument('--dropEmpties', default=False, action='store_true',
                         help='After GT reassignments, drop any SV with no remaining ' + 
                         ' non-ref samples (default: keep all SV).')
+    parser.add_argument('--simplify-INS-SVTYPEs', default=False, action='store_true',
+                        help='Resets the SVTYPE of all INS variants, including MEIs, ' + 
+                        'to be SVTYPE=INS (default: keep original SVTYPEs).')    
     parser.add_argument('--maxNCR', help='Max no-call rate among all ' + 
                         'samples before adding a flag to the record\'s FILTER field' + 
                         ' (default: 0.005)', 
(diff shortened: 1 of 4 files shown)


src/sv-pipeline/scripts/downstream_analysis_and_filtering/apply_minGQ_filter.py
```diff
@@ -414,7 +417,11 @@ def main():
             for key in 'AN AC AF N_BI_GENOS N_HOMREF N_HET N_HOMALT FREQ_HOMREF FREQ_HET FREQ_HOMALT'.split(' '):
                 if key in record.info.keys():
                     record.info.pop(key)
-        
+
+        # Standardize SVTYPE for all INS variants, if optioned
+        if args.simplify_INS_SVTYPEs and 'INS' in record.info['SVTYPE']:
+            record.info['SVTYPE'] = 'INS'
```
mwalker174, line 423: Might be better to do `and 'INS' in record.info['SVTYPE'].split(':')` - just a suggestion

mwalker174 (approved):
Looks good, one very minor suggestion. Did you test this both with and without the new flag? I can build/update dockers when you think it's ready and I update your branch and merge.

PR #107, opened 2021-02-02
Tiny terra workspace fix

cwhelan:
Removes an obsolete reference to a copy of the single sample qc definitions that was mistakenly packaged with the hg38 reference resources (as opposed to the reference panel).


epiercehoffman (approved):
Looks good, produces the right files, workspace data uploads to Terra correctly

PR #108, opened 2021-02-03
Rewrite Module04b.GetMedianSubset in python for memory scalability

epiercehoffman:
### Changes
* Rewrite Module04b.GetMedianSubset in python for memory scalability. Also adds explicit handling of variants with empty coverage data, and allows for first variant in file to be added to regenotyping list (previously ignored and treated as "header").
* Handle case where filtering in CombineReassess.MergeList removes all regenotyped variants - similar to earlier in the module, processing will stop if this is the case, and the number of filtered variants will be returned, and the output file will point to the input VCF from 04
* Remove Genotype_3 import as it is not used - should it be deleted from the repo as well?
* Fix spacing in several places
* Add top-level runtime attributes for several tasks

### Testing
* Ran test_large data through Module04b with latest master & my changes - both were successful. I compared the output of Module04b.GetMedianSubset and found that my version contains one more line - the first variant, which appears to have erroneously ignored in the master version.
* Tested one gnomAD batch and saw a massive performance improvement: 6 minutes with max 0.64 GB RAM, compared to 30 minutes using 13.5 GB RAM with the latest master version. Latest master version also produced a file of NAs due to unknown errors, while new version produces a file with the expected format. Note that the WDL used for testing was slightly modified from the one on this branch because 04b expects the full cohort rather than a partial cohort
* Tested with 6 gnomAD batches successfully

### DIFF
--- a/wdl/Module04b.wdl
+++ b/wdl/Module04b.wdl
@@ -193,22 +193,25 @@ workflow Module04b {
         runtime_attr_merge_list_creassess = runtime_attr_merge_list_creassess,
         runtime_attr_vcf2bed = runtime_attr_vcf2bed
     }
-      
-    scatter (i in range(length(Genotype_2.genotyped_vcf))) {
-      call ConcatRegenotypedVcfs{
-        input:
-          depth_vcf=depth_vcfs[i],
-          batch = batches[i],
-          regeno_vcf = Genotype_2.genotyped_vcf[i],
-          regeno_variants = CombineReassess.regeno_variants,
-          runtime_attr_override = runtime_attr_concat_regenotyped_vcfs,
-          sv_pipeline_docker = sv_pipeline_docker
+    
+    if (CombineReassess.num_regeno_filtered > 0) {
+      scatter (i in range(length(Genotype_2.genotyped_vcf))) {
+        call ConcatRegenotypedVcfs {
+          input:
+            depth_vcf=depth_vcfs[i],
+            batch = batches[i],
+            regeno_vcf = Genotype_2.genotyped_vcf[i],
+            regeno_variants = CombineReassess.regeno_variants,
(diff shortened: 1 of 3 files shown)


wdl/Module04b.wdl
```diff
@@ -558,9 +559,32 @@ task GetMedianSubset{
...
+    with open("~{medians}", 'r') as inp, open("~{batch}_to_regeno.bed", 'w') as outp:
+      for line in inp:
+        fields = line.strip().split('\t')
+        # first 4 fields are variant info (chr, start, end, varID)
+        var_info = fields[0:4]
+        # next len-4 fields are sample coverage medians for each sample in the batch
+        try:
+          coverage_medians = [float(x) for x in fields[4:]]
+        except ValueError as e:
+          if str(e) == "could not convert string to float: '.'":
```
mwalker174, line 576: Matching an error message to a string like this is risky in case it changes with an update to the statistics package or one of its dependencies. It would be better to check for the presence of missing data explicitly:
```
if any([x == '.' for x in fields[4:]):
  print("Empty coverage data for variant %s at %s:%s-%s" % (var_info[3], var_info[0], var_info[1], var_info[2]))
  continue
else:
  med = median([float(x) for x in fields[4:]])
  ...
```

wdl/Module04b.wdl
```diff
@@ -558,9 +559,32 @@ task GetMedianSubset{
...
+            # if missing data, print variant ID and chrom:start-end in case follow-up desired
+            print("Empty coverage data for variant %s at %s:%s-%s" % (var_info[3], var_info[0], var_info[1], var_info[2]))
```
mwalker174, line 578: I'm slightly concerned this could generate a lot of unnecessary stdout that would end up in the execution bucket.

mwalker174 (approved):
👍

PR #115, opened 2021-02-19
add AWS support for DeleteIntermediateFiles task

henriqueribeiro:
Add support for AWS in `DeleteIntermediateFiles` task in `Module00a`.

### DIFF
--- a/wdl/Module00a.wdl
+++ b/wdl/Module00a.wdl
@@ -358,7 +358,7 @@ task DeleteIntermediateFiles {
   input {
     Array[File] intermediates
     Array[File]? dummy # Pass in outputs that must be complete before cleanup
-    String cloud_sdk_docker
+    String cloud_sdk_docker # Cloud provider SDK docker image. For GCP use "google/cloud-sdk" and for AWS use "amazon/aws-cli"
   }
   parameter_meta {
     intermediates: {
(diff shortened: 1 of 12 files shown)


wdl/Module00a.wdl
```diff
@@ -356,6 +358,7 @@ task DeleteIntermediateFiles {
   input {
     Array[File] intermediates
     Array[File]? dummy # Pass in outputs that must be complete before cleanup
+    String cloud_sdk_docker
```
mwalker174, line 361: Can you add a comment here suggesting which images to use for GCS and AWS?

mwalker174 (changes requested):
Thanks for this @henriqueribeiro ! We do need to have the input templates in `/test_input_templates` updated as well (`batch`, `module00a`, and `single-sample` I believe). You can validate that all generated inputs are valid using `scripts/test/validate.sh`.

henriqueribeiro:
I believe we need to update the `scripts/test/validate.sh`. It seems outdated and not working correctly. Do you confirm it needs some work?

mwalker174:
@henriqueribeiro It appears the script is working correctly, but there are 2 errors in the WDLs that cause it to fail immediately. I've created PR #118 to fix that.

henriqueribeiro:
@mwalker174 all tests pass now :)

mwalker174 (approved):
Great, looks good to me!

PR #116, opened 2021-02-22
Module 07 upgrade

RCollins13:
This PR covers a reworking of several features of Module07 necessary to acommodate the switch in multiallelic CNV representation from SVTYPE=MCNV with <CN=X> alts (as in gnomAD-SV v2.1) to SVTYPE=CNV with <CNV> alts (as in the current gatk-sv repo).

Specifically, this PR includes the following changes:

1. Modified `CalcAF.wdl` and `compute_AFs.py` to compute copy-number frequencies instead of allele frequencies for all multiallelic CNVs. The representation included in this PR was discussed on Broad Slack with Chris and Xuefang.

2. Reintroduce handling of pseudoautosomal regions and hemizygous male samples for sex chromosomes when calculating and annotating variant frequency for bialellic sites. This was functionality that existed at the time of gnomAD-SV v2.1 but appears to have been lost during the codebase refactoring and migration over to the current gatk-sv repo. However, the optional `par_bed` input in `Module07XfBatchEffect.wdl` still referenced the `--par` option for `compute_AFs.py` (but this would have/should have produced an error as `--par` was not an option recognized by the existing version of the script).

3. Modify other Module07 WDLs and scripts to handle the new representation of multiallelic CNVs and their copy-number frequencies.

4. Modify `Module07MinGQ.wdl` to always produce a table of allele frequencies per variant, which is a required input to `Module07XfBatchEffect.wdl` but currently was only being generated when `MingqTraining` was `True`.

5. Remove `minGQ.wdl`, as it has been superseded by `Module07MinGQ.wdl` and may introduce confusion if both are kept in the repo.

6. Remove the unnecessary step of cutting multiallelic CNVs out of the input VCF and reintroducing them at the end of the `Module07FilterCleanupQualRecalibration.wdl` workflow. These steps were unnecessary and added additional compute time due to concatenating and re-sorting the VCF. I have modified the `filter_cleanup_and_QUAL_recalibration.PCRMinus_only.py` to handle our new multiallelic CNV representation accordingly.

7. Modify `find_batch_effects.shard_helper.R` to handle cohorts with exactly two batches, which previously caused a dataframe dimensionality error when comparing allele frequencies between batches.

8. Update `sv-pipeline` and `sv-base-mini` Dockers in `dockers.json`

I have tested all of these changes locally and on the 1000Genomes reference panel of 156 samples starting from the output of module 05/06. I have confirmed that all of the four main WDLs in module07 now operate as expected.

Thanks!


src/sv-pipeline/05_annotation/scripts/compute_AFs.py
```diff
@@ -181,11 +308,15 @@ def main():
         males = [line.split('\t')[1] for line in famfile if line.split('\t')[4] == '1']
         females = [line.split('\t')[1] for line in famfile if line.split('\t')[4] == '2']
         sexes = 'MALE FEMALE'.split()
+        if args.par is not None:
```
cwhelan, line 311: Would it make sense to intialize `parbt` to an empty BedTool, ie `parbt = pbt.BedTool("", from_string=True)`? Then you wouldn't have to check if it's `None` around the script, you could just always intersect with it. Might simplify the code a little.

src/sv-pipeline/05_annotation/scripts/compute_AFs.py
```diff
@@ -199,54 +330,87 @@ def main():
...
     '##INFO=<ID=FREQ_HET,Number=1,Type=Float,Description="Heterozygous genotype frequency (biallelic sites only).">',
-    '##INFO=<ID=FREQ_HOMALT,Number=1,Type=Float,Description="Homozygous alternate genotype frequency (biallelic sites only).">'
+    '##INFO=<ID=FREQ_HOMALT,Number=1,Type=Float,Description="Homozygous alternate genotype frequency (biallelic sites only).">',
+    '##INFO=<ID=CN_NUMBER,Number=1,Type=Integer,Description="Total number of individuals with estimated copy numbers (multiallelic CNVs only).">',
+    '##INFO=<ID=CN_COUNT,Number=.,Type=Integer,Description="Number of individuals observed at each copy state, starting from CN=0 (multiallelic CNVs only).">',
+    '##INFO=<ID=CN_FREQ,Number=.,Type=Float,Description="Frequency of individuals observed at each copy state, starting from CN=0 (multiallelic CNVs only).">',
+    '##INFO=<ID=CN_NONREF_COUNT,Number=1,Type=Integer,Description="Sum of all individuals with non-reference copy states (multiallelic CNVs only).">',
+    '##INFO=<ID=CN_NONREF_FREQ,Number=1,Type=Float,Description="Total frequency of all individuals across all non-reference copy states (multiallelic CNVs only).">'
```
cwhelan, line 347: To make the descriptions of these two tags a little more consistent, I'd change them to,

```
'##INFO=<ID=CN_NONREF_COUNT,Number=1,Type=Integer,Description="Number of samples with non-reference copy states (multiallelic CNVs only).">'
`##INFO=<ID=CN_NONREF_FREQ,Number=1,Type=Float,Description="Frequency of samples with non-reference copy states (multiallelic CNVs only).">'
```

src/sv-pipeline/05_annotation/scripts/compute_AFs.py
```diff
@@ -199,54 +330,87 @@ def main():
...
     '##INFO=<ID=N_HOMREF,Number=1,Type=Integer,Description="Number of individuals with homozygous reference genotypes (biallelic sites only).">',
     '##INFO=<ID=N_HET,Number=1,Type=Integer,Description="Number of individuals with heterozygous genotypes (biallelic sites only).">',
     '##INFO=<ID=N_HOMALT,Number=1,Type=Integer,Description="Number of individuals with homozygous alternate genotypes (biallelic sites only).">',
     '##INFO=<ID=FREQ_HOMREF,Number=1,Type=Float,Description="Homozygous reference genotype frequency (biallelic sites only).">',
     '##INFO=<ID=FREQ_HET,Number=1,Type=Float,Description="Heterozygous genotype frequency (biallelic sites only).">',
-    '##INFO=<ID=FREQ_HOMALT,Number=1,Type=Float,Description="Homozygous alternate genotype frequency (biallelic sites only).">'
+    '##INFO=<ID=FREQ_HOMALT,Number=1,Type=Float,Description="Homozygous alternate genotype frequency (biallelic sites only).">',
+    '##INFO=<ID=CN_NUMBER,Number=1,Type=Integer,Description="Total number of individuals with estimated copy numbers (multiallelic CNVs only).">',
```
cwhelan, line 343: It feels like it would be more consistent to replace `individuals` with `samples` in all of these descriptions.

src/sv-pipeline/05_annotation/scripts/compute_AFs.py
```diff
@@ -29,47 +34,144 @@ def create_pop_dict(popfile):
     return pop_dict
 
 
-def gather_allele_freqs(record, all_samples, males, females, pop_dict, pops, no_combos = False):
+def _is_biallelic(record):
```
cwhelan, line 37: This method is repeated a few times in this PR, should we move it to a single copy in `svtk.utils`?

src/sv-pipeline/scripts/downstream_analysis_and_filtering/clean_frequencies_table.R
```diff
@@ -11,56 +11,52 @@
...
-    female.AN.idx <- which(colnames(dat)==paste(pop,"FEMALE","AN",sep="_"))
-    #For variants on chrX, overwrite master counts with female-specific counts
-    dat[x.idx,master.AC.idx] <- dat[x.idx,female.AC.idx]
-    dat[x.idx,master.AN.idx] <- dat[x.idx,female.AN.idx]
-    #For all variants on chrY, overwrite master counts with male-specific counts
-    dat[y.idx,master.AC.idx] <- dat[y.idx,male.AC.idx]
-    dat[y.idx,master.AN.idx] <- dat[y.idx,male.AN.idx]
+    if(length(c(x.idx, y.idx)) > 0){
+      if(4 == length(intersect(colnames(dat),
+                               c(paste(pop,"MALE","AC",sep="_"),
```
cwhelan, line 39: A tiny nitpick, but It'd be a little cleaner to construct these names once, before the if statement, rather than constructing them all twice.

src/sv-pipeline/scripts/downstream_analysis_and_filtering/make_batch_effect_reclassification_table.PCRMinus_only.R
```diff
@@ -34,12 +33,14 @@ import.freqs <- function(freq.table.in){
   out <- data.frame("VID"=dat$VID,"PCRPLUS_AF"=plus.AFs,"PCRMINUS_AF"=minus.AFs)  
   return(out)
 }
+
 #Process failure lists for a single comparison type
-import.fails <- function(all.in,minus.in,prefix){
-  a <- read.table(all.in,header=F,sep="\t")
+import.fails <- function(minus.in,prefix){
+  a <- tryCatch(read.table(minus.in,header=F,sep="\t"), error=function(e){return(as.data.frame(matrix(ncol=2)))})
```
cwhelan, line 39: Is this to guard against a non-existent file? Would there be any way to check for that explicitly? It seems dangerous to swallow all errors here -- what if there's a transient file system error or something?

cwhelan (changes requested):
Hi Ryan, this looks pretty good. I have some suggestions for how to clean up some of the code, and one question about edge-case logic. Feel free to go through the suggestions and implement them or defer them as you see fit.

src/sv-pipeline/05_annotation/scripts/compute_AFs.py
```diff
@@ -14,6 +14,11 @@
 import pysam
 import numpy as np
 from svtk import utils as svu
+from collections import Counter
+import pybedtools as pbt
+
+
+default_sex_chroms = 'X Y chrX chrY'.split()
```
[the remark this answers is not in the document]
RCollins13: Sounds good to me

src/sv-pipeline/05_annotation/scripts/compute_AFs.py
```diff
@@ -88,6 +190,12 @@ def calc_allele_freq(record, samples, prefix = None):
             if len([allele for allele in GT if allele != 0 and allele != '.' and allele is not None]) == 2:
                 nhomalt += 1
 
+        #Adjust hemizygous allele number and allele count, if optioned
+        if hemi:
+            AN = AN / 2
+            # For hemizygous sites, AC must be the sum of all non-reference *genotypes*, not alleles
+            AC = nhet + nhomalt
```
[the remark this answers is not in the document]
RCollins13: Fine with me. Will make these changes.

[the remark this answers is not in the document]
RCollins13: This is a good point. The problem stems from the fact that our genotyper isn't ploidy-aware (i.e., I don't think we would ever see a `1/.` genotype produced), but we don't want hemizygous sites to count twice towards AC when they're genotyped as homozygous (which I recall happening fairly frequently in XY samples on allosomes).

The solution I've implemented in a revised version of the code (commit pending) is to define `n_gts_with_gt_0_alts` as a designation for the count of genotypes with at least one non-reference allele, and then set `AC` equal to this value for hemizygous sites.

Does that sound like it would handle the edge case you brought up?

src/sv-pipeline/05_annotation/scripts/compute_AFs.py
```diff
@@ -163,6 +288,8 @@ def main():
                         default = None)
     parser.add_argument('--no-combos', help='Do not compute combinations of populations ' +
                         'and sexes.', action = 'store_true', default = False)
+    parser.add_argument('--par', help='BED file of pseudoautosomal regions (used ' +
```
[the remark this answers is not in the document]
RCollins13: Good suggestion; done.

src/sv-pipeline/05_annotation/scripts/compute_AFs.py
RCollins13: Sure, that can work, although I'd prefer to only add the `PAR` definition to the VCF header if a file is passed as `--par`, so we'll still have to check if `len(parbt) > 0` instead of `par is not None:`. Agreed though this is a simpler solution, so I've made those changes.

src/sv-pipeline/05_annotation/scripts/compute_AFs.py
RCollins13: Good point—done.

src/sv-pipeline/05_annotation/scripts/compute_AFs.py
RCollins13: Sure, that seems fine to me. "Samples" will be more appropriate than "individuals" for the case of callsets with duplicated samples or technical replicates, I suppose.

RCollins13: Sure, that seems fine to me. "Samples" will be more appropriate than "individuals" for the case of callsets with duplicated samples or technical replicates, I suppose.

src/sv-pipeline/05_annotation/scripts/compute_AFs.py
RCollins13: Another good idea; done.

src/sv-pipeline/scripts/downstream_analysis_and_filtering/make_batch_effect_reclassification_table.PCRMinus_only.R
RCollins13: Yep, this was included to protect against the situation where zero records are marked as having batch effects. This is unlikely to happen in most real-world multi-batch callsets, but did happen when I was testing on the ref panel when split into two random batches.

Good point about the risk of error swallowing here, though. I've removed the `tryCatch()` statement and instead implemented a check of the input file size from within the R script itself:
```
  # Check file size and return an empty dataframe if no records are marked as batch effects
  if(file.info(minus.in)$size == 0){
    a <- as.data.frame(matrix(ncol=2))
  }else{
    a <- read.table(minus.in,header=F,sep="\t")
  }
```

I thought about moving this logic out to the WDL level, and bypass the relevant steps if no variants are marked as batch effects... but decided against it due to the possible situation where one of the two batch effect lists is empty (but the other isn't). Probably cleanest just to check each file within the R script.

src/sv-pipeline/scripts/downstream_analysis_and_filtering/clean_frequencies_table.R
RCollins13: No problem — done.

RCollins13:
Thanks for the thorough feedback, @cwhelan!

I've implemented the changes you requested; they're in the most recent commit to this branch if you want to take a look.

I'm still testing these changes to make sure everything works as expected, and will comment again on this PR once it passes all module07 workflows.

Thanks again!

RCollins13:
Alright @cwhelan I can confirm that all of the updated code has passed a test run on the reference panel cohort and should be good to merge provided there aren't any other modifications you'd like. Let me know!

src/sv-pipeline/05_annotation/scripts/compute_AFs.py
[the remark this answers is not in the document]
cwhelan: Sure, that sounds like a good solution! Again, I didn't think this was an actual problem that would cause a bug in the pipeline right now; just that the algorithm for counting alts seemed to be a little inconsistent between the two cases -- which is sometimes the sort of thing that causes bugs down the road.

cwhelan (approved):
Looks good, thanks for the updates. 👍

PR #117, opened 2021-02-26
Make outlier cutoff table input to 03 optional

epiercehoffman:
### Updates
Make outlier cutoff table input to Module03 optional. This is because the cutoff table should really be specific to the cohort, but it currently must be generated outside of the canonical pipeline, and the "default" 1KG table that we use for testing is very permissive so it doesn't impact filtering to use it. Therefore, having an optional input (rather than a no-impact default value for a required input) seems to make it clearer to the user that they have to provide their own table in order for this method of filtering to be useful. With or without a cutoff table, the IQR filtering will still be performed.

### Testing
This branch was tested with and without a cutoff table on `test_large` data and Module03 completed successfully in both cases. The outputs were identical, because the permissive default table was used for testing, but I verified that the cutoff table task did run when the cutoff table was provided (and did not run without it).

### DIFF
--- a/wdl/FilterOutliers.wdl
+++ b/wdl/FilterOutliers.wdl
@@ -24,7 +24,7 @@ workflow FilterOutlierSamples {
     Array[String] samples
     Array[String] algorithms
     Int N_IQR_cutoff
-    File outlier_cutoff_table
+    File? outlier_cutoff_table
     String sv_pipeline_docker
     String sv_base_mini_docker
     String linux_docker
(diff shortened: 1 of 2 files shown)


mwalker174 (approved):
Thanks! Looks good!

jingydz:
May I ask how to make the outlier cutoff table for the specific cohort? Also, what was the format of the outlier cutoff table originally provided? What did it look like?

epiercehoffman:
@jingydz You can see the example files we provide for our test data sets in our [input values files](https://github.com/broadinstitute/gatk-sv/tree/master/input_values). The example outlier cutoff table is located at this URI: `gs://gatk-sv-resources-public/hg38/v0/sv-resources/ref-panel/1KG/v1/module03_outlier_cutoff_table.tsv`.  It is small and in a public google bucket, so you should be able to view it easily and copy the format. As you can see, it describes upper and lower SV count outlier cutoffs, split by algorithm (ie. depth, manta) and SV type (ie. DUP, DEL). These outlier cutoffs would depend on your data and how strict you want to be, so they would need to be determined manually by examining your data. If you choose to provide this table, samples will be counted as outliers and filtered during Module03 if their SV counts for a given type and algorithm are lower than the lower cutoff or higher than the upper cutoff.

PR #119, opened 2021-02-26
Fix build_inputs.py

henriqueribeiro:
`build_inputs.py` fails if input paths starts with `/`. This bug is caused by the split that is done in `make_target_subdir`. 
This PR replaces this function by python built-in function to create directories.


cwhelan (approved):
Sorry for the delay in review, thanks for the fix @henriqueribeiro.

PR #121, opened 2021-02-26
Fix paths on validate script

henriqueribeiro:
`scripts/inputs/build_default_inputs.sh` is creating inputs files in a different directory from the one on the validate script


mwalker174 (approved):
Thank you @henriqueribeiro - I have not cleaned out my repo in some time and the validate script had been picking up inputs from old directory structure. Looks good!

PR #130, opened 2021-03-16
Small bug and aesthetic fixes for quota script

epiercehoffman:
### Updates
* Previously, if the job ID was missing for a task, the script broke. Modified to just continue processing regardless
* Previously, cached tasks may have been counted as non-preemptible in rare cases when information about the number of attempts was missing. Modified to automatically mark cached tasks as preemptible in order to avoid inflating the non-preemptible VMs count
* Various aesthetic fixes

### Testing
Tested with an array of small, large, successful, failed, clean, cached, and broken metadata files

### DIFF
--- a/scripts/cromwell/analyze_resource_acquisition.py
+++ b/scripts/cromwell/analyze_resource_acquisition.py
@@ -65,11 +65,13 @@ def get_disk_info(metadata):
     return float(0), float(0)
 
 
-def was_preemptible_vm(metadata):
+def was_preemptible_vm(metadata, was_cached):
   """
-  Source: https://github.com/broadinstitute/dsde-pipelines/blob/develop/scripts/calculate_cost.py
+  Modified from: https://github.com/broadinstitute/dsde-pipelines/blob/develop/scripts/calculate_cost.py
   """
-  if "runtimeAttributes" in metadata and "preemptible" in metadata['runtimeAttributes']:
+  if was_cached:
+    return True # if call cached, not any type of VM, but don't inflate nonpreemptible count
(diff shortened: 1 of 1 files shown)


scripts/cromwell/analyze_resource_acquisition.py
```diff
@@ -90,11 +92,12 @@ def calculate_start_end(call_info, override_warning=False, alias=None):
...
-  if alias is None or alias == "":
-    alias = job_id
-  else:
-    alias += "." + job_id
+  if 'jobId' in call_info:
+    job_id = call_info['jobId'].split('/')[-1]
+    if alias is None or alias == "":
+      alias = job_id
+    else:
+      alias += "." + job_id
```
mwalker174, line 100: So if `jobId` is absent, `alias` may end up at `None`, although it seems that `alias` is only used for a warning message in this script. Perhaps add an `else` to the first `if` to assign `alias` explicitly to something like `NA`. It may also be helpful to emit a warning explaining what situations can cause this to happen.

mwalker174 (approved):
Looks good to me, just with one suggestion.

scripts/cromwell/analyze_resource_acquisition.py
epiercehoffman: I don't think I've ever encountered a scenario where alias is None at this point (and I don't think it's really possible within the context of the larger script), which is why I didn't bother with that condition, but I've added a catch just in case

TedBrookings: The new change leaves no path for alias to be None, so I think it solves the issue.

TedBrookings (approved):
Latest commit solves the one remaining issue

PR #133, opened 2021-03-17
Module04b: parametrize sample overlap and minimum var count outlier threshold for regeno filtering

epiercehoffman:
### Updates
* In Module04b.CombineReassess.MergeList, do not count sample as outlier for regenotyped variant count unless it has >3 such variants (and add optional parameter to change that threshold)
* In the same task, add a parameter for the sample overlap threshold (between samples with raw calls & samples with regenotyped calls), which is set to 0.7 by default

### Testing
Tested on test_large, 1 gnomAD batch, and 30 gnomAD batches with default parameters. The number of regenotyped variants was unchanged in each case.

### DIFF
--- a/wdl/Module04b.wdl
+++ b/wdl/Module04b.wdl
@@ -26,6 +26,8 @@ workflow Module04b {
     Array[File] regeno_coverage_medians # one file per batch
     Float regeno_max_allele_freq = 0.01 
     Int regeno_allele_count_threshold = 3 
+    Int min_var_per_sample_outlier_threshold = 3
+    Float regeno_sample_overlap = 0.7
 
     RuntimeAttr? runtime_attr_cluster_merged_depth_beds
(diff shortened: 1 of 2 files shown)


wdl/Module04b.wdl
```diff
@@ -26,6 +26,8 @@ workflow Module04b {
     Array[File] regeno_coverage_medians # one file per batch
     Float regeno_max_allele_freq = 0.01 
     Int regeno_allele_count_threshold = 3 
+    Int min_var_per_sample_outlier_threshold = 3
+    Float regeno_sample_overlap = 0.7
```
mwalker174, line 30: Add a quick comment here explaining what these are

mwalker174 (approved):
Looks good! Just one small request.

TedBrookings (approved):
Emma's comment definitely supplies the requested comment, so approved.

PR #139, opened 2021-04-02
add -w flag to grep for varID searches in 04b

epiercehoffman:
### Updates
Add -w flag to grep in several places in Module04b when extracting variants by variant ID, to avoid matching variant IDs on a substring and getting more variants than expected. 

### Testing
Tested successfully on `test_large` data. This resulted in a change from 85 to 46 filtered regenotyped variants at the end (despite starting with the same number of original regenotyping candidates, 408).


mwalker174 (approved):
Approved, conditioned on gnomAD QC.

PR #143, opened 2021-04-07
Update clean_vcf_part1 to fix bugs in setting FORMAT EV tags

cwhelan:
A bug in bcftools 1.7 was preventing evidence annotations from being added to some mCNV records in clean_vcf_part1. 

This bug seems to be fixed in bcftools 1.9.

In our sv-pipeline dockers, conda pulls in bcftools 1.7 as a fixed dependency of pysam and puts it on the path, but we have an installation of bcftools 1.9 in /usr/local/bin. This PR changes clean_vcf_part1.sh to use the version 1.9 installation of bcftools.


mwalker174 (approved):
Thanks! This looks good 👍

PR #146, opened 2021-04-12
Use single ped file for cohort rather than batch ped files

epiercehoffman:
### Updates
* add util task to subset ped file to list of batch samples
* subset ped file at the beginning of each batch-level workflow
* remove ped file filtering from Module03
* remove ped file merging from Module0506
* change famfile to ped_file for several modules
* eliminate sample lists input from 04b - get from VCF
* add 04b to Batch WDL
* update JSON templates as necessary
* add back lost test_large samples to input values JSON
* pass runtime_attr_baf_gen to BAFFromShardedVCF in 00c

### Testing
Successful tests with `test_large` data:
* SingleSampleTest
* Phase1 (with extra samples added to ped file)
* Module04 (with extra samples added to ped file)
* Module04b (with extra samples added to ped file)
* Module0506 (with extra samples added to ped file)

Still working on testing Batch WDL due to an unrelated CramToBam error.
Also validated all test JSONs. 
Tested efficiency by running SubsetPedFile task for one ~500-sample batch on full gnomAD cohort ped file - finished in <3 seconds with normal resource settings.


mwalker174 (approved):
Thank you @epiercehoffman, the batched ped files have caused a lot of headaches in the past. I do think we should still subset the ped in single-sample mode, see comment below, in case users do not subset their ped files for custom reference panels.

PR #149, opened 2021-04-15
Add option to randomly subsample samples for trainGCNV (for Terra)

epiercehoffman:
### Updates
Add optional parameter to trainGCNV to randomly subsample a subset of samples to use for gCNV training, rather than the full sample list passed in. This feature will be useful for Terra to allow per-batch gCNV training without having to train on every sample in the batch. There is a seed parameter with a default value set in the WDL, and usage of the random subsampling parameter requires passing an additional docker image (for Python). 

### Testing
Tested with and without random subsampling with `test_small` dataset.

### DIFF
--- a/wdl/Utils.wdl
+++ b/wdl/Utils.wdl
@@ -119,3 +119,64 @@ task RunQC {
   }
 
 }
+
+task RandomSubsampleStringArray {
+  input {
+    Array[String] strings
+    Int seed
+    Int quantity
+    String prefix = "prefix"
+    String sv_pipeline_base_docker
+    RuntimeAttr? runtime_attr_override
+  }
+
+  String subsample_indices_filename = "~{prefix}.subsample_indices.list"
+  String subsampled_strings_filename = "~{prefix}.subsampled_strings.list"
+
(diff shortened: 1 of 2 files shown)


wdl/Utils.wdl
```diff
@@ -119,3 +119,64 @@ task RunQC {
   }
 
 }
+
+task RandomSubsampleStringArray {
+  input {
+    Array[String] strings
+    Int seed
+    Int quantity
+    String prefix = "prefix"
```
mwalker174, line 128: Would you ever want to use this default value? I think this should be required.

wdl/Utils.wdl
```diff
@@ -119,3 +119,64 @@ task RunQC {
   }
 
 }
+
+task RandomSubsampleStringArray {
+  input {
+    Array[String] strings
+    Int seed
+    Int quantity
```
mwalker174, line 127: `subset_size` would be a little clearer

mwalker174 (approved):
Thanks @epiercehoffman, looks good to me, just 2 very minor suggestions.

Issue #141, opened 2021-04-05
GCloud statement


henriqueribeiro:
### Affected module(s) or script(s)
- MakeBincovMatrix

### Affected version(s)
- [x] Latest public release version [version?]
- [ ] Latest master branch as of [date of test?]

### Description 
Is this line needed in the task?
https://github.com/broadinstitute/gatk-sv/blob/b8dbf083895c0623227050bda7d26b26e54fba5b/wdl/MakeBincovMatrix.wdl#L263

I'm using the AWS backend and, as expected, it fails at this line.

jingydz:
I've had this problem before, but just delete the sentence and it will work.

mwalker174:
We still use this command in a few places, most notably CramToBam (work around by enabling requester pays option). Closing, however, since do not have plans at this time to support AWS.

PR #152, opened 2021-04-23
Remove GCS authentication from ZPaste in MakeBincovMatrix

mwalker174:
Addresses #141.


cwhelan (approved):
Good call to remove the unnecessary GCS authorization -- did you mean to include the runtime attribute adjustments in CleanVcf4, or do those go with another PR you're working on?

PR #155, opened 2021-04-26
integrate Xuefang's CramToBam ReviseBase and annotation changes

epiercehoffman:
### Updates
* add option to perform ReviseBase task during CramToBam in Module00a
* optimize default resources for CramToBam ReviseBase version
* fix SplitCramToBamRequesterPays by localizing cram index
* add new required Module08 inputs to Module08 and single-sample JSON templates

### Testing
* Tested `Module00aBatch` workflow with two samples using revise base option, with and without requester pays mode
* Optimized resources after testing
* Validated all JSONs and WDLs


mwalker174 (approved):
Thanks @epiercehoffman, this all looks good. We should rewrite `ReviseBaseInBam` at some point - perhaps using GATK.

PR #157, opened 2021-04-30
Remove greps in clean_vcf_part1 which could match sample IDs to variant IDs

cwhelan:
This fixes a problem in `clean_vcf_part1.sh` which could cause problems in the calculation of median with-sex depth for allosome CNVs if a sample ID was present in the batch name. The root cause was a grep statement based on sample IDs which could match variant IDs. I've rewritten the bash pipe in question to use bcftools instead of grep/awk/vcftools. Also passed the allosome fai down into this task so that we are less dependent on hard coded searches for X and Y chromosome names, although there are still multiple places in this script where that is the case. Tested this out in a full single sample pipeline run and confirmed that a missing, clinically important chr X duplication in a male sample is now correctly called with this change.


mwalker174 (approved):
👍 Looks good to merge. Should users still be avoiding sample IDs in the batch name?

cwhelan:
As far as I know there aren't any other bugs related to this, but that doesn't rule out the possibility, so it's probably safest not to put sample IDs in the batch name. However, it's a natural thing to want to do when running the single sample pipeline. Perhaps we should just make the pipeline create a batch name automatically (maybe something like `GATKSVSingleSample_{TIMESTAMP}`?) to avoid any chance of something like this again.

PR #158, opened 2021-05-03
Add Dockerfile for ExpansionHunter Denovo.

VJalili:
This PR adds a Dockerfile for [ExpansionHunter Denovo](https://github.com/Illumina/ExpansionHunterDenovo). The resulting image is of size `13.9MB` and you may test it as it follows. 

```
$ docker build --no-cache -t ehdn .
$ docker run -it ehdn ExpansionHunterDenovo profile --help
Usage: ExpansionHunterDenovo profile [options]

Available options:
  --help                      Print help message
  --reads arg                 BAM or CRAM file with aligned reads
  --reference arg             FASTA file with reference assembly
  --output-prefix arg         Prefix for the output files
  --min-unit-len arg (=2)     Shortest repeat unit to consider
  --max-unit-len arg (=20)    Longest repeat unit to consider
  --min-anchor-mapq arg (=50) Minimum MAPQ of an anchor read
  --max-irr-mapq arg (=40)    Maximum MAPQ of an in-repeat read
  --log-reads                 Log informative reads
```

Note that at the build time, the latest version of `ExpansionHunter Devono` is pulled from its Github release, and successful builds using future releases depend on if the maintainers are persistent with their currently used norms: (1) a release is distributed in `.tar.gz`; and (2) binaries are stored under `bin` directory. We can define both the archive type and binary directory as arguments, however, for the archive type, in particular, we may need to alternate between different archive extraction programs depending on the archive type.

### DIFF
--- a/dockerfiles/expansion-hunter-denovo/Dockerfile
+++ b/dockerfiles/expansion-hunter-denovo/Dockerfile
@@ -1,10 +1,6 @@
 FROM alpine:latest
 RUN apk --no-cache add curl && \
-    curl -s https://api.github.com/repos/Illumina/ExpansionHunterDenovo/releases/latest \
-    | grep "browser_download_url.*.tar.gz" \
-    | cut -d : -f 2,3 \
-    | tr -d \" \
-    | xargs wget && \
+    wget https://github.com/Illumina/ExpansionHunterDenovo/releases/download/v0.9.0/ExpansionHunterDenovo-v0.9.0-linux_x86_64.tar.gz && \
     mkdir ehdn_extract && \
     tar -xf *.tar.gz --strip-components=1 -C ehdn_extract && \
     rm -rf *.tar.gz && \
--- a/dockerfiles/expansion-hunter-denovo/Dockerfile
+++ b/dockerfiles/expansion-hunter-denovo/Dockerfile
@@ -1,0 +1,14 @@
+FROM alpine:latest
+RUN apk --no-cache add curl && \
+    curl -s https://api.github.com/repos/Illumina/ExpansionHunterDenovo/releases/latest \
+    | grep "browser_download_url.*.tar.gz" \
+    | cut -d : -f 2,3 \
+    | tr -d \" \
+    | xargs wget && \
+    mkdir ehdn_extract && \
+    tar -xf *.tar.gz --strip-components=1 -C ehdn_extract && \
+    rm -rf *.tar.gz && \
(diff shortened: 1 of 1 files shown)


mwalker174 (commented):
Thanks @VJalili this is looking great. Can you peg the release to the current version? Sometimes tool authors introduce updates that cause issues (new bugs, performance decreases, format changes, etc.). 

Then you can simply curl a hard-coded url ie `https://github.com/Illumina/ExpansionHunterDenovo/releases/download/v0.9.0/ExpansionHunterDenovo-v0.9.0-linux_x86_64.tar.gz`

Also did you test running on a BAM/CRAM? Just to be sure there aren't any errors due to the environment.

It looks like there are some python scripts for case/control and outlier analysis, but we should leave that for later if we need them (python would just bloat the docker).

VJalili:
@mwalker174 Thanks for the review. 

> Can you peg the release to the current version? Sometimes tool authors introduce updates that cause issues (new bugs, performance decreases, format changes, etc.).

Certainly, I pinned the version of `ExpansionHunterDenovo` to its current latest version.

> Also did you test running on a BAM/CRAM? Just to be sure there aren't any errors due to the environment.

I had tested on some sample data; however, for a reproducible test, I add `docker-compose.test.yml` (xref: https://github.com/broadinstitute/gatk-sv/pull/158/commits/6ac6307bc583184939dc32b668872f488c8f308e) that you may use it as: 

```
docker-compose -f docker-compose.test.yml run --rm sut
```

This is build following [this guide](https://docs.docker.com/docker-hub/builds/automated-testing/) with possible automated testing in mind, and we can remove the commit if it is introducing too much overhead. 

> It looks like there are some python scripts for case/control and outlier analysis, but we should leave that for later if we need them (python would just bloat the docker).

Yes, that is correct. I was considering only using [`ExpansionHunterDenovo`](https://github.com/Illumina/ExpansionHunterDenovo/blob/master/documentation/05_Computing_profiles.md), but certainly, there are more (such as [this](https://github.com/Illumina/ExpansionHunterDenovo/blob/master/documentation/07_Case_control_analysis.md)).

mwalker174 (approved):
Looks good! As we discussed over DM, we should definitely consider using docker tests in the future, but we should hold off on that until we decide on a testing framework. Feel free to merge after reverting the test commit.

VJalili:
@mwalker174 Agree, we can test on the testing framework as a separate PR.

PR #159, opened 2021-05-06
Set default disk to 10GB for SR, PE, BAF, and RD Test for 02

epiercehoffman:
### Updates
Remove disk scaling and set default disk to 10 GB for SR, PE, BAF, and RD tests in Module02. Disk scaling was determined to be unnecessary, and disk sizes were too large and expensive. Module02 was run with 10 GB for each test for gnomAD, so 10 GB should be an adequate baseline disk size. 

### Testing
WDLs were validated with womtool using the validation script.
Further testing was not performed because of prior resource usage analysis for large batches, and because our test set is quite small and therefore an inadequate stress test.

### DIFF
--- a/wdl/BAFTestChromosome.wdl
+++ b/wdl/BAFTestChromosome.wdl
@@ -81,7 +81,6 @@ task BAFTest {
     String prefix
     String batch
     String sv_pipeline_docker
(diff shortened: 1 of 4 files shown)


mwalker174 (approved):
Thanks! 👍

Issue #7, opened 2020-06-30
Update WDL headers


mwalker174:
These headers can be deleted from all WDLs:
```
##########################################################################################

## Base script:   https://portal.firecloud.org/#methods/Talkowski-SV/00_batch_evidence_merging/15/wdl

## Github commit: talkowski-lab/gatk-sv-v1:<ENTER HASH HERE IN FIRECLOUD>

##########################################################################################
```
For consistency, we should add new headers to all top-level WDLs (`GATKSVPipeline*.wdl`, `Module0*.wdl`, etc) and adopt the GATK style, e.g. [joint genotyping wdl](https://portal.firecloud.org/?return=terra#methods/gatk/joint-discovery-gatk4/14/wdl). Use the description information from the README for each workflow.

epiercehoffman:
Partially addressed by removing outdated headers in #160

mwalker174:
All headers have been removed, so closing this issue in favor of providing relevant info in the documentation.

PR #160, opened 2021-05-06
Remove old WDL headers with outdated firecloud, license, copyright references

epiercehoffman:
### Updates
Remove outdated WDL headers, including:
* References to firecloud
* Copyright statements - now covered by repository license
* License statements - now covered by repository license

Left intact header lines with author information and workflow descriptions.

### Testing
Validated with womtool using validation script.


cwhelan (approved):
Looks good to me. This would address open issue https://github.com/broadinstitute/gatk-sv/issues/7 .

mwalker174 (approved):
This is marvelous! I would make one suggestion to change the author lines from using `Contact` to `Author` so users won't think they should message individuals with issues.

epiercehoffman:
> Looks good to me. This would address open issue #7 .

Thanks for pointing me to the relevant issue! That issue also says we should add updated headers with descriptions of the workflow, but it seems to me like it would be easier to maintain that documentation in just one place - the README. Thoughts? Do you think I should leave issue #7 open to add new headers or close it?

cwhelan:
> Thanks for pointing me to the relevant issue! That issue also says we should add updated headers with descriptions of the workflow, but it seems to me like it would be easier to maintain that documentation in just one place - the README. Thoughts? Do you think I should leave issue #7 open to add new headers or close it?

Eh, I guess you could just comment on that issue that this ticket partially addressed it by removing the old headers. Adding new headers and documentation to the top level WDLs is a bigger job, but might be nice, especially for people who discover them via Dockstore, for example.

PR #164, opened 2021-05-07
Make read length input to CollectWgsMetrics for Melt

epiercehoffman:
### Updates
Add read length as input to CollectWgsMetrics in Module00a.MELT.GetWgsMetrics, in case read length differs from the default 150bp assumed by that program. To accomplish this, GetMultipleMetrics was moved ahead of GetWgsMetrics to allow the calculated read length to be passed from the former to the latter if read length was not supplied as an input.

### Testing
Tested with two samples from `test_small`. Confirmed GetWgsMetrics completed successfully, using a read length input of 151, calculated in GetMultipleMetrics.

### DIFF
--- a/wdl/MELT.wdl
+++ b/wdl/MELT.wdl
@@ -86,28 +86,6 @@ workflow MELT {
       email: "tbrookin@broadinstitute.org"
     }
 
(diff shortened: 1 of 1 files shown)


mwalker174 (approved):
👍 Thanks!

Issue #103, opened 2021-01-28
Improve templating of Terra method configuration JSONS


cwhelan:
In https://github.com/broadinstitute/gatk-sv/pull/96, @epiercehoffman suggested:

> Finally, I do have some worries about template files piling up once we do a similar thing for the Terra cohort pipeline, as it looks like with this model we will need to have a separate Terra-specific template for each workflow. Of course, most of the time we will only have to actually edit the input values rather than the templates, so it's still far better than the previous JSON system. But, it could be a lot of files to edit if inputs are added/names are changed. Could it work to have one template per workflow but tag template items with "this" or "workspace" if that's how they should be configured for Terra, and do away with the separate Terra template? Or is that too complicated with the current JSON-generation implementation / are there cases where it would not work to use the same variable names in Terra?

I couldn't find way to do that in the scope of that PR, but we should investigate ways to make this more efficient and stop the proliferation of different versions of JSON templates. Ideally one could, when executing the `build_inputs.py` script, specify a special "Terra mode" at runtime, which would cause the script to write a reference to `${this.varname}` or `${workspace.varname}` instead of the actual input value. This would allow Terra method configuration JSONs to be generated from the same template as the main subworkflow input JSON.

PR #169, opened 2021-05-18
Add templates for Terra cohort mode JSONs and workspace data

epiercehoffman:
### Updates
Add templates to generate JSONs and workspace data TSV for Terra cohort-mode workspace, in the style of other templates. 

### Testing
Tested JSON & workspace data TSV generation from templates using build_default_inputs.sh script.

### Still to come
I will update this PR or make another (depending on timing) with the dashboard/documentation and the sample data table. I just wanted to stop sitting on these templates while I finalize the documentation.


mwalker174 (approved):
Magnificent. One minor filename correction I found.

Referencing #103 which applies here - the proliferation of json templates isn't ideal. In theory, it seems like we should be able to have only one template per workflow and instead populate separate input value jsons for cromwell and terra. As we discussed offline, however, there are technical snags to making that happen (I believe array inputs were one).

PR #170, opened 2021-05-18
Fix SR_GQ assignment in cleanvcf4

mwalker174:
Fixes an issue where SR_GT was being assigned to SR_GQ in a small number of variants.


cwhelan (approved):
Looks good to me.

PR #171, opened 2021-05-20
Delete unused requirements.txt for igv docker

mwalker174:
File is not used and is triggering dependabot warnings.


epiercehoffman (approved):
Looks good, can't find any references to this file in the repo

PR #173, opened 2021-05-28
Option to rename VCF samples in BAFFromShardedVCF

mwalker174:
Adds `Array[String]? replacement_snp_vcf_sample_ids` input to Module00c. When specified and using the `snp_vcfs` input to generate BAF, each vcf will be reheadered with the given sample IDs. This is useful when sample IDs have been reformatted for gatk-sv but the the vcfs have been already been generated with original IDs. 

Note that the length and order of `replacement_snp_vcf_sample_ids` must exactly match that in the vcfs (or header file if provided), and also that they may be different from the `samples` input (e.g. when the batch samples are a subset of those in the vcf, or the vcf samples are in a different order).


mwalker174:
Closing in favor of a simpler solution. I will open a new PR.

PR #174, opened 2021-05-28
Add MedianCov to Module00b

mwalker174:
Adds the median coverage workflow to Module00b for QC and batching purposes.


epiercehoffman (approved):
Looks pretty straightforward to me!

PR #176, opened 2021-06-02
Evaluate all batch-origin variants in GetRegenotype

epiercehoffman:
### Updates
Changes to evaluate all variants originally discovered in the batch in GetRegenotype in Module04, rather than just the ones that retained the variant ID containing the batch name after MergeCohortVcfs. A chr_start_end_svtype identifier is created and used to match variants that originated in the batch pre-genotyping with potentially-renamed variants post-genotyping so that they can be checked for sample loss/gain.

Other potential bugs were evaluated (space vs. tab delimiters, checking `$6` instead of `$7`) but the task was determined to be working as expected.

### Testing
Tested with two 1kgp batches and compared to a recent run in Terra. The final outputs were identical. There were 18 additional variants in the `to_regeno.bed` file outputted from GetRegenotype for batch2, and 1 additional variant in the regenotype list outputted from MergeList, which was investigated and appeared to fit the expected pattern. But the added variants were all filtered out during CombineReassess so the final regenotyped VCFs were identical.

Further testing will be performed with gnomAD batches to evaluate scaling performance.
Update: With 30 gnomAD batches, GetRegenotype took on average a few seconds longer than before, and used less memory and disk. The last test had 3407 variants to regenotype before filtering and 51 after filtering; this latest test had 3720 variants to regenotype before filtering and 50 after. The increase pre-filtering is expected because missing variants were added. The decrease in final regenotyped variants is less expected, but may have been caused by shifting outlier boundaries during filtering. Alternatively, it may have been caused by a separate change made in between these two tests; unfortunately, the intermediate files for the first test no longer exist for direct comparison.

### DIFF
--- a/wdl/Module04b.wdl
+++ b/wdl/Module04b.wdl
@@ -531,7 +531,7 @@ task GetAndCountCohortSampleList {
   }
 }
 
-task GetRegenotype{
+task GetRegenotype {
   input {
     File depth_genotyped_vcf
     File regeno_sample_counts_lookup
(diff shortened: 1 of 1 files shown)


hab45 (approved):

hab45:
Everything looks good to me

mwalker174 (approved):

PR #179, opened 2021-06-08
Replace King's dead link with a current-alive URL.

VJalili:
I assume we want to pin to `king` version `215`, if so, the currently used link (`http://people.virginia.edu/~wc9c/KING/executables/Linux-king215.tar.gz`) is down. In this PR, I replace it with a currently alive alternative link.


mwalker174 (approved):
Thanks!

PR #180, opened 2021-06-10
Add an option to build images without pushing them to GCR.

VJalili:
This option enables building images without pushing them to a google container registry (GCR). The option can be useful for debugging built images, and in continues-integration/delivery (CI/CD) where it is asserted if images can be successfully built as a result of the changes introduced in a PR without necessarily pushing the images to GCR.

### DIFF
--- a/scripts/docker/build_docker.py
+++ b/scripts/docker/build_docker.py
@@ -74,15 +74,12 @@ class CMD_line_args_parser:
 
         docker_remote_args_group.add_argument('--gcr-project',
                                               type = str,
-                                              help = 'GCR billing project to push the images to')
+                                              help = 'GCR billing project to push the images to. If not given, the '
+                                                     'built image(s) will not be push to GCR.')
 
         docker_remote_args_group.add_argument('--update-latest',
                                               action = 'store_true',
                                               help   = 'also update \"latest\" tag in remote docker repo(s)')
-
-        docker_remote_args_group.add_argument('--dry-run',
-                                              action = 'store_true',
-                                              help   = 'If given, the script will not push the built images to gcr.')
 
         # flag to turn off git protection (default modue is refusing to build when there are untracked files and/or uncommited changes)
         parser.add_argument('--disable-git-protect',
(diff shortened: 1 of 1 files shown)


scripts/docker/build_docker.py
```diff
@@ -80,6 +80,10 @@ def __init__(self, args_list):
                                               action = 'store_true',
                                               help   = 'also update \"latest\" tag in remote docker repo(s)')
 
+        docker_remote_args_group.add_argument('--dry-run',
+                                              action = 'store_true',
+                                              help   = 'If given, the script will not push the built images to gcr.')
```
mwalker174, line 85: Would it be easier to make `--gcr-project` optional? I believe the only difference will be the removal of `--squash` from the `docker` command.

VJalili: Sure, I can make `--gcr-project` optional instead of `--dry-run`. Would you mind please elaborate on `--squash`? should that be ignored if `--gcr-project` is not given?

mwalker174: You can read about it [here](https://docs.docker.com/engine/reference/commandline/build/#squash-an-images-layers---squash-experimental) - I'm not sure we actually need to be doing this in the first place, but it will not affect the functionality of the image at all.

VJalili: Thank you for the link about `--squash`, actually I was wondering if you were suggesting using this flag to achieve functionality similar to `--dry-run`. 

> 'm not sure we actually need to be doing this in the first place,

mwalker174: `--squash`

mwalker174: We discussed offline, but we do want to keep `--squash` because it can reduce final image size in some cases.

VJalili: I guess `--gcr-project` is optional by default, I just updated its help message to reflect images will not be pushed if a value for the argument is not provided. I agree leveraging this argument is simpler/clearer than a new argument.

mwalker174 (approved):
Looks fine although I guess this didn't change any code functionality.

VJalili:
Thanks, @mwalker174 Yes, I think in the end we landed on extending a description.

PR #184, opened 2021-06-11
Enable easier skipping of Melt for Terra

epiercehoffman:
### Updates
* Check if VCFs arrays have length > 0 as well as checking if they are defined, to enable skipping of Melt steps even if a `this.melt_vcfs` argument is provided, if the column in the Terra data table is empty. This is necessary because providing an argument causes the `Array[File]?` to register as defined, but if the column is empty the array will be empty (`[]`), causing errors.
* Move around some Terra JSON templates in preparation for validation script
* Add a note about `gvcf_gcs_project_for_requester_pays` argument to Module00c to Terra cohort mode dashboard

### Testing
* Tested Module00b, Module00c and Module01 in Terra to confirm that tasks will not run if an empty column in the data table is referenced, and that the added check does not interfere with the VCF arrays that are non-empty
* Tested Module03 in Terra to confirm that a `File?` is handled properly if referenced with `this.attribute` but the attribute is empty in the data table - the above issue only applies to `Array[X]?` types
* Tested input build from templates

### Resulting behavior in Terra
Melt runs by default in Module00a, but can be disabled by deleting one argument. (This behavior is up for debate.) Thereafter, no changes need to be made regardless of whether Melt was run.


cwhelan (approved):
This looks good to me!

> Melt runs by default in Module00a, but can be disabled by deleting one argument. (This behavior is up for debate.) 

I think I fall on the side that says that we default the Terra workspaces to not running MELT and then having the user configure the docker. This will prevent confusing errors for people who don't have access to the private dockerfile and forget to remove it. Just my $0.02 for the debate, though.

PR #185, opened 2021-06-15
Add validation script for Terra cohort mode input JSONs

epiercehoffman:
### Updates
* Add validation script `/scripts/test/terra_validation.py` for cursory validation of Terra cohort mode input JSONs (pipeline workflows & metrics workflows). See comment at top of script for details and usage.
* Edit `/scripts/test/validate.sh` to add optional `-t` flag that, when provided, will run Terra validation script after test inputs validation.
* Move some JSON templates around to accommodate the validation script better.

### Testing
Current branch passes validation. Tested with missing input, extra input, and malformed JSON file (the error message for the last one is a little messy since it's from `json.load`). Tested `validate.sh` with and without `-t` flag.


scripts/test/terra_validation.py
```diff
@@ -0,0 +1,135 @@
...
+  num_input_jsons = len(jsons)
+  if num_input_jsons < NUM_TERRA_INPUT_JSONS:
+    logging.warning(f"Expected {NUM_TERRA_INPUT_JSONS} Terra input JSONs but found {num_input_jsons}.")
+  jsons.sort()
+
+  metrics_jsons = [METRICS_INPUTS_PATH + x for x in listdir(inputs_path + METRICS_INPUTS_PATH) if x.endswith(".json")]
+  num_metrics_jsons = len(metrics_jsons)
+  if num_metrics_jsons < NUM_TERRA_METRICS_JSONS:
+    logging.warning(f"Expected {NUM_TERRA_METRICS_JSONS} Terra metrics input JSONs but found {num_metrics_jsons}.")
+  metrics_jsons.sort()
```
mwalker174, line 55: This looks a lot like the block above. Can you create a common function to replace both blocks?

scripts/test/terra_validation.py
```diff
@@ -0,0 +1,135 @@
...
+    path_to_wdl = wdl_path + basename(json_file)[:-5] + ".wdl"
+    if isfile(path_to_wdl):
+      yield path_to_wdl, terra_inputs_path + json_file
+    else:
+      logging.warning(f"Can't find WDL corresponding to {basename(json_file)} at {path_to_wdl}.")
+
+
+def validate_terra_json(wdl, terra_json, womtool_jar):
+  womtool_input_generation_command = "java -jar " + womtool_jar + " inputs " + wdl + " > " + TMP_JSON
+  subprocess.run(womtool_input_generation_command, shell=True)
```
mwalker174, line 74: You can use `capture_output=True` to load stdout directly into memory as a string rather than writing to a temp file.

mwalker174 (commented):
This looks good. I have a few suggestions to improve the script. I was also wondering if you could generalize it a bit by taking in the directory path(s) to the jsons as an argument. It seems like this could then replace the test bash script for the cromwell inputs.

epiercehoffman:
> This looks good. I have a few suggestions to improve the script. I was also wondering if you could generalize it a bit by taking in the directory path(s) to the jsons as an argument. It seems like this could then replace the test bash script for the cromwell inputs.

Thanks for the review! I think I've resolved all your comments, as well as some linting errors (all that is left is line length but I made sure the lines < 120 characters). 

I don't think this should be used as-is for the cromwell inputs because it performs a much more limited validation than `womtool validate`. As such, I've left the base directory as the input because that seemed easier than requiring the user to provide the full path to the Terra JSONs.

Additionally, in the interest of using this for CI in the future, I did add Exceptions for the wrong number of JSONs, as well as one at the end if the number of PASS JSONs < the number of total JSONs. I wanted to ask if this behavior is fine, or if you'd prefer to error out as soon as one JSON fails? I did it this way so that the user could see all errors at once, if there are multiple, but I do recognize it could be more difficult to find the detailed error messages this way as they will be in the middle of all the PASS lines.

mwalker174 (approved):
Thank you, this looks great! I agree it is a good idea to fail-slow rather than fail-fast for this.

PR #186, opened 2021-06-15
Support original sample ids in snp vcfs

mwalker174:
Adds optional input `vcf_samples` to Module00c so that the user doesn't need to modify snp vcf headers if they've changed sample ids.


epiercehoffman (approved):
Looks great! Should be simple to use in cromwell or Terra. This will be really helpful, much better than manually reheadering huge VCF files

PR #190, opened 2021-06-21
A Github action to test building docker images

VJalili:
This PR implements a Github action that triggers when a PR is issued against the `master` branch. The action invokes the `build_docker.py` script to build `all` the docker images if the PR involves any changes to the `src/*` directory. If the `build_docker.py` script exists successfully, the action passes, otherwise, it fails.

### DIFF
--- a/.github/workflows/sv_pipeline_docker.yml
+++ b/.github/workflows/sv_pipeline_docker.yml
@@ -6,11 +6,13 @@ on:
       - master
     paths:
       - 'src/**'
+      - 'dockerfiles/**'
   pull_request:
     branches:
       - master
     paths:
       - 'src/**'
+      - 'dockerfiles/**'
 
 jobs:
   test_build:
--- a/.github/workflows/sv_pipeline_docker.yml
+++ b/.github/workflows/sv_pipeline_docker.yml
@@ -4,13 +4,11 @@ on:
   push:
     branches:
       - master
-      - docker_build
     paths:
       - 'src/**'
   pull_request:
     branches:
       - master
-      - docker_build
     paths:
       - 'src/**'
 
--- a/.github/workflows/sv_pipeline_docker.yml
+++ b/.github/workflows/sv_pipeline_docker.yml
@@ -1,0 +1,43 @@
+name: Build Docker Images
+
+on: 
+  push:
+    branches:
+      - master
+      - docker_build
+    paths:
+      - 'src/**'
+  pull_request:
+    branches:
+      - master
+      - docker_build
+    paths:
+      - 'src/**'
+
+jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/sv_pipeline_docker.yml
```diff
@@ -0,0 +1,42 @@
...
+          
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install termcolor
+
+      - name: Run build_docker.py
+        run: |
+          cd ./scripts/docker/
+          python build_docker.py --targets all --image-tag ${PR_NUM}-${COMMIT_SHA::8}
```
mwalker174, line 42: Fine for now - `all` will not build MELT, which is the desired behavior in this case because the binaries need to be downloaded manually. It would be ideal to trigger a warning on the PR when a MELT dependency changes, if possible in a future PR.

.github/workflows/sv_pipeline_docker.yml
```diff
@@ -0,0 +1,42 @@
+name: Build Docker Images
+
+on: 
+  push:
+    branches:
+      - master
+    paths:
+      - 'src/**'
```
mwalker174, line 8: Should we also include `dockerfiles/**`?

.github/workflows/sv_pipeline_docker.yml
```diff
@@ -0,0 +1,42 @@
...
+  pull_request:
+    branches:
+      - master
+    paths:
+      - 'src/**'
+
+jobs:
+  test_build:
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
VJalili: Sure!

.github/workflows/sv_pipeline_docker.yml
VJalili: Agree, updated.

.github/workflows/sv_pipeline_docker.yml
VJalili: Agree we should also build `MELT`, I left it for future PRs as it requires hosted binaries (at least accessible to GitHub actions). I am not sure [if Github actions currently support warning, in addition to fail/pass](https://github.com/isaacs/github/issues/1465), so maybe we would leave checks/builds on `MELT` to continue to be manual for a while until we automate their builds?!

mwalker174: Hm well doesn't look like there is much interest in warnings. But yes, let's leave MELT to manual builds for now.

mwalker174 (approved):
Looks good to me! Excited to see how it runs!

PR #192, opened 2021-06-25
Add regenotyped depth vcf indexes to Module04b output

mwalker174:
These indexes are expected by Module0506. Passes womtool validation and tested on Cromwell.


epiercehoffman (approved):
Looks fine to me!

PR #194, opened 2021-06-30
Add ExpansionHunter Denovo WDL 

VJalili:
This PR implements a WDL to use ExpansionHunter Denovo's (EHdn) [case-control analysis](https://github.com/Illumina/ExpansionHunterDenovo/blob/master/documentation/03_Case_control_quickstart.md#quickstart-case-control-analysis). 

- [x] Implement WDL;
- [x] Update `ehdn` docker image. The image is based on `python:3.7-slim`. The version `3.7` of python is chosen because of EHdn's script's pinned requirements (`numpy==1.17.1` and `scipy==1.3.1`). `slim` is chosen instead of `alpine` because of dependency issues at the time of dev/debugging. Hopefully, the image size is no big deal, if so, we can revert to `alpine` and try to resolve the dependency issues. 
- [ ] ~~Implement Carrot unit test.~~ (postponed to a separate PR: https://github.com/broadinstitute/gatk-sv/pull/209)

### DIFF
--- a/wdl/ExpansionHunterDenovo.wdl
+++ b/wdl/ExpansionHunterDenovo.wdl
@@ -1,9 +1,10 @@
-## ExpansionHunter denovo (EHdn)
+##  ExpansionHunter denovo (EHdn)
 ##
-## This WDL implements workflows for
-## EHdn's two execution modes:
-## (i) case-control analysis;
-## (ii) outliner analysis.
+##  This WDL implements workflows for EHdn's two execution modes:
+##  - case-control analysis (read EHdn docs on this:
+##    https://github.com/Illumina/ExpansionHunterDenovo/blob/master/documentation/03_Case_control_quickstart.md);
+##  - outliner analysis (read EHdn docs on this:
+##    https://github.com/Illumina/ExpansionHunterDenovo/blob/master/documentation/04_Outlier_quickstart.md).
 
 version 1.0
 
@@ -36,10 +37,12 @@ workflow EHdnSTRAnalysis {
   }
 
   parameter_meta {
-    analysis_type: "possible values: {casecontrol, outlier, both}"
-    str_comparison_type: "possible values: {locus, motif, both}"
-    case_reads_filenames: ""
-    control_reads_filenames: ""
+    analysis_type: "Sets the analysis type; accepted values are: `casecontrol`, `outlier`, and `both`."
+    str_comparison_type: "Set the STR comparison type; accepted values are: `locus`, `motif`, and `both`."
+    case_reads_filenames: "A list of BAM files to be used as `case` samples."
+    case_indexes_filenames: "A list of index files (.bam.bai) for the `case` samples. Files should be in the same order as the case samples"
+    control_reads_filenames: "A list of BAM files to be used as `control` samples."
+    control_indexes_filenames: "A list of index files (.bam.bai) for `control` samples. Files should be in the same order as the control samples (i.e., the `control_reads_filenames` list)."
     manifest_filename: ""
     reference_filename: ""
     min_anchor_mapq: ""
--- a/wdl/ExpansionHunterDenovo.wdl
+++ b/wdl/ExpansionHunterDenovo.wdl
@@ -110,32 +110,24 @@ workflow EHdnSTRAnalysis {
   Array[File] cases_str_profile_json = select_all(case_str_profile_json)
   Array[File] controls_str_profile_json = select_all(control_str_profile_json)
 
-  call Merge {
+  call STRAnalyze {
     input:
       cases = cases_str_profile_json,
       controls = controls_str_profile_json,
       reference_fasta = reference_fasta,
       reference_fasta_index = reference_fasta_index,
-      ehdn_docker = ehdn_docker,
-      runtime_attr_override = runtime_attr_merge,
-      postfixes = postfixes
-  }
-
-  call STRAnalyze {
-    input:
+      postfixes = postfixes,
       analysis_type = analysis_type,
       str_comparison_type = str_comparison_type,
-      manifest = Merge.manifest,
-      multisample_profile = Merge.multisample_profile,
       ehdn_docker = ehdn_docker,
       runtime_attr_override = runtime_attr_analysis
   }
 
   output {
+    File multisample_profile = STRAnalyze.multisample_profile
     Array[File] cases_locus = ComputeSTRProfile.locus
     Array[File] cases_motif = ComputeSTRProfile.motif
     Array[File] cases_str_profile = ComputeSTRProfile.str_profile
-    File multisample_profile = Merge.multisample_profile
     Array[File] analysis_results = STRAnalyze.results
   }
 }
@@ -203,25 +195,29 @@ task ComputeSTRProfile {
   }
 }
 
-task Merge {
+task STRAnalyze {
   input {
+    String analysis_type
+    String str_comparison_type
     Array[File] cases
     Array[File] controls
     File reference_fasta
     File? reference_fasta_index
+    FilenamePostfixes postfixes
     String ehdn_docker
     RuntimeAttr? runtime_attr_override
-    FilenamePostfixes postfixes
   }
 
   output {
     File manifest = "manifest.tsv"
     File multisample_profile = "${output_prefix}${postfixes.merged_profile}"
+    Array[File] results = glob("result_*.tsv")
   }
 
   Int cases_length = length(cases)
   Int controls_length = length(controls)
   String output_prefix = "merged"
+  String multisample_profile = output_prefix + postfixes.merged_profile
 
   # Defining this varialbe is requires since it
   # will trigger localization of the file. The
@@ -245,6 +241,7 @@ task Merge {
   #
   # - Call EHdn's `merge` method using the input and
   #   the generaged manifest file.
+
   command <<<
     set -euxo pipefail
 
@@ -256,8 +253,6 @@ task Merge {
     }
     manifest_filename="manifest.tsv"
 
-    echo ~{cases_length}
-    echo ~{controls_length}
     if [ ~{cases_length} -ne 0 ]; then
       cases_arr=(~{sep=" " cases})
       for i in "${cases_arr[@]}"; do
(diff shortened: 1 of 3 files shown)


wdl/ExpansionHunterDenovo.wdl
```diff
@@ -0,0 +1,347 @@
...
+
+workflow EHdnSTRAnalysis {
+
+  input {
+    String analysis_type
+    String str_comparison_type
+    Array[File] case_reads_filenames
+    Array[File] case_indexes_filenames
+    Array[File] control_reads_filenames
+    Array[File] control_indexes_filenames
```
mwalker174, line 28: Does ehdn support bams or crams or both? I would use that in the variable name e.g. `case_crams` or `case_crams_or_bams`.

mwalker174: Also this will be tricky to map to the Terra data model. I would propose having a single array for the reads and a single array for the indexes (without case/control designation), and then have a third array `Array[String] sample_status` with possible values `case` or `control`. Then you can use a scatter to select each like this:
```
scatter (i in range(length(sample_crams_or_bams)) {
    if (sample_status[i] == "case") {
        # Creates Array[File?] case_crams_or_bams_ and case_indexes_
        File case_crams_or_bams_ = sample_crams_or_bams[i]
        File case_indexes_ = sample_indexes[i]
    }
    if (sample_status[i] == "control") {
        File control_crams_or_bams_ = sample_crams_or_bams[i]
        File control_indexes_ = sample_indexes[i]
    }
}
```
and then when calling the Case task you just use `select_all` like this:
```
reads_filename = select_all(case_crams_or_bams_),
index_filename = select_all(case_indexes_)
```
and similarly for the Controls call.

wdl/ExpansionHunterDenovo.wdl
```diff
@@ -0,0 +1,347 @@
...
+    cpu_cores: 1,
+    mem_gb: 4,
+    disk_gb: 10,
+    boot_disk_gb: 10,
+    preemptible_tries: 0,
+    max_retries: 1
+  }
+  RuntimeAttr runtime_attr_str_profile = select_first([
+    runtime_attr_str_profile,
+    runtime_attr_str_profile_default])
```
mwalker174, line 87: Did you move these outside the task so you can have separate defaults for different calls of the same task? i.e. `CasesProfiles` and `ControlsProfiles`?

wdl/ExpansionHunterDenovo.wdl
```diff
@@ -0,0 +1,347 @@
...
+  Int profile_len
+}
+
+workflow EHdnSTRAnalysis {
+
+  input {
+    String analysis_type
+    String str_comparison_type
+    Array[File] case_reads_filenames
+    Array[File] case_indexes_filenames
```
mwalker174, line 26: I would make the index column optional and infer index URIs if not provided - see `Module00a`'s `bam_or_cram_index` input as an example.

wdl/ExpansionHunterDenovo.wdl
```diff
@@ -0,0 +1,347 @@
...
+    Array[File] control_reads_filenames
+    Array[File] control_indexes_filenames
+    File reference_filename
```
mwalker174, line 29: I would again avoid `filename` - usually we use `reference_fasta`. Also, does ehdn require this to be indexed? You may also need to include that (file extension `.fai`).

wdl/ExpansionHunterDenovo.wdl
```diff
@@ -0,0 +1,347 @@
...
+  # - Create a "manifest" file based on the given inputs.
+  #   The manifest file is composed of three columns:
+  #   (1) Sample name (infered from the filename);
+  #   (2) case/control (infered from files in the inputs
+  #       `cases` and `controls`);
+  #   (3) file path.
+  #
+  # - Call EHdn's `merge` method using the input and
+  #   the generaged manifest file.
+  command <<<
```
mwalker174, line 251: Always start this section with `set -euo pipefail` (or `set -euxo pipefail` if it won't generate excessive logging).

wdl/ExpansionHunterDenovo.wdl
```diff
@@ -0,0 +1,347 @@
...
+    File multisample_profile
+    String ehdn_docker
+    RuntimeAttr runtime_attr
+  }
+
+  output {
+    Array[File] results = glob("result_*.tsv")
+  }
+
+  command <<<
```
mwalker174, line 305: `set -euo pipefail`

wdl/ExpansionHunterDenovo.wdl
```diff
@@ -0,0 +1,347 @@
+## ExpansionHunter denovo (EHdn)
+##
+## This WDL implements workflows for
+## EHdn's two execution modes:
+## (i) case-control analysis;
+## (ii) outliner analysis.
```
mwalker174, line 6: `outlier`

wdl/ExpansionHunterDenovo.wdl
VJalili: ExpansionHunter denovo requires a [manifest file](https://github.com/Illumina/ExpansionHunterDenovo/blob/master/examples/case-control/manifest.tsv) that serves a similar purpose to the `sample_status` array you suggested. I found compiling that file would be a bit cumbersome for user and error prune so instead, I switched to using two lists and auto-generating the manifest accordingly. However, if tricky on terra, I'm happy to revert it. Which one do you think will be more intuitive and simpler on Terra, manifest file or `sample_status` array?

wdl/ExpansionHunterDenovo.wdl
VJalili: Good point, I added `set -euxo pipefail` to all the commands.

wdl/ExpansionHunterDenovo.wdl
VJalili: I was trying to make the tasks less crowded, but I guess it made them less portable. I'll implement these defaults in tasks instead.

wdl/ExpansionHunterDenovo.wdl
VJalili: Thanks for the suggestion, so even if a file is defined in the body of a workflow, Cromwell will localize that file.

wdl/ExpansionHunterDenovo.wdl
mwalker174: Terra has users upload sample-level metadata as a table and define sets of samples (batches) using sample IDs. The user can then launch batched workflows, like this one, with some inputs (eg sample id, cram path, or case/control status) generated from columns of the table as arrays. 

They would have to create and upload the manifest file separately and then manually enter the bucket URL into the inputs. You're right - that's extra work and prone to errors. Therefore we should use arrays for all the inputs and compose the manifest on the fly, even if it's a small cost. 

Note also that Terra does not support certain WDL functions so you may need to write a task to create the manifest (see [here](https://support.terra.bio/hc/en-us/community/posts/360071465631-write-lines-write-map-write-tsv-write-json-fail-when-run-in-a-workflow-rather-than-in-a-task)).

wdl/ExpansionHunterDenovo.wdl
```diff
@@ -0,0 +1,332 @@
...
+
+  >>>
+
+  RuntimeAttr runtime_attr_str_profile_default = object {
+    cpu_cores: 1,
+    mem_gb: 4,
+    disk_gb: 10,
+    boot_disk_gb: 10,
+    preemptible_tries: 0,
+    max_retries: 1
```
mwalker174, line 181: Before merging, I would at least try to make disk_gb scale with input size, ie: `disk_gb: 10 + size([reference_fasta, bam_or_cram]),`

We can put off memory optimizations until you have done tests with real BAMs.

Also how long does this task typically take? Rule of thumb is if it takes less than 2 hours then set `preemptible_tries: 3`, otherwise try to parallelize the task with a `scatter`. If the latter is not possible we use `preemptible_tries: 0` but the cost is 5x higher. 

Same applies to `STRanalyze` below.

mwalker174 (commented):
This is looking very good. One set of comments regarding runtime attributes. Before approving, I'd want to test the WDL on a set of real samples. What testing have you done so far?

wdl/ExpansionHunterDenovo.wdl
VJalili: Good point, thanks! I updated accordingly.

VJalili:
@mwalker174 Thank you for the comments. I run this workflow on six samples as in the following json, and the workflow executed successfully. The run ID is: `e17c4f36-9941-4a20-b828-60b89fe11bad`

```json
"EHdnSTRAnalysis.sample_bams_or_crams": [
    "gs://fc-56ac46ea-efc4-4683-b6d5-6d95bed41c5e/CCDG_14151/Project_CCDG_14151_B01_GRM_WGS.cram.2020-02-12/Sample_HG00405/analysis/HG00405.final.cram",
    "gs://fc-56ac46ea-efc4-4683-b6d5-6d95bed41c5e/CCDG_14151/Project_CCDG_14151_B01_GRM_WGS.cram.2020-02-12/Sample_HG00408/analysis/HG00408.final.cram",
    "gs://fc-56ac46ea-efc4-4683-b6d5-6d95bed41c5e/CCDG_14151/Project_CCDG_14151_B01_GRM_WGS.cram.2020-02-12/Sample_HG00418/analysis/HG00418.final.cram",
    "gs://fc-56ac46ea-efc4-4683-b6d5-6d95bed41c5e/CCDG_14151/Project_CCDG_14151_B01_GRM_WGS.cram.2020-02-12/Sample_HG00420/analysis/HG00420.final.cram",
    "gs://fc-56ac46ea-efc4-4683-b6d5-6d95bed41c5e/CCDG_14151/Project_CCDG_14151_B01_GRM_WGS.cram.2020-02-12/Sample_HG00423/analysis/HG00423.final.cram",
    "gs://fc-56ac46ea-efc4-4683-b6d5-6d95bed41c5e/CCDG_14151/Project_CCDG_14151_B01_GRM_WGS.cram.2020-02-12/Sample_HG00427/analysis/HG00427.final.cram",
    "gs://fc-56ac46ea-efc4-4683-b6d5-6d95bed41c5e/CCDG_14151/Project_CCDG_14151_B01_GRM_WGS.cram.2020-02-12/Sample_HG00427/analysis/HG00427.final.cram"
],
```

mwalker174 (approved):
Good to merge! Looking forward to seeing testing for this!

VJalili:
Thanks, @mwalker174 I am also excited to have tests for wdls.

PR #197, opened 2021-07-09
Fix svtest vcf pass evaluation

mwalker174:
Fixes a glaring issue where PASS variants in the baseline vcf were compared against all variants in the test vcf regardless of filter status.


cwhelan (approved):
👍

PR #198, opened 2021-07-10
Separate build logic for PR & push in diff. steps.

VJalili:
Concerning building docker images, we'd like to take different steps for PRs and pushes (e.g., PR merge commits). For instance, while we might only be interested in testing if images are successfully built for PRs, we'd like to build and push the images to GCR and update `input_values/dockers.json` accordingly. Therefore, defining separate tasks for each can simplify tailoring specific steps for PRs and pushes. This PR takes the first step in that direction and solves the issue of setting incorrect tags for images on merge commits. 

### Images tag
We're currently using the `{PR number}-{SHA8}` pattern to tag docker images. While this pattern works for images built within the context of a PR, it does not work for pushes (e.g., merge commits) since they happen outside the PR context. This PR implements the `{Date}-{SHA8}` pattern (e.g., `20210709-12345678`) for images built for merge commits.

### DIFF
--- a/.github/workflows/sv_pipeline_docker.yml
+++ b/.github/workflows/sv_pipeline_docker.yml
@@ -1,6 +1,6 @@
 name: Build Docker Images
 
-on: 
+on:
   push:
     branches:
       - master
@@ -8,6 +8,7 @@ on:
       - 'src/**'
       - 'dockerfiles/**'
       - 'scripts/**'
+      - '.github/workflows/sv_pipeline_docker.yml'
   pull_request:
     branches:
       - master
@@ -15,6 +16,7 @@ on:
       - 'src/**'
       - 'dockerfiles/**'
       - 'scripts/**'
+      - '.github/workflows/sv_pipeline_docker.yml'
 
 jobs:
   test_build:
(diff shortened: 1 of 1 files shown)


.github/workflows/sv_pipeline_docker.yml
```diff
@@ -1,44 +1,71 @@
...
-on: 
+on:
   push:
     branches:
       - master
     paths:
       - 'src/**'
       - 'dockerfiles/**'
+      - 'scripts/**'
+      - '.github/workflows/sv_pipeline_docker.yml'
```
mwalker174, line 11: Why were these added? (same below?)

VJalili: A change on this file can impact the image build and push to GCR. For instance, if we would like to change how we tag the images, or a PR may change how the `build_docker` is called and break building images. Without adding this file to the `paths` list, any such PRs (or pushes) will not trigger building and pushing the images.

mwalker174: I see - can you change `scripts/**` to `scripts/docker/build_docker.py` and `scripts/docker/resources.Dockerfile`? Those are the only files that should affect the build.

VJalili: Good point, thanks. Changed.

mwalker174 (approved):
Thanks @VJalili this is great! Feel free to merge when you feel ready.

VJalili:
Thanks, @mwalker174!

PR #210, opened 2021-08-02
WDL and docker file for ExpansionHunter

VJalili:
The WDL runs ExpansionHunter, the docker file contains ExpansionHunter and TRTools. The WDL will be extended to run the tools in TRTools.

### DIFF
--- a/dockerfiles/str/Dockerfile
+++ b/dockerfiles/str/Dockerfile
@@ -6,8 +6,8 @@
 
 FROM ubuntu:20.04
 
-RUN apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get install -qqy \
-    python3 \
+RUN apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get install --no-install-recommends -qqy \
+    python3-dev \
     python3-pip \
     python \
     python-dev \
--- a/dockerfiles/str/Dockerfile
+++ b/dockerfiles/str/Dockerfile
@@ -19,11 +19,11 @@ RUN apt-get update && DEBIAN_FRONTEND="noninteractive" apt-get install -qqy \
 RUN pip3 install pybedtools==0.8.2 pyvcf==0.6.8 scipy==1.7.1 numpy==1.21.1
 
 # Install samtools (needed to index reference fasta files)
-RUN wget -O samtools-1.9.tar.bz2 https://github.com/samtools/samtools/releases/download/1.9/samtools-1.9.tar.bz2
-RUN tar -xjf samtools-1.9.tar.bz2
-WORKDIR samtools-1.9
-RUN ./configure --without-curses && make && make install
-WORKDIR ..
+RUN wget -O samtools-1.9.tar.bz2 https://github.com/samtools/samtools/releases/download/1.9/samtools-1.9.tar.bz2 \
+    && tar -xjf samtools-1.9.tar.bz2 \
+    && cd samtools-1.9 \
+    && ./configure --without-curses && make && make install \
+    && cd ..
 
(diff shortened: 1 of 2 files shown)


dockerfiles/str/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+    git \
+    libbz2-dev \
+    liblzma-dev \
+    make \
+    pkg-config \
+    wget \
+    unzip \
+    zlib1g-dev
+
+RUN pip3 install pybedtools pyvcf scipy numpy
```
mwalker174, line 19: Ideally I would peg all dependencies to specific versions, if possible. This can help us rule out changes to dependencies as a reason for differences in output with future builds.

dockerfiles/str/Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+
+# Install samtools (needed to index reference fasta files)
+RUN wget -O samtools-1.9.tar.bz2 https://github.com/samtools/samtools/releases/download/1.9/samtools-1.9.tar.bz2
+RUN tar -xjf samtools-1.9.tar.bz2
+WORKDIR samtools-1.9
+RUN ./configure --without-curses && make && make install
+WORKDIR ..
```
mwalker174, line 26: Is there an advantage to using `WORKDIR` rather than `cd` inside of `RUN`? I remember reading that combining multiple steps into single `RUN` statements can help reduce image size, although I'm unsure if that's true if you flatten the image at the end.

dockerfiles/str/Dockerfile
VJalili: That makes sense; I pinned `pybedtools`, `pyvcf`, `scipy`, and `numpy` to their current latest releases.

dockerfiles/str/Dockerfile
VJalili: Good point. I guess using `WORKDIR` over `cd` is suggested in [docker docs](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#workdir) for clarity. Also, I am not sure about image size either. I gathered multiple `RUN`s in a single `RUN`, it had very little impact on the _compressed_ size at least: [`683.6 MB` for multiple `RUN`s vs `683.71 MB` for single `RUN`](https://hub.docker.com/repository/registry-1.docker.io/vjalili/str/tags?page=1&ordering=last_updated).

mwalker174 (commented):
Overall looks good to me. I have some comments/suggestions below. As with EHDN, I'd be sure to test this with real data to make sure it runs smoothly as well (assuming you've done another check for correctness with test data). 

I think the image size is acceptable but I did come across this article with some interesting strategies for trying to shrink the python down: https://towardsdatascience.com/how-to-shrink-numpy-scipy-pandas-and-matplotlib-for-your-data-product-4ec8d7e86ee4

pdiakumis:
Hi @mwalker174 @VJalili , sorry for intruding like this, but have you considered using/copying the Quay docker images generated automatically via bioconda? e.g. 

* https://quay.io/repository/biocontainers/expansionhunter?tab=tags (via https://anaconda.org/bioconda/expansionhunter)
* https://quay.io/repository/biocontainers/trtools?tab=tags (via https://anaconda.org/bioconda/trtools)

Or are you trying to combine multiple tools in single images? Just curious.

VJalili:
Hi @pdiakumis that is a good suggestion, thank you, using bioconda images can save us from re-building them. I guess in long run we're interested in building images for particular tasks that include a set of related tools (e.g., an image for STR calling that includes ExH, EHdn, TRTools, GangSTR, and etc.) and ideally hosting the images on GCR to avoid possible egress charges.

PR #211, opened 2021-08-03
Run metrics workflows as part of main module WDLs

epiercehoffman:
### Updates
Instead of separate Module0XTest WDLs that run Module0X and Module0XMetrics, import Module0XMetrics into Module0X and run at the end by default. The metrics workflows can provide useful debugging information, so it makes sense to run them regularly, and this will reduce clutter in `/wdl` and in the Terra workspace. Details:
* All modules run metrics at the end by default (but can toggle on/off via an input `run_module_metrics`) _except_ Module00cMetrics is off by default because it is expensive and long-running (the others are short and cheap). All metrics workflows are off for the single sample WDL by default, since there is a separate single-sample metrics workflow.
* Module00aBatchMetrics was deprecated and replaced by Module00aMetrics, which runs on a per-sample basis, for compatibility with Terra which uses Module00a instead of Module00aBatch. Module00aBatch runs the combine-metrics step at the end. 
* The metrics workflows run as part of the individual modules when running the batch WDL, instead of all at the end. The combine-metrics steps run at the end of the batch WDL.
* Deleted all the Module0XTest WDLs (but not SingleSampleTest) and Module00aBatchMetrics (but have not done so yet in case this is not desired), Module0XTest JSON templates, and Terra metrics JSON templates

### Testing
* Batch WDL: Succeeded in parts. Run 1 succeeded through Module0506 but failed due to a missing input in Module0506Metrics. Runs 2-3 did not cache and failed unexpectedly in Module04 and I was not able to track down the source of the error. For Run 4 I created a modified WDL that ran Module0506 and later steps, and I provided inputs from Run 1; it succeeded.
* Single-sample WDL: Succeeded. Just running this one to make sure I didn't mess anything up - all the module metrics workflows are off because the single-sample pipeline has its own metrics.
* Womtool JSON validation: reduced to 38 test JSONs and 16 Terra JSONs


mwalker174 (approved):
Looks good! Just one thing I was curious about below.

PR #221, opened 2021-08-31
Rewrite cleanvcf part 3 in python

epiercehoffman:
Needs downstream change in part 4 because files are named differently

### DIFF
--- a/src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part3.py
+++ b/src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part3.py
@@ -1,0 +1,58 @@
+#!/bin/python
+
+import argparse
+from collections import defaultdict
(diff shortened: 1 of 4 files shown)


epiercehoffman:
> Do you think this can be done with a single iteration over the input file? It should be sorted already. May not be necessary if the run time / memory footprint are already small, though?
> 

I could envision counting the variants and sharding in a single pass IF the variant count dictionary didn't need to be sorted before sharding. I believe the counts are sorted in descending order so that the shards will be more even - ie. if partitioning variants with counts 1, 1, 6000 if the order is chromosomal you could end up with 1, 6000, 1 in which case there will be 3 shards; but if the order is 6000, 1, 1 (after sorting) there will be two shards. That's why I kept the sorting from the original algorithm, although I didn't do a real-life comparison with and without sorting.

PR #222, opened 2021-09-01
Eliminate Array[File?] workflow outputs not supported by Terra

epiercehoffman:
### Updates
Turns out null values in Array[File?] workflow outputs are not supported in Terra. So, had to eliminate them from FilterBatchSites and PlotSVCountsPerSample and update other inputs and JSONs accordingly.
* As a side effect, in the current implementation, the sv_counts calculated in PlotSVCountsPerSample can no longer be used in FilterBatchSamples because they may be in the wrong order of algorithms. So, cromwell will attempt to recalculate them but just call-cache. An alternative would be to create a version of PlotSVCountsPerSample that is specific to the FilterBatch context and a separate general version.

### Testing
* Retested FilterBatch and verified outputs are identical
* Tested upload of JSON input to PlotSVCountsPerSample and tested a workflow with [this.filtered_manta_vcf, this.filtered_depth_vcf] input style to make sure it ran as expected

### DIFF
--- a/wdl/FilterBatchSamples.wdl
+++ b/wdl/FilterBatchSamples.wdl
@@ -9,8 +9,12 @@ import "IdentifyOutlierSamples.wdl" as identify_outliers
 workflow FilterBatchSamples {
   input {
     String batch
-    Array[File?] vcfs  # in order of algorithms array: ["manta", "delly", "wham", "melt", "depth"]. To skip one, use null keyword - ex. ["manta.vcf.gz", null, "wham.vcf.gz", null, "depth.vcf.gz"]
-    Array[File?] sv_counts  # one SV counts file from PlotSVCountsPerSample per VCF in the same order
+    File? sites_filtered_manta_vcf
+    File? sites_filtered_delly_vcf
+    File? sites_filtered_wham_vcf
+    File? sites_filtered_melt_vcf
+    File? sites_filtered_depth_vcf
+    Array[File?]? sv_counts_array  # one SV counts file from PlotSVCountsPerSample per VCF in the order ["manta", "delly", "wham", "melt", "depth"]. Enter null for missing SV counts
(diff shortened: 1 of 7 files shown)


wdl/FilterBatchSamples.wdl
```diff
@@ -9,8 +9,12 @@ import "IdentifyOutlierSamples.wdl" as identify_outliers
 workflow FilterBatchSamples {
   input {
     String batch
-    Array[File?] vcfs  # in order of algorithms array: ["manta", "delly", "wham", "melt", "depth"]. To skip one, use null keyword - ex. ["manta.vcf.gz", null, "wham.vcf.gz", null, "depth.vcf.gz"]
-    Array[File?] sv_counts  # one SV counts file from PlotSVCountsPerSample per VCF in the same order
+    File? sites_filtered_manta_vcf
+    File? sites_filtered_delly_vcf
+    File? sites_filtered_wham_vcf
+    File? sites_filtered_melt_vcf
+    File? sites_filtered_depth_vcf
```
mwalker174, line 16: Do these need to be "sites filtered"? It seems like this would work on arbitrary vcfs

wdl/FilterBatchSamples.wdl
```diff
@@ -9,8 +9,12 @@ import "IdentifyOutlierSamples.wdl" as identify_outliers
...
+    Array[File?]? sv_counts_array  # one SV counts file from PlotSVCountsPerSample per VCF in the order ["manta", "delly", "wham", "melt", "depth"]. Enter null for missing SV counts
```
mwalker174, line 17: I think it would be clearer to just have:
```
    File? manta_counts
    File? delly_counts
    File? wham_counts
    File? melt_counts
    File? depth_counts
```

mwalker174 (approved):

PR #225, opened 2021-09-08
Multiple declarations of primary_contigs_fai

pdiakumis:
Keeping a single declaration so that `miniwdl check` passes at least.


mwalker174 (approved):
Thank you @pdiakumis!

PR #228, opened 2021-09-10
tws clean vcf part1

tedsharpe:


mwalker174:
@tedsharpe Can you also add the functionality from this script? 

https://github.com/broadinstitute/gatk-sv/blob/master/src/sv-pipeline/04_variant_resolution/scripts/add_bothsides_support_filter.py

It's very simple. This is called from the WDL immediately after the shell script, but I think we can avoid another pass on the vcf rolling it in here.

PR #229, opened 2021-09-13
Add docs for sample exclusion based on improperly paired read %

epiercehoffman:
Add docs to README recommending sample exclusion prior to GatherSampleEvidence for samples with a high percentage of improperly paired reads. Also add note to Terra cost estimate that the cost may vary depending on several factors and linking back to new sample exclusion docs


mwalker174 (approved):
Looks good! What do you think about also linking to this from a `Sample quality` subheading under `Quickstart`?

mwalker174 (approved):

PR #230, opened 2021-09-13
EvidenceQC.MedianCov memory & make batched workflow output filenames unique

epiercehoffman:
### Updates
* Add dynamic memory allocation based on number of samples to EvidenceQC.MedianCov (used same formula as GatherBatchEvidence.MedianCov)
* Make workflow output names that appear in Terra sample_sets table unique: outlier files from EvidenceQC and trained_PE/SR_metrics from GenotypeBatch (involved updating sv_pipeline_docker)
  * Requested by beta tester so that files from multiple batches can be downloaded from GCS without overwriting & without having to change the file name

### Testing
* Womtool validation of WDLs & JSONs
* Ran EvidenceQC on test_large data and verified that 60 GB of memory was allocated for 105 samples, that the memory allocated was reasonable compared to what was used (could probably do away with the 7.5 GB padding but doesn't appear to be a huge deal), and that the outlier files had `test_large.` prepended to the file names
* Ran `PE_genotype.sh` in the docker image on 1kgp data and verified that other than the filenames, all the files were bitwise identical to latest Terra run
* Ran GenotypeBatch on test_large data with rebuilt `sv_pipeline_docker` and verified that it completed successfully and the trained_PE_metrics and trained_SR_metrics files had `test_large.` prepended to the file names

### DIFF
--- a/input_values/dockers.json
+++ b/input_values/dockers.json
@@ -13,7 +13,7 @@
   "sv_base_docker" : "us.gcr.io/broad-dsde-methods/markw/sv-base:mw-gnomad-0506-pr-087d4df",
   "sv_base_mini_docker" : "us.gcr.io/broad-dsde-methods/markw/sv-base-mini:mw-gnomad-0506-pr-087d4df",
   "sv_pipeline_base_docker" : "us.gcr.io/broad-dsde-methods/markw/sv-pipeline-base:mw-gnomad-0506-pr-087d4df",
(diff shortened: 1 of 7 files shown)


mwalker174 (approved):
Looks okay to me. It may have been easier to just rename the script outputs with `mv` in the WDL command, but this fine as well.

epiercehoffman:
> Looks okay to me. It may have been easier to just rename the script outputs with `mv` in the WDL command, but this fine as well.

I wish I'd thought of that! I do think it's cleaner in the long run to just rename them in the script but definitely less work to rename in the WDL command block

PR #231, opened 2021-09-14
Clean vcf part 5 rewrite

cwhelan:
This PR rewrites clean vcf part 5 to make it scalable for processing very large VCFs.

I've structured this as a subworkflow that scatters the input VCF. Each shard of the input VCF is processed and updates to records are performed, while a VCF consisting solely of multiallelic variants and a list of variants with no called samples are built for each shard. The multiallelic VCFs are then merged and searched for redundancy (> 50% overlap). The final step simultaneously filters out redundant multiallelic records and sites with no called variants and concatenates the shards of updated records.

Verified that the results are identical to the previous run on the reference panel with two exceptions:

- There was a bug in the current script that was causing multiallelics with additional entries in the FILTER column (for example BOTHSIDES_SUPPORT) to be dropped; this has been fixed so these additional records are emitted.
- For sites where RD_CN was revised for male samples due to sex chromosome ploidy, the CN format field also reflects the update.

When tested on the previous gnomAD chr22 shard, this brings the runtime down from ~3 days to under 8 hours.


cwhelan:
By comparing my output to previous gnomAD shards I've found an additional bug that this rewrite seems to fix:

For the old script, the first record in the file was not revised into a CNV variant from DEL or DUP even if the record met the criteria. I believe this is due to the `NR>1` in the first awk statements in these commands from the old script:

```
zcat copystate.RD_CN.FORMAT.gz \
   |fgrep -wf <(zcat overlap.revise.bed.gz|awk -F"\t" '{if ($5=="DEL" && $3-$2>=1000) print $4}' ) \
   |awk 'NR>1{for(i=3;i<=NF;i++) if ($i!="." && $i>3) print  $1 }' \
   |sort \
   |uniq -c \
   |awk -v vf_1=$vf_1 '{if ($1>vf_1)print $2}' \
   |gzip \
    >multi.del.ids.txt.gz || true

zcat copystate.RD_CN.FORMAT.gz \
   |fgrep -wf <(zcat overlap.revise.bed.gz|awk -F"\t" '{if ($5=="DUP" && $3-$2>=1000) print $4}' ) \
   |awk 'NR>1{for(i=3;i<=NF;i++) if ($i!="." && $i>4) print  $1 }' \
   |sort \
   |uniq -c \
   |awk -v vf_1=$vf_1 '{if ($1>vf_1)print $2}' \
    >multi.dup.ids.txt || true
```

The fgreps on the second lines of each pipeline already removed the headers from the `copystate.RD_CN.FORMAT.gz` file, so skipping the first line was a bug there.

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -0,0 +1,185 @@
...
+    parser.add_argument('multi_geno_ids_txt')
+    parser.add_argument('outlier_samples_list', type=argparse.FileType('r'))
+    args = parser.parse_args()
+
+    # load the revised lines and index by ID
+    revised_lines_by_id = {}
+    with pysam.VariantFile(args.revise_vcf_lines) as revise_vcf:
+        header2 = revise_vcf.header
+        for record in revise_vcf:
+            revised_lines_by_id[record.id] = record
```
TedBrookings, line 28:
```suggestion
    with pysam.VariantFile(args.revise_vcf_lines) as revise_vcf:
        header2 = revise_vcf.header
    revised_lines_by_id = {record.id: record for record in revise_vcf}
```
When they're possible comprehensions are generally faster than appends

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -0,0 +1,185 @@
...
+            continue
+        fields = line.rstrip().split("\t")
+        if fields[4] == '1':
+            male_samples.add(fields[1])
+    print("identified {} male samples".format(len(male_samples)), file=sys.stderr)
+
+    if args.sexchr_revise.endswith(".gz"):
+        sexchr_revise = set([line.rstrip() for line in gzip.open(args.sexchr_revise, 'rt')])
+    else:
+        sexchr_revise = set([line.rstrip() for line in open(args.sexchr_revise, 'rt')])
```
TedBrookings, line 46:
```suggestion
        sexchr_revise = {line.rstrip() for line in gzip.open(args.sexchr_revise, 'rt')}
    else:
        sexchr_revise = {line.rstrip() for line in open(args.sexchr_revise, 'rt')}
```
Can avoid creating intermediate list

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -0,0 +1,185 @@
...
+    print("{} sites to revise on sex chromosomes".format(len(sexchr_revise)), file=sys.stderr)
+
+    if args.multi_geno_ids_txt.endswith(".gz"):
+        multi_geno_ids = set([line.rstrip() for line in gzip.open(args.multi_geno_ids_txt, 'rt')])
+    else:
+        multi_geno_ids = set([line.rstrip() for line in open(args.multi_geno_ids_txt, 'rt')])
```
TedBrookings, line 52:
```suggestion
        multi_geno_ids = {line.rstrip() for line in gzip.open(args.multi_geno_ids_txt, 'rt')}
    else:
        multi_geno_ids = {line.rstrip() for line in open(args.multi_geno_ids_txt, 'rt')}
```

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -0,0 +1,185 @@
...
+                        '##FILTER=<ID=MULTIALLELIC,Description="Multiallelic site">']
+
+    # # Add metadata lines for annotations
+    header1 = normal_vcf.header
+
+    for f in NEW_HEADER_LINES:
+        header1.add_line(f)
+        header2.add_line(f)
+
+    non_outlier_samples = set([s for s in header1.samples if s not in outlier_samples])
```
TedBrookings, line 70:
```suggestion
    non_outlier_samples = {s for s in header1.samples if s not in outlier_samples}
```

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -0,0 +1,185 @@
...
+            print("processed {} records".format(idx), file=sys.stderr)
+        if record.id in revised_lines_by_id:
+            record = revised_lines_by_id[record.id]
+        if record.info.get('SVTYPE', None) == 'DEL':
+            if abs(record.stop - record.pos) >= 1000:
+                sample_cn_map = {s: record.samples[s]['RD_CN'] for s in non_outlier_samples}
+                if len([s for s in sample_cn_map if (sample_cn_map[s] is not None and sample_cn_map[s] > 3)]) > vf_1:
+                    multi_del = True
+            gts = [record.samples[s]['GT'] for s in non_outlier_samples]
+            if len([gt for gt in gts if gt not in biallelic_gts]) > 0:
```
TedBrookings, line 101:
```suggestion
            if any(gt not in biallelic_gts for gt in gts):
```
Works with generator so faster, avoids memory alloc and dealloc.

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -0,0 +1,185 @@
...
+                gt5kb_del = True
+            if abs(record.stop - record.pos) >= 5000:
+                if not multi_del:
+                    gt5kb_del = True
+
+        if record.info.get('SVTYPE', None) == 'DUP':
+            if abs(record.stop - record.pos) >= 1000:
+                sample_cn_map = {s: record.samples[s]['RD_CN'] for s in non_outlier_samples}
+                if len([s for s in sample_cn_map if (sample_cn_map[s] is not None and sample_cn_map[s] > 4)]) > vf_1:
```
TedBrookings, line 110:
```suggestion
                if sum(1 for s in sample_cn_map if sample_cn_map[s] is not None and sample_cn_map[s] > 4) > vf_1:
```
Works with generator, no array alloc / dealloc

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -0,0 +1,185 @@
...
+                    multi_dup = True
+                if len([x for x in Counter(sample_cn_map.values()) if (x is not None and (x < 1 or x > 4))]) > 4:
```
TedBrookings, line 112:
```suggestion
                if sum(1 for x in Counter(sample_cn_map.values()) if x is not None and (x < 1 or x > 4)) > 4:
```

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -0,0 +1,185 @@
...
+                    gt4_copystate = True
+                if len([s for s in sample_cn_map if (sample_cn_map[s] is not None and
+                                                     (sample_cn_map[s] < 1 or sample_cn_map[s] > 4) and
+                                                     gt4_copystate)]) > vf_1:
```
TedBrookings, line 116:
```suggestion
                if sum(1 for s in sample_cn_map if sample_cn_map[s] is not None and
                                                     (sample_cn_map[s] < 1 or sample_cn_map[s] > 4) and
                                                     gt4_copystate) > vf_1:
```

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -0,0 +1,185 @@
...
+                    multi_dup = True
+            gts = [record.samples[s]['GT'] for s in non_outlier_samples]
+            if len([gt for gt in gts if gt not in biallelic_gts]) > 0:
```
TedBrookings, line 119:
```suggestion
            if any([gt not in biallelic_gts for gt in gts):
```

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -0,0 +1,185 @@
...
+        sexchr_revise = set([line.rstrip() for line in open(args.sexchr_revise, 'rt')])
+    print("{} sites to revise on sex chromosomes".format(len(sexchr_revise)), file=sys.stderr)
+
+    if args.multi_geno_ids_txt.endswith(".gz"):
+        multi_geno_ids = set([line.rstrip() for line in gzip.open(args.multi_geno_ids_txt, 'rt')])
+    else:
+        multi_geno_ids = set([line.rstrip() for line in open(args.multi_geno_ids_txt, 'rt')])
+    print("{} multiallelic sites".format(len(multi_geno_ids)), file=sys.stderr)
+
+    normal_vcf = pysam.VariantFile(args.normal_revise_vcf)
```
TedBrookings, line 55: It would be best practices to use a context manager (like you do at the top of the file). Since all the files you open are known at this point, you can do them all at once and only have one indent level:
```
with open("foo", 'r') as foo_in, pysam.VariantFile("bar.vcf", 'w') as f_out, pysam.VariantFile("snarf", 'r') as snarf_in:
    do_stuff()
```

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -0,0 +1,185 @@
...
+                    gt5kb_dup = True
+
+        if gt5kb_del:
+            for sample in record.samples:
+                if not record.samples[sample]['GQ'] is None and record.samples[sample]['RD_CN'] >= 2:
+                    record.samples[sample]['GT'] = (0, 0)
+                elif not record.samples[sample]['GQ'] is None and record.samples[sample]['RD_CN'] == 1:
+                    record.samples[sample]['GT'] = (0, 1)
+                elif not record.samples[sample]['GQ'] is None:
+                    record.samples[sample]['GT'] = (1, 1)  # RD_CN 0 DEL
```
TedBrookings, line 132:
```suggestion
            for sample_obj in record.samples.itervalues():
                if not sample_obj['GQ'] is None and sample_obj['RD_CN'] >= 2:
                    sample_obj['GT'] = (0, 0)
                elif not sample_obj['GQ'] is None and sample_obj['RD_CN'] == 1:
                    sample_obj['GT'] = (0, 1)
                elif not sample_obj['GQ'] is None:
                    sample_obj['GT'] = (1, 1)  # RD_CN 0 DEL
```
You can iterate over the sample record objects to avoid lots of dereferencing, which should provide a good speed-up.

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -0,0 +1,185 @@
...
+
+        if gt5kb_dup:
+            for sample in record.samples:
+                if not record.samples[sample]['GQ'] is None and record.samples[sample]['RD_CN'] <= 2:
+                    record.samples[sample]['GT'] = (0, 0)
+                elif not record.samples[sample]['GQ'] is None and record.samples[sample]['RD_CN'] == 3:
+                    record.samples[sample]['GT'] = (0, 1)
+                elif not record.samples[sample]['GQ'] is None:
+                    record.samples[sample]['GT'] = (1, 1)  # RD_CN > 3 DUP
```
TedBrookings, line 141:
```suggestion
            for sample_obj in record.samples.itervalues():
                if not sample_obj['GQ'] is None and sample_obj['RD_CN'] <= 2:
                    sample_obj['GT'] = (0, 0)
                elif not sample_obj['GQ'] is None and sample_obj['RD_CN'] == 3:
                    sample_obj['GT'] = (0, 1)
                elif not sample_obj['GQ'] is None:
                    sample_obj['GT'] = (1, 1)  # RD_CN > 3 DUP
```

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -0,0 +1,185 @@
...
+        if record.id in multi_geno_ids:
+            record.filter.add('PESR_GT_OVERDISPERSION')
+
+        if multi_del or multi_dup:
+            record.filter.add('MULTIALLELIC')
+            for j, sample in enumerate(record.samples):
+                record.samples[sample]['GT'] = None
+                record.samples[sample]['GQ'] = None
+                record.samples[sample]['CN'] = record.samples[sample]['RD_CN']
+                record.samples[sample]['CNQ'] = record.samples[sample]['RD_GQ']
```
TedBrookings, line 152:
```suggestion
            for j, sample_obj in enumerate(record.samples.itervalues()):
                sample_obj['GT'] = None
                sample_obj['GQ'] = None
                sample_obj['CN'] = sample_obj['RD_CN']
                sample_obj['CNQ'] = sample_obj['RD_GQ']
```

TedBrookings (commented):
I'm in python optimization mode since that's what I just finished in my own PR, so I looked over this just with an eye for speed-ups. I'm sure that all the suggestions will make it faster, but I don't know if any of them will be faster enough to make a real difference. Use them if you think they're helpful (or not).

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
cwhelan: I couldn't make this work for all the input and output files at once since the outputs need the modified header from the input, but I was able to use two nested `with` statements.

cwhelan:
@TedBrookings Thanks for the tips, I've added your suggestions. It doesn't seem like they changed the runtime much, if at all, on my large file tests but they do make the code cleaner. I'm afraid the tasks are just hopelessly I/O (and VCF record parsing) bound when deployed against very large files.

cwhelan:
I pushed several additional commits that, among other things, rework how the redundant multiallelic site list is calculated (testing against large file shards revealed several differences in my implementation compared to the old script). I believe that that my implementation is now producing the same results as the old script with the exceptions noted above.

wdl/CleanVcf5.wdl
```diff
@@ -1,1817 +1,302 @@
...
-    call CleanVcf5_29_TRUE_2 {
-      input:
-        malecols=CleanVcf5_29_TRUE_1.malecols,
-        sex_chr_revise=sex_chr_revise,
-        cleantagandmulti_vcf=CleanVcf5_27.cleantagandmulti_vcf,
-        sv_pipeline_docker=sv_pipeline_docker
+    call ScatterVcf as ScatterVcfPart5 {
+        input:
+            vcf=normal_revise_vcf,
+            records_per_shard = 5000,
```
mwalker174, line 28: Can you expose this to `MakeCohortVcf`?

wdl/CleanVcf5.wdl
```diff
@@ -1,1817 +1,302 @@
...
+        File? outlier_samples_list
+
+        File? shard_script
+        File? make_clean_gq_script
+        File? find_redundant_sites_script
+
+        String sv_base_mini_docker
+        String sv_pipeline_docker
+
+        RuntimeAttr? runtime_attr_override
```
mwalker174, line 22: Need to expose these separately for each task

wdl/CleanVcf5.wdl
```diff
@@ -1,1817 +1,302 @@
...
-    |egrep ^# \
-    |tail -n 1 \
-    |tr '\t' '\n' \
-    |cat -n - \
-    >col.txt
+task ScatterVcf {
+    input {
+        File vcf
+        File? shard_script
+        String prefix = "shard"
```
mwalker174, line 85: I'd make this required and set it to something like `~{cohort_name}.~{contig}.shard_clean_vcf_5`

wdl/CleanVcf5.wdl
```diff
@@ -1,1817 +1,302 @@
...
+        Int? records_per_shard = 5000
```
mwalker174, line 86: Also make required / no default value

wdl/CleanVcf5.wdl
```diff
@@ -1,1817 +1,302 @@
...
-    zcat ~{cleantagandmulti_vcf}\
-    |egrep ^# \
-    |tail -n 1 \
-    |tr '\t' '\n' \
-    |cat -n - \
-    >col.txt
+task ScatterVcf {
+    input {
+        File vcf
+        File? shard_script
```
mwalker174, line 84: Was this for debugging?

wdl/CleanVcf5.wdl
```diff
@@ -1,1817 +1,302 @@
...
+    }
 
-    awk '{if ($5==1) print $2}' ~{famfile} \
-    |{ fgrep -wf - col.txt || true; } \
-    >malecols.txt
-  >>>
+    Float input_size = size(vcf, "GB")
+    Float base_disk_gb = 10.0
+    Float input_mem_scale = 3.0
+    Float input_disk_scale = 5.0
```
mwalker174, line 94: I've been trying to avoid these since it's a bit harder to read and just write everything in-line in the object declaration below

src/sv-pipeline/04_variant_resolution/scripts/shard_pysam.py
```diff
@@ -0,0 +1,34 @@
...
+    for record in input_vcf:
+        shard = int(record_idx / records_per_shard)
+        if shard != current_shard:
+            if out_vcf is not None:
+                out_vcf.close()
+            out_vcf = VariantFile("{}.{:05d}.vcf.gz".format(args.shard_prefix, shard), 'w', header=input_vcf.header, threads=2)
+        current_shard = shard
+        out_vcf.write(record)
+        record_idx = record_idx + 1
+    out_vcf.close()
```
mwalker174, line 29: I don't think this would happen in cleanvcf5, but if we wanted to use this script generally I think this will crash if there are 0 records (`out_vcf` is never initialized). I think it's also better to emit a single empty shard if there are no records, since our concat task fails when there are 0 shards.

wdl/CleanVcf5.wdl
```diff
@@ -1,1817 +1,302 @@
...
+
+        cat ~{redundant_multiallelics_list} ~{sep=" " no_sample_lists} > ids_to_remove.list
+        /usr/local/bin/bcftools concat --no-version --output-type u --file-list ~{write_lines(clean_gq_vcfs)} | \
+            /usr/local/bin/bcftools view --no-version \
+                --exclude 'ID=@ids_to_remove.list' \
+                --output-type z -o polished.need_reheader.vcf.gz --threads 3
+
+        # do the last bit of header cleanup
+        bcftools view -h polished.need_reheader.vcf.gz | awk 'NR == 1' > new_header.vcf
+        ls -l new_header.vcf
```
mwalker174, line 289: Debugging line?

mwalker174 (commented):
A few little comments. In places where you've used multithreading, was that optimized for gnomad? Might be good to expose those as parameters in case we use smaller VMs normally

cwhelan:
@mwalker174 I've tried to make the changes you requested. I also did some more experimentation with `bcftools scatter` and ended up replacing my psyam sharding script with that.

Re: the threading; I added threading after adopting @TedBrookings new docker image with the latest pysam. It doesn't seem to have a big impact on runtime, so I could take it out if you'd prefer less complexity, but since I'm requesting 16G of RAM (which seems to be required for my pysam scripts) I figured we might as well use the extra cores which are going to be available, even if they don't make a big difference.

I have not rebuilt the dockers, partly because I'm using the new pysam version with the thread argument and the Dockerfile for `sv-pipeline-updates` isn't checked in yet, so I've been running and testing using localized versions of my scripts instead. I also figured we might as well wait to rebuild dockers until all these rewrites are in. But if you'd like me to do so (and drop the threads argument) let me know.

PR #232, opened 2021-09-16
Rewrite resolve_cpx_cnv_redundancies

TedBrookings:
Replace resolve_CPX_CNV_redundancies.sh with resolve_cpx_cnv_redundancies.py
* Replace bash, python, R with python
* Increase speed
* * Create new temporary docker sv-pipeline-updates with pysam 0.15.4
    allowing multithreaded compression/decompression of vcf files
* * new docker also has bcftools/htslib 1.13
* Avoid filling cromwell boot disk with option to set base temp dir
* Fix several bugs with incorrect processing of SVs
* * Fix in original script to demonstrate equivalence of new script
* * * Process variants with SVTYPE=CNV
* * * Determine if interval is CPX by keeping track of original SVTYPE
      instead of using variant ID
* * * Fix error in computing sample overlap
* * * Select best genotype using CN instead of GT for CNVs
* * Additional fixes in svtk
* * * In vcf2bed, yield sink for complex insertions
* * * In bedcluster, compute connections correctly if variant IDs repeat


TedBrookings:
This one is a little weird, so a quick explanation:
I altered the base algorithm of resolve_CPX_CNV_redundancies.sh to avoid unnecessary calculations while achieving the same output. However there were also numerous bugs. Thus I've included the old code with bug fixes, and the new python code which produces identical output (up to pysam rearranging header lines). I would presumably remove the old code upon approval. I can alter the structure of this PR if this is causing problems.

I'm also currently running the new code in cromwell one last time, hoping to refine the memory requirements for the task.

PR #233, opened 2021-09-16
Handle records with empty EVIDENCE field in MakeCohortVcf

epiercehoffman:
### Updates
Add handling for records with empty EVIDENCE fields in overlap_breakpoint_filter.py in MakeCohortVcf.ResolveComplexVariants.BreakpointOverlap. Records have empty EVIDENCE fields if they are all-ref but "salvaged."

Unrelatedly, also added pipeline diagram to Terra dashboard. Live on 1kgp workspace.

### Testing
Tested with 1kgp batches on Terra with updated docker image - successfully ran where it had previously failed on chrY.


input_templates/terra_workspaces/cohort_mode/cohort_mode_workspace_dashboard.md.tmpl
```diff
@@ -47,6 +47,8 @@ The following are the main pipeline outputs. For more information on the outputs
 
 ### Pipeline overview
 
+<img alt="pipeline_diagram" title="Pipeline diagram" src="https://i.imgur.com/ZTBjo1i.jpg" width="1000">
```
mwalker174, line 50: It would be better to host this in the repo, can you put it in the root dir?

mwalker174 (approved):
Thanks! Pegging the diagram to a release rather than master seems like the way to go, as long as we can remember to update the URL with any updates to the image.

PR #244, opened 2021-10-20
Update the Github action to build only the images impacted by the files changed in a PR

VJalili:
The `sv_pipeline_docker.yml` triggers building all the docker images if a PR changes any files defined in `paths`. In other words, it always uses `--targets all` when calling `build_docker.py`. However, since rebuilding all the docker images takes a considerable amount of time (about 1h) and rebuilding all may not be necessary depending on the files changed in a PR, this PR improves the `sv_pipeline_docker` by setting `--targets` to only the images impacted by the files changed in the PR.

The action currently builds the images according to the following table: 

| Image                  | When to build                                                |
|------------------------|--------------------------------------------------------------|
| Delly                  | if the docker file is changed                                |
| Manta                  | if the docker file is changed                                |
| Wham                   | if the docker file is changed                                |
| sv-base-mini           | if the docker file is changed                                |
| sv-base                | if the docker file is changed                                |
| samtools-cloud         | if the docker file is changed                                |
| cnmops                 | /src/WGD                                                     |
| sv-pipeline-base       | /src/svtk /src/sv-pipeline /src/svtest /src/svqc             |
| sv-pipeline-children-r | /src/svtk /src/sv-pipeline /src/svtest /src/svqc             |
| sv-pipeline            | /src/svtk /src/sv-pipeline /src/svtest /src/svqc             |
| sv-pipeline-rdtest     | /src/svtk /src/sv-pipeline /src/svtest /src/svqc /src/RdTest |
| sv-pipeline-qc         | /src/svtk /src/sv-pipeline /src/svtest /src/svqc /src/WGD    |

### DIFF
--- a/.github/workflows/sv_pipeline_docker.yml
+++ b/.github/workflows/sv_pipeline_docker.yml
@@ -90,7 +90,7 @@ jobs:
             if [[ $i == *".github/workflows/sv_pipeline_docker.yml"* ]]; then
               # A change to this file may impact how images are built,
               # hence, if this file is changed, all the images are rebuilt.
-              try_add_target "all"
+              TARGETS=("all")
               break
             elif [[ $i == *"src/svtk"* ||
                   $i == *"src/sv-pipeline"* ||
--- a/.github/workflows/sv_pipeline_docker.yml
+++ b/.github/workflows/sv_pipeline_docker.yml
@@ -56,11 +56,9 @@ jobs:
           # The commit SHA of the base branch which this PR targets;
           # ideally broadinstitute/gatk-sv:master.
           BASE_COMMIT_SHA=$(echo "$GITHUB_CONTEXT"| jq -r '.event.pull_request.base.sha')
-          echo "::debug::BASE_COMMIT_SHA: $BASE_COMMIT_SHA"
 
           # The commit SHA of the commit that triggered the action.
           COMMIT_SHA=${{ github.event.pull_request.head.sha }}
-          echo "::debug::COMMIT_SHA: $COMMIT_SHA"
 
           # A list of all the files changed between the current commit & base.
           CHANGED_FILES=$(git diff --name-only $BASE_COMMIT_SHA $COMMIT_SHA)
@@ -73,6 +71,12 @@ jobs:
           }
           for i in "${CHANGED_FILES[@]}"
           do
+            if [[ $i == *".github/workflows/sv_pipeline_docker.yml" ]]; then
+              # A change to this file may impact how images are built,
+              # hence all the images are rebuilt.
+              try_add_target "all"
+              break
+            fi
             if [[ $i == *"src/svtk"* ||
                   $i == *"src/sv-pipeline"* ||
                   $i == *"src/svtest"* ||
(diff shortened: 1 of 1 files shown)


.github/workflows/sv_pipeline_docker.yml
```diff
@@ -43,6 +50,92 @@ jobs:
...
+              TARGETS+=($1)
+            fi
+          }
+
+          for i in "${CHANGED_FILES[@]}"
+          do
+            if [[ $i == *".github/workflows/sv_pipeline_docker.yml"* ]]; then
+              # A change to this file may impact how images are built,
+              # hence, if this file is changed, all the images are rebuilt.
+              try_add_target "all"
```
TedBrookings, line 93: Right now build_docker.py throws an error if "all" is in the targets list any anything else is too, which seems like it could happen depending on the order of changed files.. So this should be changed to
```suggestion
              TARGETS=("all")
```

TedBrookings (approved):
This script accomplishes a lot in bash, and in a weird environment to boot. It's well-commented, which is necessary because otherwise it would be tough to follow (because bash). I'm approving it, and don't see any obvious avenue for improvement with this approach other than the one comment I made addressing a corner-case. However, I'm going to take this opportunity to suggest a future approach:

1) Introduce new command-line arguments to build_docker.py `--base-git-sha` and `--head-git-sha`. If `base-git-sha` is not specified, build_docker.py expects to be passed targets like usual, so everything works as before. But if it is, then all the target determination gets done inside build_docker.py. If `head-git-sha` is not specified, the diff is just vs the current status of the git repo.
2) Dependencies that don't need to be rebuilt should have the docker image corresponding to `base-git-sha` tagged with `head-git-sha` if possible. Currently the action will be to rebuild them anyway, even though they have not changed (because they are a dependency and a docker image with the correct tag does not exist).

The advantage of this approach is that

1. python is in my opinion much easier to understand and maintain
2. fewer images will be built, because sometimes dependency images can be re-tagged instead of built
3. this approach could be used locally during development work
4. it would allow eventual simplification of this github action (down to simply passing `git-base-sha` and `git-head-sha`).

.github/workflows/sv_pipeline_docker.yml
VJalili: Good point, thank you.

VJalili:
Thank you, @TedBrookings! I agree a good next step is to move the procedure determining the list of changed files from this action to `build_docker.py`.

mwalker174 (approved):
LGTM. I agree with @TedBrookings that incorporating this functionality into build_docker.py would be a cleaner, more portable solution. 

In addition, I'm wondering if there are ways we could speed up our docker builds. Experimenting with `docker-compose` and `docker container commit` may enable us to load source changes without having to install all dependencies. If we only had to copy the source in (assuming no dependencies change), we could quickly build all dockers and not have to worry about selective builds.

VJalili:
We might also be able to shorten the build time if we could parallelize builds. In terms of GitHub actions, one possible approach would be defining separate jobs for each image build.

PR #250, opened 2021-10-29
Add subset ped step at start of CleanVcf

epiercehoffman:
### Updates
Add step to subset ped file to only sample IDs that are in the input VCF at the start of CleanVcf. This is a patch to avoid a bug that can occur in CleanVcf1_20 when there are extra samples in the ped file (that are not in the VCF) whose sex is listed as a value other than 1 or 2 (under some circumstances). This subset step will become unnecessary and should be removed once the rewrite of CleanVcf, which uses the sample IDs in the VCF as a starting point, is merged.

### Testing
* Successfully ran MakeCohortVcf.wdl with test_large dataset with extra sex=0 samples in the PED file. 
  * Caveat: based on the outputs of CleanVcf1_5, CleanVcf1_6 and subsequently CleanVcf1_20 were skipped. Please advise if further testing is recommended. It has already been confirmed that manually subsetting the PED file to remove extra samples solved the problem for the user that encountered it.

### DIFF
--- a/wdl/CleanVcf.wdl
+++ b/wdl/CleanVcf.wdl
@@ -2,6 +2,7 @@ version 1.0
 
 import "CleanVcfChromosome.wdl" as CleanVcfChromosome
 import "TasksMakeCohortVcf.wdl" as MiniTasks
+import "Utils.wdl" as util
 
 workflow CleanVcf {
   input {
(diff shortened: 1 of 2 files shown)


mwalker174 (approved):
I think this looks good. Let's let the user test the change after you cut another Terra release.

PR #252, opened 2021-11-01
Add Dockerfile for PanGenie

VJalili:
This PR adds a Dockerfile for [PanGenie](https://bitbucket.org/jana_ebler/pangenie/src/master/).

### DIFF
--- a/dockerfiles/pangenie/Dockerfile
+++ b/dockerfiles/pangenie/Dockerfile
@@ -17,7 +17,7 @@
 # 'continuumio/miniconda3' as base image.
 #   https://pythonspeed.com/articles/activate-conda-dockerfile/
 
-FROM continuumio/miniconda3
+FROM continuumio/miniconda3:4.10.3p0
 
 RUN apt-get update && apt-get install --no-install-recommends -qqy \
     build-essential \
--- a/dockerfiles/pangenie/Dockerfile
+++ b/dockerfiles/pangenie/Dockerfile
@@ -41,7 +41,8 @@ SHELL ["/bin/bash", "--login", "-c"]
 # Install PanGenie
 # Without setting `PKG_CONFIG_PATH`, cmake/make cannot resolve
 # jellyfish dependency.
-RUN conda install -y jellyfish && \
+# The jellyfish version is pinned to its current latest release.
+RUN conda install -y jellyfish=2.2.10 -c bioconda && \
     export PKG_CONFIG_PATH="/opt/conda/envs/pangenie/lib/pkgconfig/" && \
     cd pangenie && \
     mkdir build && \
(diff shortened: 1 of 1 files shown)


dockerfiles/pangenie/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+#   RUN bash miniconda.sh -b -p /opt/conda && \
+#       /opt/conda/bin/conda init bash && \
+#       bash -c "source ${HOME}/.bashrc"
+#
+# A similar issue is detailed in the following blog post; therefore, as
+# suggested in this blog post, this docker files is using
+# 'continuumio/miniconda3' as base image.
+#   https://pythonspeed.com/articles/activate-conda-dockerfile/
+
+FROM continuumio/miniconda3
```
mwalker174, line 20: Can you peg this to a specific tag?

dockerfiles/pangenie/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+RUN apt-get update && apt-get install --no-install-recommends -qqy \
+    build-essential \
+    git \
+    make \
+    cmake \
+    wget \
+    pkg-config
+
+RUN git clone https://jana_ebler@bitbucket.org/jana_ebler/pangenie.git && \
+    cd pangenie && \
```
mwalker174, line 31: Can you peg this to the current release?

dockerfiles/pangenie/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+    conda env create -f environment.yml
+
+# Make RUN commands use the new environment:
+RUN echo "conda activate pangenie" >> ~/.bashrc
+SHELL ["/bin/bash", "--login", "-c"]
+
+# Install PanGenie
+# Without setting `PKG_CONFIG_PATH`, cmake/make cannot resolve
+# jellyfish dependency.
+RUN conda install -y jellyfish && \
```
mwalker174, line 41: Also peg jellyfish version

mwalker174 (commented):
Thank you @VJalili! I would suggest specifying specific versions for most of the installs. Also could you add a quick cleanup step at the end to delete any temp files?

dockerfiles/pangenie/Dockerfile
VJalili: Good point, thanks! I pinned the docker image and all the tools installed to their current latest tag and versions respectively.

VJalili:
Thank you @mwalker174 for the suggestions.

I added a cleanup code. The image size is `1.3GB` compressed, and the cleanup code does not change it noticeably. The largest directories are those related to conda, and [it seems that can be reduced through a multi-stage build](https://pythonspeed.com/articles/conda-docker-image-size/). IMHO, this can be a good optimization when improving this Docker image for a large-scale execution.

mwalker174 (approved):
Thank you for incorporating my suggestions. I'd also suggest adding a build of this to `input_values/dockers.json` even though there is no WDL for this yet, just so we know where to find a build of this.

VJalili:
Thank you for reviewing this @mwalker174.

PR #253, opened 2021-11-03
Add a WDL to run ExpansionHunter on multiple samples

VJalili:
This PR adds `ExpansionHunterScatter.wdl` that takes multiple bam/cram as input and runs them on separate instances of ExpansionHunter in parallel.

### DIFF
--- a/wdl/ExpansionHunterScatter.wdl
+++ b/wdl/ExpansionHunterScatter.wdl
@@ -1,0 +1,47 @@
+version 1.0
+
+import "Structs.wdl"
+import "ExpansionHunter.wdl" as ExpansionHunter
+
+workflow ExpansionHunterScatter {
+
+    input {
+        Array[File] bams_or_crams
+        Array[File]? bams_or_crams_indexes
+        File reference_fasta
+        File? reference_fasta_index
+        File variant_catalog
+        String eh_docker
+        RuntimeAttr? runtime_attr
+    }
+
+    scatter (i in range(length(bams_or_crams))) {
+        File bam_or_cram_ = bams_or_crams[i]
+        Boolean is_bam =
+            basename(bam_or_cram_, ".bam") + ".bam" == basename(bam_or_cram_)
+        File bam_or_cram_index_ =
+            if defined(bams_or_crams_indexes) then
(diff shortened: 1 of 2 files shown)


wdl/ExpansionHunterScatter.wdl
```diff
@@ -0,0 +1,47 @@
...
+            reference_fasta_index, reference_fasta + ".fai"])
+
+        call ExpansionHunter.ExpansionHunter as expanionHunter {
+            input:
+                bam_or_cram=bam_or_cram_,
+                bam_or_cram_index=bam_or_cram_index_,
+                reference_fasta=reference_fasta,
+                reference_fasta_index=reference_fasta_index_,
+                variant_catalog=variant_catalog,
+                docker_file=eh_docker,
```
mwalker174, line 37: For consistency and clarity, can you make these variables both `expansion_hunter_docker`?

wdl/ExpansionHunterScatter.wdl
```diff
@@ -0,0 +1,47 @@
+version 1.0
+
+import "Structs.wdl"
+import "ExpansionHunter.wdl" as ExpansionHunter
+
+workflow ExpansionHunterScatter {
+
+    input {
+        Array[File] bams_or_crams
```
mwalker174, line 9: Regarding output prefixes: often bam/cram filenames can be difficult to interpret at a glance. I would suggest adding an optional input `Array[String]? sample_ids` that are used as the output prefix for each sample, if provided. This can also guarantee that, within a given cohort, the filenames are unique, which can be helpful for managing data downstream.

wdl/ExpansionHunterScatter.wdl
VJalili: Good point, I renamed them according.

wdl/ExpansionHunterScatter.wdl
VJalili: That would be very useful, thank you. I added the optional list.

mwalker174 (approved):
Thanks for addressing my comments - looks good!

VJalili:
Thank you @mwalker174

PR #254, opened 2021-11-05
Get male-only var IDs from ped file and GTs not log p-value

epiercehoffman:
### Updates
Issue: Previous method of determining which variants on chrX appeared only in males and should use evidence from males instead of (preferentially, by default) females caused issues with some variants in SRTest that had a log p-value of 0 for one position (start or end) but not the other, resulting in mixing and matching SR coordinates.

Fix: Determine male-only variant IDs for each caller VCF from GTs and ped file and supply the list to SRTest, PETest, and RDTest. Use male evidence only for variant IDs appearing in the male-only list.

### Testing
Tested with one full-genome gnomAD batch and spot-checked that one variant was properly rescued (compared to original gnomAD run) and one variant had the correct matched SR coordinates (compared to previous master). Also confirmed no changes other than on chrX.


mwalker174 (approved):
Thank you @epiercehoffman, looks good from here.

PR #255, opened 2021-11-09
Expose runtime attributes in AnnotateVcf

epiercehoffman:
### Updates
Expose runtime attributes in AnnotateVcf.wdl and subworkflows. Requested by an internal user who ran out of memory in `AnnotateVcf.AnnotateExternalAF.split_query_vcf` and was not able to increase the memory in Terra.

### Testing
* Validated AnnotateVcf.wdl with womtool. 
* Did not run a test because we lack test data for AnnotateVcf.wdl in the repo, but I have another PR coming soon that will hopefully address that, and the user can also test the fix and report back.


mwalker174 (approved):
LGTM. I don't see any areas where there would be errors at runtime, so validating with womtool should be fine here.

PR #256, opened 2021-11-15
Update VCFs in AnnotateVcf test JSON

epiercehoffman:
### Updates
Generated a new clean VCF from MakeCohortVcf with test_large data and added to test_large input_values JSON. Use this VCF and index file as inputs to AnnotateVcf test JSON in place of an internal dataset. 
* Note that a new clean VCF was needed because the baseline_cleaned_vcf contained MCNV records which broke AnnotateVcf.
* Note that the number of JSONs to validate is reduced to 42 from 43 after this change, because an AnnotateVcf JSON is only created for test_large, not test_small, which doesn't have a clean_vcf attribute.

### Testing
* AnnotateVcf.wdl ran successfully with the new input files


mwalker174:
While you're in there, can you move everything from `gs://talkowskicromwellresults` into `gs://gatk-sv-resources`?

epiercehoffman:
I moved the cohort depth and PESR test_large VCFs, and the other files were former inputs to RegenotypeCNVs/Module04b so I just deleted them from the test_large input values JSON. I didn't clean out the `gs://talkowskicromwellresults` bucket itself though because there's a lot of stuff in there.

mwalker174 (approved):

PR #263, opened 2021-11-18
Subset to chrX for GetMaleOnlyVariantIDs

epiercehoffman:
### Updates
Subset to chrX for GetMaleOnlyVariantIDs for space and memory efficiency, as the only variants for which the male-only status matters are on chrX. 
Note: bcftools view -t was used in place of -r because the input VCFs do not appear to be indexed. It is still highly efficient.

### Testing
* Tested one gnomAD batch - it finished successfully, and all of the GetMaleOnlyVariantIDs outputs were identical to the chrX portion of the previous master outputs. Note that I did not compare the final module outputs directly because the gnomAD processing was done on a branch with other changes to enable processing chrX only.

### DIFF
--- a/wdl/GATKSVPipelineSingleSample.wdl
+++ b/wdl/GATKSVPipelineSingleSample.wdl
@@ -292,6 +292,7 @@ workflow GATKSVPipelineSingleSample {
 
     File rmsk
     File segdups
+    String? chr_x
 
     Int? min_large_pesr_call_size_for_filtering
     Float? min_large_pesr_depth_overlap_fraction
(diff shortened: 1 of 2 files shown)


mwalker174 (changes requested):
Can you make chrX a parameter `String contig`, which is set by a `String chr_x` workflow input? Want to support our dog genome friends as well :). Also makes it clear that this only grabs VIDs from a particular contig.

epiercehoffman:
Testing for latest commit: 
* Re-validated all WDLs and JSONs with womtool
* Re-ran GenerateBatchMetrics successfully and verified the outputs of GetMaleOnlyVariantIDs were unchanged and the subsequent steps cached
* Did not test the single-sample pipeline because it validated and the changes were identical to those in GenerateBatchMetrics

mwalker174 (approved):
Thanks!

PR #267, opened 2021-11-29
Filter sites with NA score

mwalker174:
Fixes an issue in `FilterBatch` where variants with a score of `NA` were not being filtered. This case seems to only affect depth-only calls and occurs for one of three reasons:

1. all samples are non-ref, 
2. all samples are ref,
3. or insufficient read count bins

The cause in most cases is (2). In our 1KGP panel (batch 1), this eliminates 1518 sites, 1488 of which are on chrY. We currently allow raw calls on chrY in female samples, but `GenerateBatchMetrics` only considers males. Therefore there are many depth-only / female-only CNVs with all-ref genotypes when subsetted to male samples.

### DIFF
--- a/wdl/FilterBatchSites.wdl
+++ b/wdl/FilterBatchSites.wdl
@@ -186,8 +186,8 @@ task FilterAnnotateVcf {
 
     set -euo pipefail
     cat \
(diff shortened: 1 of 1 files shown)


epiercehoffman (approved):
Looks good to me

PR #279, opened 2022-01-07
Scatter expansion hunter on batches of variant catalogs

VJalili:
This PR splits a given JSON file containing a catalog of variants into smaller JSON files each containing a maximum of  `variant_catalog_batch_size` variants (defaults to `10,000`). Then calls expansion hunter on each of the input samples using the split variant catalogs separately. Hence, running expansion hunter for `n * m` times where `n` and `m` are respectively the number of input samples and the number of split JSON files.

### DIFF
--- a/wdl/ExpansionHunter.wdl
+++ b/wdl/ExpansionHunter.wdl
@@ -3,6 +3,7 @@
 version 1.0
 
 import "Structs.wdl"
+import "TasksMakeCohortVcf.wdl" as MiniTasks
 
 struct FilenamePostfixes {
     String locus
@@ -22,8 +23,10 @@ workflow ExpansionHunter {
         File variant_catalog
         File? output_prefix
         String expansion_hunter_docker
+        String sv_base_mini_docker
         Int variant_catalog_batch_size
         RuntimeAttr? runtime_attr
+        RuntimeAttr? runtime_override_concat
     }
 
     Boolean is_bam = basename(bam_or_cram, ".bam") + ".bam" == basename(bam_or_cram)
--- a/wdl/ExpansionHunter.wdl
+++ b/wdl/ExpansionHunter.wdl
@@ -183,7 +183,7 @@ task ConcatEHOutputs {
         jq -s 'reduce .[] as $item ({}; . * $item)' ~{sep=" " jsons} > ~{output_prefix}.json
 
         VCFS="~{write_lines(vcfs_gz)}"
-        bcftools concat --no-version --naive --naive-force --output-type z --file-list ${VCFS} --output "~{output_prefix}.vcf.gz"
+        bcftools concat --no-version --naive-force --output-type z --file-list ${VCFS} --output "~{output_prefix}.vcf.gz"
 
         BAMS="~{write_lines(overlapping_reads)}"
         samtools merge ~{output_prefix}.bam -b ${BAMS}
--- a/wdl/ExpansionHunter.wdl
+++ b/wdl/ExpansionHunter.wdl
@@ -3,7 +3,6 @@
 version 1.0
 
 import "Structs.wdl"
-import "TasksMakeCohortVcf.wdl" as MiniTasks
 
 struct FilenamePostfixes {
     String locus
@@ -70,11 +69,21 @@ workflow ExpansionHunter {
         }
     }
 
+    call ConcatEHOutputs {
+        input:
+            vcfs_gz = RunExpansionHunter.vcf_gz,
+            jsons = RunExpansionHunter.json,
+            overlapping_reads = RunExpansionHunter.overlapping_reads,
+            timings = RunExpansionHunter.timing,
+            output_prefix = output_prefix_,
+            expansion_hunter_docker = expansion_hunter_docker
+    }
+
     output {
-        Array[File] json = RunExpansionHunter.json
-        Array[File] vcf = RunExpansionHunter.vcf
-        Array[File] overlapping_reads = RunExpansionHunter.overlapping_reads
-        Array[File] timing = RunExpansionHunter.timing
+        File json = ConcatEHOutputs.json
+        File vcf_gz = ConcatEHOutputs.vcf_gz
+        File overlapping_reads = ConcatEHOutputs.overlapping_reads
+        File timing = ConcatEHOutputs.timing
     }
 }
 
@@ -92,7 +101,7 @@ task RunExpansionHunter {
 
     output {
         File json = "${output_prefix}.json"
-        File vcf = "${output_prefix}.vcf"
+        File vcf_gz = "${output_prefix}.vcf.gz"
         File overlapping_reads = "${output_prefix}_realigned.bam"
         File timing = "${output_prefix}_timing.tsv"
     }
@@ -107,6 +116,8 @@ task RunExpansionHunter {
             --output-prefix ~{output_prefix} \
             --cache-mates \
             --record-timing
+
+        bgzip ~{output_prefix}.vcf
     >>>
 
     RuntimeAttr runtime_attr_str_profile_default = object {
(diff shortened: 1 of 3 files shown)


wdl/ExpansionHunter.wdl
```diff
@@ -142,3 +165,145 @@ task RunExpansionHunter {
         maxRetries: runtime_attr.max_retries
     }
 }
+
+task SplitVariantCatalog {
+    input {
+        File variant_catalog
+        Int batch_size
+        String? split_variant_catalog_docker
+        String? output_prefix
```
mwalker174, line 174: I think it's better to make these required. it's easier to keep track of what dockers are being used across the whole pipeline by just being able to reference `/inputs/dockers.json` rather than having to comb through all the WDLs. I'd also argue for a required prefix since, as a general practice, it tends to be easier to keep track of output files (i.e. when we have to copy them to a single directory for debugging) when the output names are unique.
```suggestion
        String? python_docker
        String? output_prefix
```

wdl/ExpansionHunter.wdl
```diff
@@ -42,24 +44,43 @@ workflow ExpansionHunter {
         reference_fasta_index,
         reference_fasta + ".fai"])
 
-    call RunExpansionHunter {
+    call SplitVariantCatalog {
```
mwalker174, line 47: Since this is the same for every sample, definitely want to move this to the `Scatter` wdl and run outside of the scatter and pass in the scattered catalog to this wdl.

wdl/ExpansionHunter.wdl
```diff
@@ -142,3 +165,145 @@ task RunExpansionHunter {
...
+            subset_catalogs = []
+            for catalog in catalogs:
+                subset_catalogs.append(catalog)
+                i += 1
+                if i == ~{batch_size}:
+                    serialize()
+                    i = 0
+                    subset_counter += 1
+                    subset_catalogs = []
+            serialize()
```
mwalker174, line 226: I think this might get run even if `subset_catalogs` is empty, so you'd want to check that that's not the case.

wdl/ExpansionHunter.wdl
```diff
@@ -142,3 +165,145 @@ task RunExpansionHunter {
...
+        else
+            "python:slim-buster"
+
+    command <<<
+        set -euxo pipefail
+
+        python <<CODE
+        import json
+        import sys
+        import os
```
mwalker174, line 200:
```suggestion

```

wdl/ExpansionHunter.wdl
```diff
@@ -142,3 +165,145 @@ task RunExpansionHunter {
...
+            select_first([split_variant_catalog_docker])
+        else
+            "python:slim-buster"
+
+    command <<<
+        set -euxo pipefail
+
+        python <<CODE
+        import json
+        import sys
```
mwalker174, line 199:
```suggestion

```

wdl/ExpansionHunter.wdl
```diff
@@ -142,3 +165,145 @@ task RunExpansionHunter {
...
+        File timing = "${output_prefix}_timing.tsv"
+    }
+
+    command <<<
+        set -euxo pipefail
+
+        jq -s 'reduce .[] as $item ({}; . * $item)' ~{sep=" " jsons} > ~{output_prefix}.json
+
+        VCFS="~{write_lines(vcfs_gz)}"
+        bcftools concat --no-version --naive --naive-force --output-type z --file-list ${VCFS} --output "~{output_prefix}.vcf.gz"
```
mwalker174, line 277:
```suggestion
        bcftools concat --no-version --naive-force --output-type z --file-list ${VCFS} --output "~{output_prefix}.vcf.gz"
```

wdl/ExpansionHunter.wdl
```diff
@@ -142,3 +165,145 @@ task RunExpansionHunter {
...
+
+        BAMS="~{write_lines(overlapping_reads)}"
+        samtools merge ~{output_prefix}.bam -b ${BAMS}
+
+        TIMINGS=(~{sep=" " timings})
+        FIRST_TIMING=${TIMINGS[0]}
+        head -1 ${FIRST_TIMING} > ~{output_prefix}_timing.tsv
+        tail -n +2 -q ~{sep=" " timings} >> ~{output_prefix}_timing.tsv
```
mwalker174, line 285: I've found using `tail -n +N` to be inefficient (I think it has to load the whole output into memory). Are the header lines indicated somehow, i.e. beginning with `#`? If so I would rewrite like `cat ~{sep=" " timings} | grep -v ^# >> ~{output_prefix}_timing.tsv`

wdl/ExpansionHunter.wdl
```diff
@@ -142,3 +165,145 @@ task RunExpansionHunter {
...
+    input {
+        File variant_catalog
+        Int batch_size
+        String? split_variant_catalog_docker
+        String? output_prefix
+        RuntimeAttr? runtime_attr_override
+    }
+
+    output {
+        Array[File] catalogs_json = glob("${output_prefix_}*.json")
```
mwalker174, line 179: One tricky thing about globs is maintaining order. If your outputs are `output_prefix_0.json`, `output_prefix_1.json`, `output_prefix_2.json`, etc. then `glob` will sort lexicographically and could mess up the order e.g. if you have more than 10 shards: `output_prefix_0.json`, `output_prefix_1.json`, `output_prefix_10.json`, `output_prefix_11.json`, etc. This could cause problems for your concatenation later.

The workaround we usually do is to pad the index with 0's, e.g. `output_prefix_000.json`, `output_prefix_001.json`, `output_prefix_002.json` etc. In theory you could figure out the number of necessary padding digits by counting the number of entries before doing the splitting. To me that seems a little complicated - I think you can just pad with 6 digits and call it a day.

wdl/ExpansionHunterScatter.wdl
```diff
@@ -22,6 +23,12 @@ workflow ExpansionHunterScatter {
         sample_ids: "One ID per sample, in the same order as the files in bams_or_crams. These IDs must match the ID given in the second column (`Individual ID` column) of the given PED file. These IDs will also be used as an output prefix."
     }
 
+    String variant_catalog_batch_size_ =
+        if defined(variant_catalog_batch_size) then
+            select_first([variant_catalog_batch_size])
+        else
+            10000
```
mwalker174, line 30:
```suggestion
    String variant_catalog_batch_size_ = select_first([variant_catalog_batch_size, 1000])
```

wdl/ExpansionHunter.wdl
```diff
@@ -142,3 +165,145 @@ task RunExpansionHunter {
...
+        if defined(output_prefix) then
+            select_first([output_prefix])
+        else
+            "output_"
+
+    String split_variant_catalog_docker_ =
+        if defined(split_variant_catalog_docker) then
+            select_first([split_variant_catalog_docker])
+        else
+            "python:slim-buster"
```
mwalker174, line 192: See comment about `variant_catalog_batch_size_` for a shorter way to write these.

wdl/ExpansionHunter.wdl
VJalili: Good suggestions, thank you! changed accordingly.

wdl/ExpansionHunter.wdl
VJalili: That will lead to splitting variant catalog only once, compared to the current implementation that splits for every sample. It is a very good improvement, thanks for the suggestion.

wdl/ExpansionHunter.wdl
VJalili: Good catch, thanks! I added a condition check, hopefully, that will prevent serialising an empty array.

wdl/ExpansionHunter.wdl
VJalili: That would be ideal if the header was starting with `#`. Do you think awk would be any faster/readable?!

wdl/ExpansionHunter.wdl
VJalili: Thanks!

wdl/ExpansionHunter.wdl
VJalili: Good point, thanks! I added padding to the counter.

I'm actually not sure about the necessity of `output_prefix` for this task. It results in filenames such as `variant_catalog.jsonvariant_catalog_subset_000000.json`. Do you think it would be safe to remove `output_prefix`?

wdl/ExpansionHunterScatter.wdl
```diff
@@ -42,18 +54,90 @@ workflow ExpansionHunterScatter {
...
+
+    runtime {
+        docker: python_docker
+        cpu: runtime_attr.cpu_cores
+        memory: runtime_attr.mem_gb + " GiB"
+        disks: "local-disk " + runtime_attr.disk_gb + " HDD"
+        bootDiskSizeGb: runtime_attr.boot_disk_gb
+        preemptible: runtime_attr.preemptible_tries
+        maxRetries: runtime_attr.max_retries
+    }
```
mwalker174, line 142: Didn't notice this before, but this looks like a simplified version of what we usually do. The `RuntimeAttr` struct looks like this:
```
struct RuntimeAttr {
    Float? mem_gb
    Int? cpu_cores
    Int? disk_gb
    Int? boot_disk_gb
    Int? preemptible_tries
    Int? max_retries
}
```
So each member of the struct is Optional. This is because the user is meant to be able to provide a subset of the members, for example to only change the memory they would provide:
```
"runtime_attr_my_task": { "mem_gb": "8" },
```
The way it's written here, the workflow would crash because it would attempt to use the undefined fields (`runtime_attr.cpu_cores`, `runtime_attr.disk_gb`, etc.). This is why we use a more complicated looking runtime section like this for most tasks:
```
  runtime {
    cpu: select_first([runtime_attr.cpu_cores, default_attr.cpu_cores])
    memory: mem_gb + " GiB"
    disks: "local-disk " + select_first([runtime_attr.disk_gb, default_attr.disk_gb]) + " HDD"
    bootDiskSizeGb: select_first([runtime_attr.boot_disk_gb, default_attr.boot_disk_gb])
    docker: sv_pipeline_docker
    preemptible: select_first([runtime_attr.preemptible_tries, default_attr.preemptible_tries])
    maxRetries: select_first([runtime_attr.max_retries, default_attr.max_retries])
  }
```
Here it will properly select overrides if they are provided, but use defaults if they are missing.

wdl/ExpansionHunterScatter.wdl
```diff
@@ -13,7 +13,9 @@ workflow ExpansionHunterScatter {
         File reference_fasta
         File? reference_fasta_index
         File variant_catalog
```
mwalker174, line 15:
```suggestion
        File variant_catalog_json
```
To make it clear this should be a json and not a bed file

wdl/ExpansionHunterScatter.wdl
```diff
@@ -22,6 +24,16 @@ workflow ExpansionHunterScatter {
         sample_ids: "One ID per sample, in the same order as the files in bams_or_crams. These IDs must match the ID given in the second column (`Individual ID` column) of the given PED file. These IDs will also be used as an output prefix."
     }
 
+    String variant_catalog_batch_size_ = select_first([variant_catalog_batch_size, 1000])
+
+    call SplitVariantCatalog as svc {
+        input:
+            variant_catalog = variant_catalog,
+            batch_size = variant_catalog_batch_size_,
+            output_prefix = basename(variant_catalog),
```
mwalker174, line 33:
```suggestion
            output_prefix = basename(variant_catalog, ".json"),
```
Strips off the extension for a cleaner prefix

wdl/ExpansionHunterScatter.wdl
VJalili: Agree, that will make it more user-friendly, thanks. I updated accordingly.

wdl/ExpansionHunter.wdl
mwalker174: Yeah awk is better here

mwalker174 (approved):

PR #281, opened 2022-01-10
ExpansionHunter extract sex from a ped file.

VJalili:
EH processes male and female samples differently; this PR adds a functionality where sample sex is extracted from a provided [`PED`](https://gatk.broadinstitute.org/hc/en-us/articles/360035531972-PED-Pedigree-format) file.

### DIFF
--- a/wdl/ExpansionHunter.wdl
+++ b/wdl/ExpansionHunter.wdl
@@ -59,8 +59,8 @@ workflow ExpansionHunter {
             reference_fasta = reference_fasta,
             reference_fasta_index = reference_fasta_index_,
             variant_catalog = variant_catalog,
-            individual_id = individual_id,
-            ped_filename = ped_filename,
+            sample_id = sample_id,
+            ped_file = ped_file,
             output_prefix = output_prefix_,
             expansion_hunter_docker = expansion_hunter_docker,
             runtime_attr_override = runtime_attr,
@@ -81,7 +81,7 @@ task RunExpansionHunter {
         File reference_fasta
         File reference_fasta_index
         File variant_catalog
-        String? individual_id
+        String? sample_id
         File? ped_file
         String output_prefix
         String expansion_hunter_docker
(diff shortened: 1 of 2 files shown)


wdl/ExpansionHunter.wdl
```diff
@@ -87,13 +93,22 @@ task RunExpansionHunter {
     command <<<
         set -euxo pipefail
 
+        sex=""
+        if [ ! -z "~{ped_filename}" -a "~{ped_filename}" != " " ]; then
+            if [ -e "~{ped_filename}" ]; then
```
mwalker174, line 98: Easier to just write `if ~{defined(ped_filename)}; then`

wdl/ExpansionHunter.wdl
```diff
@@ -87,13 +93,22 @@ task RunExpansionHunter {
...
+                sex=$(awk -F '\t' '{if ($2 == "~{individual_id}") {if ($5 == "1") {print "--sex male"; exit 0} else if ($5 == "2") {print "--sex female"; exit 0} else {exit 1}}}' < ~{ped_filename} )
```
mwalker174, line 99: I would actually try to emit a helpful error message if the sex is undefined. I think it would be best to define `sex="NA"` if the ped file is provided. Then use `print "NA"; exit 0` instead of `exit 1` in the last `if` clause. Then check the value of `sex` below - if it's `NA` then warn that either the sample id was not found or the sex is undefined and fail.

wdl/ExpansionHunter.wdl
```diff
@@ -20,6 +20,8 @@ workflow ExpansionHunter {
         File reference_fasta
         File? reference_fasta_index
         File variant_catalog
+        String? individual_id
+        File? ped_filename
```
mwalker174, line 24:
```suggestion
        File? ped_file
```

wdl/ExpansionHunterScatter.wdl
```diff
@@ -9,6 +9,7 @@ workflow ExpansionHunterScatter {
         Array[File] bams_or_crams
         Array[File]? bams_or_crams_indexes
         Array[String]? sample_ids
+        File? ped_filename
```
mwalker174, line 12:
```suggestion
        File? ped_file
```

wdl/ExpansionHunter.wdl
```diff
@@ -72,6 +76,8 @@ task RunExpansionHunter {
         File reference_fasta
         File reference_fasta_index
         File variant_catalog
+        String? individual_id
+        File? ped_filename
```
mwalker174, line 80:
```suggestion
        File? ped_file
```

wdl/ExpansionHunter.wdl
mwalker174: Also need to check if the sample id is defined. If not, throw an error.

wdl/ExpansionHunter.wdl
```diff
@@ -20,6 +20,8 @@ workflow ExpansionHunter {
         File reference_fasta
         File? reference_fasta_index
         File variant_catalog
+        String? individual_id
```
mwalker174, line 23:
```suggestion
        String? sample_id
```
Just so it's crystal clear to the user, since `individual` could be construed as a slightly different entity under a legacy Terra data model.

Also, can you add a comment that the sample id and ped file must be provided together?

wdl/ExpansionHunterScatter.wdl
```diff
@@ -37,13 +38,21 @@ workflow ExpansionHunterScatter {
                 else
                     basename(bam_or_cram_, ".cram")
 
+        String individual_id =
```
mwalker174, line 41:
```suggestion
        String sample_id =
```

mwalker174 (changes requested):

wdl/ExpansionHunter.wdl
VJalili: Please check if the updated condition would be equally good.

mwalker174 (approved):
Looks good! One minor suggestion.

PR #283, opened 2022-01-14
External AF SVID INFO key is String type

epiercehoffman:
### Updates
Tweaks to AnnotateVcf workflow:
* SVID INFO key from AnnotateExternalAF should be a String type, not Float. This was causing issues with Hail
  * There are two ModifyVcf tasks, one in AnnotateExternalAF and one in AnnotateExternalAFPerContig. It appears that only the one in the per-contig WDL is used, but I updated both because they aren't identical so I was wary of deleting the other 
* Output name of annotated VCF was `{prefix}.annotated.vcf.vcf.gz`; removed extra `.vcf` 
* Spacing

### Testing
* Tested AnnotateVcf and verified that the output VCF name had one `.vcf` and the header reflected that the SVID INFO key is of type String

### DIFF
--- a/wdl/AnnotateExternalAF.wdl
+++ b/wdl/AnnotateExternalAF.wdl
@@ -32,13 +32,13 @@ workflow AnnotateExternalAF {
         RuntimeAttr? runtime_attr_select_matched_svs
 
     }
-    call SplitBed as split_ref_bed{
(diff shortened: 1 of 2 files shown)


mwalker174 (approved):
Thank you, looks fine to me!

PR #289, opened 2022-01-31
fix missing index extension

henriqueribeiro:
`disc_files_idx` was being created without the index extension, making the pipeline to fail.


mwalker174 (approved):
Thank you for catching this. I guess this would break only when `localization_optional` feature is not supported, which is why we haven't seen this error.

henriqueribeiro:
exactly @mwalker174. You are welcome

PR #291, opened 2022-02-04
Adding -k -f parameters to gunzip to avoid failures of GermlineCNV processes

spatel-gfb:
Adding -k -f flag to gunzip command as on local the file overwriting fails. 
So -k will keep the original gz files and -f will force the unzipping of files and will overwrite the files on local so the process won't fail.

Error :
gzip: /gfb-v2-dev-sv-fsx-results-us-east-2/cromwell-execution/GATKSVPipelineBatch/2b6de38e-b3eb-47b8-8f9c-b935dc1c741f/call-GATKSVPipelinePhase1/GATKSVPipelinePhase1/01429d16-a31c-48c9-b374-0e856d5e9c81/call-GatherBatchEvidence/GatherBatchEvidence/4eb237c7-ddc3-4f2b-b1ed-8272eb49715b/call-CondenseReadCounts/shard-1/condensed_counts.HG00129.tsv already exists;	not overwritten

### DIFF
--- a/wdl/GermlineCNVCase.wdl
+++ b/wdl/GermlineCNVCase.wdl
@@ -245,7 +245,7 @@ task DetermineGermlineContigPloidyCaseMode {
         tar xzf ~{contig_ploidy_model_tar} -C input-contig-ploidy-model
 
         read_count_files_list=~{write_lines(read_count_files)}
-        grep gz$ $read_count_files_list | xargs -l1 -P0 gunzip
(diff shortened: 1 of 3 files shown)


mwalker174:
Hi @spatel-gfb, thanks for contributing. Would it be possible for you to use the "copy" localization/caching strategy option described [here](https://cromwell.readthedocs.io/en/stable/Configuring/#local-filesystem-options)?

Issue #18, opened 2020-07-01
BAF from BAM


mwalker174:
BAF generation could be improved by using GATK's `CollectAllelicCounts` on sample BAMs rather than gVCFs/VCFs, which is not only more costly but also operationally more complicated (eg see #8). I propose collecting only at gnomAD SNP sites with >0.1% AF.

mwalker174:
Following up on this with more detail.

We currently generate BAF in `GenerateBatchMetrics` (formerly `Module00c`) from GATK SNP calls using one of two workflows. If the user provides gVCFs then `BAFFromGVCFs` is used, and if they provide a joint-called VCF (or set of sharded VCFs) then `BAFFromShardedVCF` is used. This has been a pain point for a number of reasons:

1. Obviously, the SNP calls must be generated first. This introduces delays for projects where SNV and SV generation are planned in parallel, and often creates more obtaining SNV calls in a format digestible by the pipeline.
2. gVCFs are extremely large and storage is expensive.
3. Joint VCFs can also be extremely large and again it creates work if it needs to be sharded.
4. For larger cohorts especially, SNV calls may only be available for a subset of samples. The pipeline does accommodate this scenario, but it is somewhat confusing and error-prone to set up BAF generation, particularly in Terra (mistakes have happened here in the past).
5. BAF is generated per-batch, which makes it difficult to share BAF records for individual samples across overlapping cohorts.
6. `GenerateBatchMetrics` is already a large workflow, and it would be good to move work to other modules where possible.

For these reasons, it would be a huge improvement to generate BAF directly from the CRAMs. Since BAF is only used for random forest training in `FilterBatch`, we don't expect slight changes in methodology to affect results substantially. The new strategy will be:

1. Move BAF generation to `PESRCollection` and rename the workflow `CollectSVEvidence`. This workflow should use a sites list (in VCF format) of common SNPs from gnomAD to collect allelic counts (`LocusDepth` features) with GATK's `CollectSVEvidence` tool. BAF generation should be optional, however, as it is not needed for the single-sample mode. Delete the old BAF workflows from `GatherBatchEvidence`.
[...]

PR #293, opened 2022-02-09
Migrate to 1KGP reference panel for testing

mwalker174:
Addresses issue #10
- Reorganized input json directory structure, with templates at `/inputs/templates`, value jsons at `/inputs/values`, and default builds in `/inputs/build`. Build paths have been cleaned up to be more intuitive.
- Removes `ref_panel_v1x` from the set of test data
- Fresh run of the 1000 Genomes reference panel, `ref_panel_1kg`. This was generated using a new workflow by running `GATKSVPipelineBatch` using the Cromwell config option `--final-workflow-outputs-dir` to copy outputs to a permanent location, and then feeding the metadata to a new script at `scripts/inputs/create_test_batch.py` to generate input values at `inputs/values/ref_panel_1kg.json`.
- Updated readme to reflect changes, including a new section: "Generating a reference panel"
- Renamed `wdl_test` dir to `carrot`


README.md
```diff
@@ -174,38 +177,48 @@ For larger cohorts, samples should be split up into batches of about 100-500 sam
...
+    inputs/templates/test/GATKSVPipelineSingleSample \
+    inputs/build/NA19240/test_my_ref_panel \
+    -a '{ "single_sample" : "test_single_sample_NA19240", "ref_panel" : "my_ref_panel" }'
+> # Build files for a Terra workspace
+> python scripts/inputs/build_inputs.py \
+    inputs/values \
+    inputs/templates/terra_workspaces/single_sample \
+    inputs/build/NA12878/terra_my_ref_panel \
+    -a '{ "single_sample" : "test_single_sample_NA12878", "ref_panel" : "my_ref_panel" }'
+```
```
epiercehoffman, line 217: Could also note that build_default_inputs.py can be used to build all at once

epiercehoffman (changes requested):
This is a long-awaited move and a lot of work! I commented with some small things I noticed, but overall I think it looks pretty solid. I tested building the JSONs and validating them and it went pretty smoothly and the organization makes sense.

A few bigger-picture comments:

- I think it would be nice to have a smaller test batch for when we are testing changes to GatherSampleEvidence - I usually use `test_small` for this, but here you'd have to run all 156 samples or manually edit a JSON
- I think it would be beneficial to be able to generate test JSONs for a 2-batch cohort matching the contents of the Terra workspace, both to test multi-batch steps and to match the Terra workspace more closely. I'm not totally sure how best to implement that - it could work as a separate test_batch with its own values JSON, but there will be duplicated information across the 1-batch and 2-batch JSONs in that case
- The `inputs/` directory gets nested so deeply which is a little cumbersome to work with. I'm not certain what to change though
- Nesting the Terra cohort mode JSONs under `ref_panel_1kg` doesn't fully make sense since the Terra workspace uses 2 batches so it is a superset of the ref panel samples. If a 2-batch test set were to be added, it would fit better under there

wdl/ClusterBatch.wdl
```diff
@@ -170,10 +170,15 @@ workflow ClusterBatch {
 
   output {
     File depth_vcf = ClusterDepth.clustered_vcf
```
epiercehoffman, line 172: Missed this at first - this is one of the workflows that has a separate output JSON in Terra because the output names are the same as some of the other workflows, and that doesn't play nicely with the data tables. So when the outputs are changed like this, we either need to update the outputs JSON to add the new outputs, or (preferred) change the workflow output names to be unique - ie. merged_depth_vcf or clustered_depth_vcf or similar

epiercehoffman: (The other workflows with Terra output JSONs did not have their outputs change in this PR)

epiercehoffman: Actually this might be outdated because I can't find another workflow that has these output names. Regardless, because it was set up this way, either the Terra JSONs need to be updated or the WDL does

inputs/templates/test/TrainGCNV/TrainGCNV.json.tmpl
```diff
@@ -57,7 +57,16 @@
   "TrainGCNV.gcnv_sample_psi_scale" : 0.000001,
   "TrainGCNV.ploidy_sample_psi_scale" : 0.001,
 
-  "TrainGCNV.cohort": "ref_panel_v1b",
+  "TrainGCNV.runtime_attr_bundle": {
```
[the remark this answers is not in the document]
mwalker174: This actually isn't even a task anymore. Cleaned that up.

inputs/values/ref_panel_1kg.json
```diff
@@ -0,0 +1,3032 @@
+{
+    "PE_files": [
```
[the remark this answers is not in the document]
mwalker174: Ah this is because they are capitalized. I could go through and change the names to lower-case but this will be a bit of trouble. We can try to address this in #16

inputs/templates/test/RegenotypeCNVs/RegenotypeCNVs.json.tmpl
```diff
@@ -6,9 +6,9 @@
   "RegenotypeCNVs.n_RdTest_bins": "100000",
   "RegenotypeCNVs.n_per_split": "5000",
 
-  "RegenotypeCNVs.cohort": {{ test_batch.batch_name | tojson }},
+  "RegenotypeCNVs.cohort": {{ test_batch.name | tojson }},
   "RegenotypeCNVs.contig_list": {{ reference_resources.primary_contigs_list | tojson }},
-  "RegenotypeCNVs.regeno_coverage_medians": {{ test_batch.regeno_coverage_medians | tojson }},
+  "RegenotypeCNVs.regeno_coverage_medians": [{{ test_batch.medianfile | tojson }}],
```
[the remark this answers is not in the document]
mwalker174: Great catch

wdl/GATKSVPipelineBatch.wdl
```diff
@@ -339,38 +350,109 @@ workflow GATKSVPipelineBatch {
...
+
+    File filtered_depth_vcf = select_first([GATKSVPipelinePhase1.filtered_depth_vcf])
+    File filtered_pesr_vcf = select_first([GATKSVPipelinePhase1.filtered_pesr_vcf])
+    File cohort_pesr_vcf = select_first([GATKSVPipelinePhase1.filtered_pesr_vcf])
+    File cohort_depth_vcf = select_first([GATKSVPipelinePhase1.filtered_depth_vcf])
+    File? sites_filtered_manta_vcf = GATKSVPipelinePhase1.sites_filtered_manta_vcf
+    File? sites_filtered_wham_vcf = GATKSVPipelinePhase1.sites_filtered_wham_vcf
+    File? sites_filtered_melt_vcf = GATKSVPipelinePhase1.sites_filtered_melt_vcf
+    File? sites_filtered_depth_vcf = GATKSVPipelinePhase1.sites_filtered_depth_vcf
+    # TODO : can't get FilterBatchSamples inputs since PlotSVCountsPerSample uses a select_all
```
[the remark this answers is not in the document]
mwalker174: Good point - I've removed the count inputs

inputs/templates/test/GATKSVPipelineBatch/GATKSVPipelineBatch.json.tmpl
```diff
@@ -1,47 +1,50 @@
 {
```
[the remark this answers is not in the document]
mwalker174: Done. Also fixed capitalization with mosaic/vapor/module09.

inputs/templates/test/GATKSVPipelineBatch/GATKSVPipelineBatch.FromSampleEvidence.json.tmpl
```diff
@@ -137,7 +143,7 @@
   "GATKSVPipelineBatch.MakeCohortVcf.random_seed": 0,
   "GATKSVPipelineBatch.MakeCohortVcf.max_shard_size_resolve": 500,
 
-  "GATKSVPipelineBatch.BatchQC.qc_definitions" : "gs://gatk-sv-resources-public/test/batch/batch_sv.test_large.qc_definitions.tsv",
+  "GATKSVPipelineBatch.qc_definitions" : "gs://gatk-sv-resources-public/test/batch/batch_sv.test_large.qc_definitions.tsv",
```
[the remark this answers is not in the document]
mwalker174: Actually we should be updating these qc definitions files for each batch, since they include extensive properties such as variant counts. I've created a new script `scripts/inputs/calibrate_qc_metrics.py` that generates a set of qc definitions from a baseline metrics file, and updated the templates and ref_panel_1kg inputs with updated versions.

inputs/templates/test/GATKSVPipelineBatch/GATKSVPipelineBatch.json.tmpl
```diff
@@ -1,47 +1,50 @@
...
-  "GATKSVPipelineBatch.allosome_file" : {{ reference_resources.allosome_file | tojson }},
-  "GATKSVPipelineBatch.autosome_file" : {{ reference_resources.autosome_file | tojson }},
-  "GATKSVPipelineBatch.primary_contigs_list" : {{ reference_resources.primary_contigs_list | tojson }},
-  "GATKSVPipelineBatch.primary_contigs_fai" : {{ reference_resources.primary_contigs_fai | tojson }},
-  "GATKSVPipelineBatch.genome_file" : {{ reference_resources.genome_file | tojson }},
-  
-  "GATKSVPipelineBatch.reference_dict" : {{ reference_resources.reference_dict | tojson }},
-  "GATKSVPipelineBatch.reference_fasta" : {{ reference_resources.reference_fasta | tojson }},
-  "GATKSVPipelineBatch.reference_index" : {{ reference_resources.reference_index | tojson }},
+  "GATKSVPipelineBatch.GATKSVPipelinePhase1.gvcf_gcs_project_for_requester_pays" : "broad-dsde-methods",
```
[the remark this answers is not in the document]
mwalker174: This is a good point, as I've seen at least one instance of it getting used. I've added a new parameters group `cloud_env` to the inputs script with example json at `inputs/values/google_cloud.json'. I updated the readme `Building inputs` section with a mention of this, and added a step under `Execution` to show how to define the project id.

scripts/inputs/build_inputs.py
```diff
@@ -12,23 +12,23 @@
 # potentially tsv files for Terra import) based on the input templates in the repository and the
 # set of specified input values.
 #
```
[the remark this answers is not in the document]
mwalker174: Added logging with default WARN

README.md
mwalker174: Well that has ref_panel_1kg hard-coded, so you couldn't use it without modifying it?

wdl/ClusterBatch.wdl
mwalker174: Changed the WDL output names

mwalker174:
Thanks for the thorough review, @epiercehoffman !

> * I think it would be nice to have a smaller test batch for when we are testing changes to GatherSampleEvidence - I usually use `test_small` for this, but here you'd have to run all 156 samples or manually edit a JSON

I agree it would be nice to have a smaller cohort. IMO, it's a bit cluttering and adds maintenance to have two separate cohorts, especially if one can only be run partway through the pipeline. I think eventually we'll want to consider running on a couple of chromosomes (say 21 and Y), but that will require some updates to the wdls and likely some bug fixes in places.

> * I think it would be beneficial to be able to generate test JSONs for a 2-batch cohort matching the contents of the Terra workspace, both to test multi-batch steps and to match the Terra workspace more closely. I'm not totally sure how best to implement that - it could work as a separate test_batch with its own values JSON, but there will be duplicated information across the 1-batch and 2-batch JSONs in that case

Yes we should have something for the 2-batch testing and Terra workspace. I think that would have to be defined as a separate test batch, even if some of the data are the same. Might be good to open an issue for this.

> * The `inputs/` directory gets nested so deeply which is a little cumbersome to work with. I'm not certain what to change though

I think it's an improvement to have everything in an `inputs` directory - the old organization was a bit confusing IMO. I don't know what to do about nesting but I feel like everything is organized in a more logical way now.

> * Nesting the Terra cohort mode JSONs under `ref_panel_1kg` doesn't fully make sense since the Terra workspace uses 2 batches so it is a superset of the ref panel samples. If a 2-batch test set were to be added, it would fit better under there

Yes when we add a 2-batch test, we can also generate a workspace for that. Not sure if we have a 1-batch workspace currently? But this way one can create the workspace for single-batch at least without needing access to our workspaces.

inputs/templates/test/GatherBatchEvidence/GatherBatchEvidence.baf_from_vcf.json.tmpl
```diff
@@ -50,7 +50,7 @@
   "GatherBatchEvidence.gcnv_log_emission_sampling_median_rel_error": "0.001",
   "GatherBatchEvidence.gcnv_qs_cutoff": "30",
```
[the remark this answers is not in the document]
mwalker174: Somehow I missed this comment - generating the VCFs now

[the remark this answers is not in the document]
mwalker174: Actually I am going to punt on this since the RP bucket bug is currently preventing me from running GatherSampleEvidence. This workflow will not be needed much longer, once #18 is implemented.

README.md
```diff
@@ -116,7 +121,8 @@ We recommend running the pipeline on a dedicated [Cromwell](https://github.com/b
 > cp $GATK_SV_ROOT/wdl/*.wdl .
 > zip dep.zip *.wdl
 > cd ..
-> bash scripts/inputs/build_default_inputs.sh -d $GATK_SV_ROOT
+> echo '{ "google_project_id": "my-google-project-id" }' > inputs/values/google_cloud.my_project.json
```
epiercehoffman, line 124: Would it be a little easier to just provide the project directly to the build script? ie. `-c "broad-dsde-methods"`. I tested this process and it seems to work fine, but I wonder if it would not be easier to just directly supply the project rather than creating a file

inputs/templates/terra_workspaces/cohort_mode/workflow_configurations/GatherBatchEvidence.json.tmpl
```diff
@@ -61,5 +61,6 @@
   "GatherBatchEvidence.samples": "${this.samples.sample_id}",
   "GatherBatchEvidence.melt_vcfs": "${this.samples.melt_vcf}",
   "GatherBatchEvidence.wham_vcfs": "${this.samples.wham_vcf}",
-  "GatherBatchEvidence.gvcfs": "${this.samples.gvcf}"
+  "GatherBatchEvidence.gvcfs": "${this.samples.gvcf}",
+  "GatherBatchEvidence.gvcf_gcs_project_for_requester_pays" : "${workspace.google_project_id}"
```
epiercehoffman, line 65: I like this idea, but this does mean that "broad-firecloud-dsde-methods" will show up in the Workspace Data table that will get cloned along with the workflows - previously I just entered it manually to avoid this. Should also update dashboard to reflect that users should enter their own project in Workspace Data (and direct them to where to find the project ID)

wdl/ClusterBatch.wdl
```diff
@@ -169,16 +169,16 @@ workflow ClusterBatch {
...
-    File depth_vcf_index = ClusterDepth.clustered_vcf_index
-    File? manta_vcf = ClusterPESR_manta.clustered_vcf
-    File? manta_vcf_index = ClusterPESR_manta.clustered_vcf_index
-    File? delly_vcf = ClusterPESR_delly.clustered_vcf
-    File? delly_vcf_index = ClusterPESR_delly.clustered_vcf_index
-    File? wham_vcf = ClusterPESR_wham.clustered_vcf
-    File? wham_vcf_index = ClusterPESR_wham.clustered_vcf_index
-    File? melt_vcf = ClusterPESR_melt.clustered_vcf
-    File? melt_vcf_index = ClusterPESR_melt.clustered_vcf_index
+    File clustered_depth_vcf = ClusterDepth.clustered_vcf
```
epiercehoffman, line 172: Thanks for doing this. Can also now delete the Terra output JSON for ClusterBatch

epiercehoffman (approved):
Thanks for making these changes! I left a few small comments but they are more subjective / don't necessarily need to block merging. Everything appears to be in working order - building default inputs (with less output to wade through!), FilterBatchSamples JSON gets built, and the workspace data tables look like they're formatted properly. 

Regarding the question of a 2-batch test set & the organization in Terra: there are currently single-batch workflow configurations that appear in the 2-batch workspace, so that users can clone one workspace and use whichever workflows apply to them. I do have a test workspace where I only processed one batch (so the workflows are the same across both workspaces, but the data is different), but I don't really maintain it. The necessary files will get generated to be able to process one or two batches in Terra regardless, and it is more a question of organization in the repo. I think the more _functionally_ important part of adding a 2-batch test set will be the ability to test cohort-level workflows in cromwell on 2 batches.

Probably a good idea to open a separate issue to create a 2-batch test set. Maybe also an issue to track that we don't currently have SNP VCFs for the ref panel, which can be resolved by either adding SNP VCFs or implementing #18.

README.md
mwalker174: It would be a little simpler to do that. I think we should keep it as a file definition to remain consistent with how all inputs are defined. It also makes it clear/intuitive that the google project id is a workflow input and as well as that it is not somehow being used for the build process.

inputs/templates/terra_workspaces/cohort_mode/workflow_configurations/GatherBatchEvidence.json.tmpl
mwalker174: Actually now that you point this out, this workspace variable is populated using `cloud_env.google_project_id`, which is always going to be different from the project used for a Cromwell run. I've added another variable `cloud_env.terra_billing_project_id` for workspaces.

PR #294, opened 2022-02-10
Add TrainGCNV input specifying subset list of samples for training

epiercehoffman:
### Updates
Add `sample_ids_training_subset` input to `TrainGCNV` to allow Terra users to specify a subsetted list of sample IDs on which to train the gCNV model, while keeping the same batching they will use for GatherBatchEvidence. Intended as an alternative to the random downsampling option if the user wants a greater level of control over the samples chosen.

### Testing
* Tested `TrainGCNV` on `test_small` test set and specified four sample IDs for training, out of order. Verified that the four were sorted according to the order of the `samples` input (and other sample-level inputs) and the rest of the tasks were performed on the subset of samples.

### DIFF
--- a/wdl/TrainGCNV.wdl
+++ b/wdl/TrainGCNV.wdl
@@ -18,8 +18,12 @@ workflow TrainGCNV {
     File reference_index    # Index (.fai), must be in same dir as fasta
     File reference_dict     # Dictionary (.dict), must be in same dir as fasta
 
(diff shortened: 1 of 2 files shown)


mwalker174 (changes requested):

mwalker174 (approved):
Looks good. Thanks!

PR #295, opened 2022-02-16
Single sample pipeline fixes for MakeCohortVcf

mwalker174:
Addresses two bugs introduced in #288 that occur in the single sample pipeline:
- Deleted function in merge_pesr_depth.py that "cleaned" depth-only call records by basing them on PESR records. This was causing an error when there were no PESR records, and turns out to be unnecessary anymore.
- Fixed a `grep` error caused by empty files in IntegrateResolvedVcfs.

Tested `MakeCohortVcf` and `GATKSVSingleSamplePipeline` for correctness. Metrics for the latter (blue) compared to the current featured workspace (orange): [mw-merge-pesr-depth-empty.single_sample_metrics_comparison.pdf](https://github.com/broadinstitute/gatk-sv/files/8082701/mw-merge-pesr-depth-empty.single_sample_metrics_comparison.pdf)


epiercehoffman (approved):
This looks fine to me! 

I don't totally understand why the depth records had to be "cleaned" before or what changed so it's not necessary, but taking that as a given, these changes make sense to me. And the metrics look pretty similar - I'm guessing some of the small differences come from other changes that have been implemented since the featured workspace was updated.

PR #297, opened 2022-02-23
Add minGQ_NCR INFO field

mwalker174:
Adds a new INFO field `minGQ_NCR` to all non-multiallelic sites during minGQ filtering giving the fraction of no-call genotypes, in addition to the `HIGH_X_NOCALL_RATE` filter status.


epiercehoffman (approved):
Looks fine to me

PR #299, opened 2022-02-24
Restore --gcr-project argument with deprecation warning

TedBrookings:


scripts/docker/build_docker.py
```diff
@@ -697,6 +700,10 @@ def __parse_arguments(args_list: List[str]) -> argparse.Namespace:
             raise ValueError(
                 "Current directory has uncommitted changes or untracked files. Cautiously refusing to proceed."
             )
+
+    if parsed_args.gcr_project is not None:
+        print(colored("--gcr-project is deprecated, use --docker-repo instead.", "red"))
+        parsed_args.docker_repo = f"us.gcr.io/{parsed_args.gcr_project}"
```
mwalker174, line 706: Can you throw an error if both `--gcr-project` and `--docker-repo` are specified? Or make the arguments [mutually exclusive](https://docs.python.org/3/library/argparse.html#mutual-exclusion)

mwalker174 (approved):
Looks good, just one suggestion to avoid possible confusion

scripts/docker/build_docker.py
TedBrookings: I pushed a solution. I opted to throw an error if they are both specified, because it keeps the change to one block of code that's easier to maintain/remove.

PR #300, opened 2022-03-02
Optimize trio ID filtering in optimize_minGQ_ROC_v2.R

mwalker174:
Yields over 100x speedup

### DIFF
--- a/src/sv-pipeline/scripts/downstream_analysis_and_filtering/optimize_minGQ_ROC_v2.R
+++ b/src/sv-pipeline/scripts/downstream_analysis_and_filtering/optimize_minGQ_ROC_v2.R
@@ -27,11 +27,11 @@ readTrioDat <- function(dn_dat){
   return(dat.out)
 }
 #Get counts of all inherited & de novo variants given min GQ
-filterGQStats <- function(dat,trio,minGQ=0){
+filterGQStats <- function(dat,minGQ=0){
   #Get variant counts
-  max.inh <- nrow(dat[which(dat$fam==trio & dat$label=="inh"),])
-  ret.inh <- nrow(dat[which(dat$fam==trio & dat$pro_GQ>minGQ & dat$label=="inh"),])
-  ret.dn <- nrow(dat[which(dat$fam==trio & dat$label=="dn" &
+  max.inh <- nrow(dat[which(dat$label=="inh"),])
+  ret.inh <- nrow(dat[which(dat$pro_GQ>minGQ & dat$label=="inh"),])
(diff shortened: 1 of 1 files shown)


src/sv-pipeline/scripts/downstream_analysis_and_filtering/optimize_minGQ_ROC_v2.R
```diff
@@ -27,11 +27,11 @@ readTrioDat <- function(dn_dat){
   return(dat.out)
 }
 #Get counts of all inherited & de novo variants given min GQ
-filterGQStats <- function(dat,trio,minGQ=0){
+filterGQStats <- function(dat,minGQ=0){
   #Get variant counts
-  max.inh <- nrow(dat[which(dat$fam==trio & dat$label=="inh"),])
-  ret.inh <- nrow(dat[which(dat$fam==trio & dat$pro_GQ>minGQ & dat$label=="inh"),])
-  ret.dn <- nrow(dat[which(dat$fam==trio & dat$label=="dn" &
+  max.inh <- nrow(dat[which(dat$label=="inh"),])
```
cwhelan, line 32:
```suggestion
  max.inh <- sum(dat$label == "inh")
  ret.inh <- sum(dat$pro_GQ>minGQ & dat$label=="inh")
  ret.dn  <- sum(dat$label=="dn" &
                             dat$pro_GQ>minGQ & dat$fa_GQ>minGQ & dat$mo_GQ>minGQ))
```

It would probably be even faster to sum the boolean values returned by these queries (in R `TRUE` = 1 and `FALSE` = 0 in `sum`) and not form new intermediate data tables. Also simpler to read.

cwhelan (approved):
Looks like a sensible change -- amazing that the speedup is 100x, I would have guessed more like 3x. If you want to refine further I have a suggestion that simplifies the code and might provide additional speed (but maybe if the really slow part was just subsetting to the trio it won't show up any different in timing).

src/sv-pipeline/scripts/downstream_analysis_and_filtering/optimize_minGQ_ROC_v2.R
mwalker174: Thanks R is definitely not my strong suit 😅 I would incorporate your suggestion except we've lost the test run to TTL and xuefang is on vacation, so I don't have a way to test right now. The 100x speedup (might've been even higher) was probably more dramatic since it was on gnomAD, so subsetting to the trio first saves iterating on the 140K rows 2 extra times.

PR #301, opened 2022-03-04
Revert gatk dockers to 4.1.8.1

mwalker174:


epiercehoffman (approved):
Makes sense, since the condense counts docker is separate. Did you also want to update the dockers that should have gotten updated by the docker deploy git action?

PR #302, opened 2022-03-04
Fix default dockers.json path in docker scripts

mwalker174:
Updates default location of `dockers.json` (changed in #293) in the docker build and update scripts.

### DIFF
--- a/scripts/docker/build_docker.py
+++ b/scripts/docker/build_docker.py
@@ -61,6 +61,7 @@ Instead it prints the changes that it would make if you ran without
 class Paths:
     this_script_folder = os.path.dirname(os.path.abspath(__file__))
     gatk_sv_path = os.path.dirname(os.path.dirname(this_script_folder))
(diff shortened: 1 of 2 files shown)


TedBrookings:
@mwalker174 This PR does what it says on the tin, and it's so simple I nearly approved it. However, I'm curious, wouldn't it break build_default_inputs.sh without changes to the expected location of dockers.json. And should the existing dockers.json move too as part of the PR?

mwalker174:
@TedBrookings I realize now my description is a little confusing - `dockers.json` was moved to `/inputs/values` in #293 already (so `build_docker.py` and `update_docker_json.py` are broken on master). Hard-coded paths in `build_default_inputs.sh` were updated with #293, so it works correctly already.

TedBrookings (approved):
I have a very mild preference to move dockers_json_path to Paths.dockers_json (not an issue that originates with this PR, just one that brought it to my attention). But that's totally trivial. Merge it as is if that's the easiest thing.

PR #306, opened 2022-03-10
Adjust VM disk size w.r.t the input files size.

VJalili:
This PR sets the disk size of the VM of the task that combines the results of the scattered ExpansionHunter runs to the sum of input files size multiplied by two, in order to fit the input files and their merged versions.


mwalker174 (approved):
Looks good to me - always a good idea to do this kind of thing when inputs/outputs have the potential to be large (~GB).

VJalili:
Thanks for the review @mwalker174!

PR #308, opened 2022-03-11
Determine the docker images to be built using build_docker.py script

VJalili:
Currently, the `workflows/sv_pipeline_docker.yml` uses two git commits to determine the files changed between the commits, and uses the list of changed files to determine which docker images should be rebuilt. This functionality is now available in `build_docker.py`; hence, in order to simplify the workflow, this PR implements changes to leverage `build_docker.py` to decide which docker images should be rebuilt.

### DIFF
--- a/.github/workflows/sv_pipeline_docker.yml
+++ b/.github/workflows/sv_pipeline_docker.yml
@@ -25,8 +25,9 @@ jobs:
     env:
       GITHUB_CONTEXT: ${{ toJson(github) }}
     outputs:
-      target_images: ${{ steps.target_images.outputs.TARGETS }}
       images_tag: ${{ steps.image_tag.outputs.IMAGE_TAG }}
+      base_sha: ${{ steps.commit_sha.outputs.BASE_SHA }}
+      head_sha: ${{ steps.commit_sha.outputs.HEAD_SHA }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


TedBrookings:
I think there's no reason to compute and pass the head commit. It will always be the current git repo status, which is the default. I'd suggest removing it to further reduce complexity.

.github/workflows/sv_pipeline_docker.yml
```diff
@@ -205,7 +132,7 @@ jobs:
       - name: Run build_docker.py
         run: |
           cd ./scripts/docker/
-          python build_docker.py --targets ${{ needs.build_args_job.outputs.target_images }} --image-tag ${{ needs.build_args_job.outputs.images_tag }}
+          python build_docker.py --base-git-commit ${{ needs.build_args_job.outputs.base_sha }} --current-git-commit ${{ needs.build_args_job.outputs.head_sha }} --image-tag ${{ needs.build_args_job.outputs.images_tag }}
```
TedBrookings, line 135: There's no need to compute and pass the --current-git-commit. In this context it will always be the current git repo status, which is the default value.

TedBrookings, line 135: There's no need to compute and pass --current-git-commit. In this context, the correct value is always the current git repo status, which is the default.

TedBrookings, line 135: There's also no need to compute the --image-tag. build_docker.py has a default which will work.

TedBrookings (changes requested):
I still don't see the need to keep the docker-repo "secret" since it's publicly discernible and not exploitable, although I won't require changing that if you insist, since it will work this way.

I made some suggestions that should enable further simplification of the script.

.github/workflows/sv_pipeline_docker.yml
VJalili: Updated accordingly.

VJalili:
Thanks for the review @TedBrookings. I agree with you on the `docker-repo` environment variable. I kept this PR focused on determining the target docker images, and we can work on defining `docker-repo` as an env variable in a separate PR.

.github/workflows/sv_pipeline_docker.yml
TedBrookings: To me, the situation is that the calculation of the head commit for a non-PR is complicated and involves piping stuff through jq, sed, etc. As you say, the final commit, regardless of being a surrogate must have the same changes to project files as the head commit that was pushed, otherwise the merge couldn't work. So I just see another opportunity to strip complexity out of this script. And I'm waiting for the day when github changes its API slightly, at which point we'd only have one thing to figure out how to compute (the base commit). Since the PR as written works, I'm not going to insist.

VJalili: Thank you!

PR #309, opened 2022-03-11
AWS execution related code changes

spatel-gfb:
Through this PR , we are updating the wdl's to handle shared_filesystem/loca/FSx for luster and also the steps needed for running it on AWS.

We created this PR to get the inputs and clarify the isssues/concerns arising out of the changes. The shared_filesystem was discussed with @epiercehoffman  for handling different local file systems. 

Please don't merge this yet as we are running the code with these changes on AWS. Once succesfully completed, we will request you to merge it.


epiercehoffman (changes requested):
Thanks for working with us as we figure out the best way to accommodate multiple execution platforms. I have a few changes to request:

* After further conversation with our team, we have decided that it is preferable to avoid introducing multiple code paths that could become difficult to maintain, and we are okay with increasing disk size for VMs to accommodate changes that avoid modifying input files for shared file systems. So, we will approve the changes you have suggested without the `shared_filesystem` flag and conditionals, provided that they run successfully on GCP when we test them. I realize this is a change from our previous request, so again, thank you for bearing with us as we figure out the best way to handle this.
* Since our team does not have the infrastructure to test functionality on AWS or maintain AWS-specific code, we are not comfortable hosting AWS setup scripts or documentation. We would recommend setting up a fork of GATK-SV or your own separate repository to host those scripts and documentation, and once they are ready, we would be happy to add a link from our README to your repository. 

To summarize: (1) please make the WDL changes without the `shared_filesystem` input and conditionals, and (2) please remove the AWS-specific documentation and scripts, to be hosted on your own repository or fork. Once you have completed your testing, let us know so we can test the changes on GCP before approving the PR.

spatel-gfb:
@epiercehoffman / @mwalker174 : The changes are made as per request and the pipeline was tested on our end. It worked fine.

Please review the changes and test it on your end. Shouldn't cause any issues ideally.

Once reviewed and merged, please create a tagged release specific to this release accommodating changes to work on AWS infrastructure,  so we can use in our AWS Setup templates.

wdl/MELT.wdl
```diff
@@ -600,12 +600,16 @@ task RunMELT {
...
+    # Converting to cp for FSx/Local
+    cp ~{bam_file} ~{sample_id}.bam
+    cp ~{bam_index_file} ~{sample_id}.bam.bai
 
     # these locations should be stable
     MELT_DIR="/MELT"
-    CROMWELL_ROOT="/cromwell_root"
+    # Adding sample id to ensure the MELT files are created in a separate folder even if multiple jobs run on same compute instance.
+    # Hence adding mkdir to create the folder
+    CROMWELL_ROOT="/cromwell_root/~{sample_id}"
```
epiercehoffman, line 611: It is on our to-do list to clean up the absolute paths in this workflow. We may request further changes in relation to this, or we may make those changes ourselves later on

epiercehoffman (changes requested):
Thanks for making the requested changes. A few more changes to request related to comments and using relative paths over absolute paths throughout. In addition, please update your branch with the changes from #329 which is necessary to sync up the workflows with the latest docker images.

I have also started testing your changes on GCP. I will continue once you make the requested changes and then let you know the outcome of the tests.

wdl/CollectCoverage.wdl
```diff
@@ -59,7 +59,10 @@ task CollectCounts {
       ~{sep=' ' disabled_read_filters_arr}
 
     sed -ri "s/@RG\tID:GATKCopyNumber\tSM:.+/@RG\tID:GATKCopyNumber\tSM:~{sample_id}/g" ~{sample_id}.counts.tsv
-    bgzip ~{sample_id}.counts.tsv
+    # Updated bgzip command to keep original file as its used by downstream process in case of FSx/local
+    # Also, this won't impact the GCP process
```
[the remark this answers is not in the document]
spatel-gfb: removed.

wdl/MELT.wdl
spatel-gfb: @epiercehoffman : are any changes needed in this from our side for your testing? It shouldn't be an issue on your end as well.

wdl/BAFFromGVCFs.wdl
```diff
@@ -215,6 +215,11 @@ task GenotypeGVCFs {
   command <<<
     set -e
 
+    # Adding this for FSx/local FS
```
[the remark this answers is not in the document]
epiercehoffman: Hi @spatel-gfb - the impact of this change seems likely to be specific to the platform. It would require increasing the size of the boot disk when running with cromwell on GCP, which would increase cost and introduce another resource whose size we need to determine and which could run out of space. And, if we increased the size of the boot disk and decreased the size of the main disk, the I/O speed would decrease, since I/O improves with increasing disk size on GCP. Due to the potential for a regression for GCP, we have concerns about merging this change and would prefer to retain the current behavior. It may be necessary going forward for your AWS fork to differ slightly in cases like this.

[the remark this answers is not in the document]
spatel-gfb: @epiercehoffman : would adding the shared_filesystem flag just for these processes work for you? Coz by default it will be false, but in case of AWS or other teams wanting to use local, can enable it ?

wdl/MELT.wdl
```diff
@@ -600,12 +600,16 @@ task RunMELT {
     set -Eeuo pipefail
 
     # MELT expects the BAM index to have extension ".bam.bai"
-    mv ~{bam_file} ~{sample_id}.bam
-    mv ~{bam_index_file} ~{sample_id}.bam.bai
+    # Converting to cp for FSx/Local
+    cp ~{bam_file} ~{sample_id}.bam
```
epiercehoffman, line 604: I think it's dangerous to make a copy of a BAM file as they can be >100GB in some cases. I did notice that this task ran out of disk when I tested, even for a relatively small BAM file. Can we soft-link instead?

spatel-gfb: @epiercehoffman we have added rm at the end and changed cp to mv to facilitate both type of runs. After double checking the bams are not used after this process so it was safe to add rm here.

wdl/BAFFromGVCFs.wdl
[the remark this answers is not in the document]
spatel-gfb: @epiercehoffman : any updates or comments on this ?

wdl/MELT.wdl
epiercehoffman: `mv` is fine on GCP, and `rm` at the end is fine but unnecessary as the local copy of the file will be deleted right afterwards, but for a shared file system I want to make sure you realize that MELT should be running in parallel with Manta and Wham and PESRCollection, which all also use the BAM. I imagine that is why you wanted to use `cp` in the first place. Again, happy to use `mv` here for GCP, but wanted to double check that this will accomplish your goals.

spatel-gfb: @epiercehoffman , yes but the good thing is MELT process runs longer than Manta, Whamg etc. so the removal as last step didnt affect our run.

spatel-gfb: @epiercehoffman : I guess now I see/understand what you are talking about. Could we have a solution for this in similar way as we are discussing for ImportGVCF and GEnotypeVCF as above ?

wdl/MELT.wdl
```diff
@@ -625,8 +628,9 @@ task RunMELT {
     echo "JVM memory: $JVM_MAX_MEM"
 
     # call MELT
+    cwd=$(pwd)
     "$MELT_SCRIPT" \
-      "~{sample_id}.bam" \
+      "$cwd/~{sample_id}.bam" \
```
mwalker174, line 633: It's not clear to my why this is needed, can't you just do `"~{sample_id}.bam" \` ?

wdl/MELT.wdl
mwalker174: Not worth fixing since we are deprecating melt support in the near feature. Your changes here are ok.

mwalker174 (changes requested):
Thanks for working with us on this @spatel-gfb and your patience (I was out a couple of weeks ago and still catching up). I have a few requests that you should apply throughout. To summarize:

1. Please delete added newlines
2. See my comment about /tmp on lustre. We need to stick with local temp dirs, but note this issue will be obviated for genomicsdbimport in the near future.
3. The new checks for indexes are ok but not ideal - really the indexes should already be generated and available in all cases. Please see my suggestion and apply throughout.
4. Anywhere you have copied files or deleted `rm` commands that would affect disk space, please adjust the default disk size accordingly by increasing the disk factor.

wdl/BAFFromGVCFs.wdl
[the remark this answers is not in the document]
spatel-gfb: @mwalker174 We already raised this issue with GATK ream : https://github.com/broadinstitute/gatk/issues/7646 coz its pertaining to genomicsdb tool and not cromwell. There suggestions still didnt improve the performance and the latest suggestion would need to be implemented on the pipeline level and hence we haven't test it yet and would need your teams help to test it out first.

The genomicsdb tool itself has performance issues with posix type of FS. Hence were using this workaround. Initially as per Emma's suggestion we added shared_fs flag. That would default be False (no changes ideally would be needed to run on GCP) but for AWS, we can pass it as true to handle it or even on GCP if someone is using posix type of FS.

We still think using that flag just for this and MELT wdl would be really helpful and get this out sooner instead of waiting few months for the newer/updated version of pipeline.

We welcome your inputs and would like to follow whatever is best for both the platforms.

wdl/MELT.wdl
spatel-gfb: @mwalker174 : Similar flag usage here would help both parties.

wdl/MELT.wdl
spatel-gfb: This is corrected now.

mwalker174:
@spatel-gfb Closing since this has been inactive for some time. Please re-open if you would like to submit further changes.

PR #311, opened 2022-03-14
Add section on alternative backends to readme

epiercehoffman:


mwalker174 (changes requested):
Thanks @epiercehoffman this was a solid start. I've updated with policies and guidelines. Can you also please add a clearly visible line near the top of `issue_template.md` linking to this section?

mwalker174 (approved):

PR #312, opened 2022-03-15
Increase EH VM memory size & refactor a task name

VJalili:


wdl/ExpansionHunter.wdl
```diff
@@ -134,7 +134,7 @@ task RunExpansionHunter {
 
     RuntimeAttr default_runtime_ = object {
         cpu_cores: 1,
-        mem_gb: 4,
+        mem_gb: 8,
```
mwalker174, line 137: Are you sure you want to increase the default? It seemed like 99% of shards worked with 4?

VJalili: You're right; based on the resource usage repots, analysis of only a few target loci need more than `4GB` RAM, and since we may drop those loci, I guess keeping default to `4GB` is wiser option. Hence, I'm closing this PR.

PR #315, opened 2022-03-18
Fix build_dockers.py error updating dockers.json

TedBrookings:
Previously build_dockers.py would erroneously update the gatk dockers, which it is not supposed to update.

build_dockers.py would erroneously update the gatk dockers, which it was
not supposed to maintain. This caused one of the images to clobber the
other. Now build_dockers.py only updates images it maintains.


epiercehoffman (approved):
Thanks for making this change so quickly. I tested it and it did not update the GATK docker when the build failed, so it seems like it is working properly.

PR #317, opened 2022-03-18
Add R mirrors

mwalker174:
Deletes microsoft R mirror, which has been unavailable for a few days now


TedBrookings (approved):
This PR should do what it intends (fix the problem with the Microsoft cran). The one thing I would suggest is decreasing the number of repos. I think two reliable repos should be enough for errors that are fixed by having a working repo. But this particular error causes failure if *any* repo goes down, so more repos -> more chance of failure.

Issue #81, opened 2020-11-16
Peg GATK release in sv-base Dockerfile


mwalker174:
We can first use a permanent commit from master once `PrintSVEvidence` is merged (currently a commit hash is used, which may change if the branch is deleted or rebased). Ultimately this should be pegged to a GATK release.

mwalker174:
Note (per #145) that `--skip-header` is not needed in the GATK release version of `PrintSVEvidence`

PR #324, opened 2022-03-29
Update GATK versions

mwalker174:
- Pegged GATK jar embedded in the sv-pipeline dockers to the latest release (4.2.6.1) per #81.
- Removed tabix commands, which are automatically generated by PrintSVEvidence.
- Removed unneeded `gatk_docker_pesr_override` input.

Testing of the single sample workflow and GatherBatchEvidence is underway.


mwalker174:
This was addressed in #719

PR #326, opened 2022-03-31
point to single sample QC definitions for single sample pipeline

epiercehoffman:
### Updates
Point single-sample test JSONs & Terra workspace data templates to single-sample version of `qc_definitions` input. This was previously an item in the reference panel values JSON, but got replaced by a cohort-mode version of the QC definitions file in #293, which causes many metrics to appear to be failing because the single-sample QC metrics are being compared against cohort-level metrics. This PR adds the same old single-sample QC definitions file to the reference resources JSON (to avoid repetition across multiple single sample values JSONs) and updates references to point to that version.

### Testing
All WDLs and JSONs built and validated with womtool. Confirmed workspace.tsv formatted correctly for single-sample Terra workspace and workspace.tsv and test JSONs contain the updated file path.


cwhelan (approved):
Looks good. I guess putting it in the reference resources is as good a place as any and doesn't require it to be an orphan gs URL hanging out in the JSON templates.

Issue #332, opened 2022-04-11
gatk-sv-single-sample workflow (v.0.9.1-beta-hotfix) failing at RunWhamgIncludelist subtask


riazgillani:
### Affected module(s) or script(s)

gatk-sv-single-sample workflow, Module00a, Whamg task, RunWhamgIncludelist subtask

### Affected version(s)

v.0.9.1-beta-hotfix

### Description 

I've successfully run this version of the workflow to call germline SVs on WGS CRAMs using a modified version of "GATKSVPipelineSingleSample.ref_panel_1kg.na12878.no_melt.json" for inputs. Over the last week, I've encountered recurrent failures in this workflow on multiple samples (including those on which this has previously run successfully), with the following error in the RunWhamgIncludelist subtask: "cat: /sys/fs/cgroup/memory/memory.stat: No such file or directory". I've attached the full log for broader context on this error.

#### Expected behavior

The gatk-sv-single-sample workflow should run to completion.

#### Actual behavior

The gatk-sv-single-sample workflow fails at this step. I've confirmed that this fails on samples on which the workflow has run successfully before, as well as new samples. I am providing a cram and matching crai for input, and I have not changed any other inputs to this workflow, so am having a hard time deciphering what may be contributing to this failure. Is it possible that the docker for this task has changed in some way?

[gatk_sv_single_sample_whamg_log_4_11_22.txt](https://github.com/broadinstitute/gatk-sv/files/8467327/gatk_sv_single_sample_whamg_log_4_11_22.txt)

riazgillani:
From what I am finding online, this error seems to occur when there is a change to the virtual environment. This may be an error inherent to the docker image used for this task ("gatksv/wham:8645aa"). Thank you for any direction you are able to provide here

cwhelan:
It seems like something might have changed in the way some GCE VMs are configured such that this `memory.stat` file is no longer in the same location. The line in question is only used for debugging memory issues occasionally encountered by Wham so it should be safe to remove from the WDL file, if you are able to modify it. We will remove the line in question and test out the change as well.

riazgillani:
Thanks Chris - the WDL I have is encapsulated, so I don't believe I can access and modify the command line. I appreciate your help looking into this

riazgillani:
Thank you for your help. I have a few questions:

[...]

PR #333, opened 2022-04-12
Option to provide credentials JSON to access CRAMs & fix bug in Whamg

epiercehoffman:
### Updates
* Add optional `String? service_account_json` input to `GatherSampleEvidence`. If provided, `CramToBam` will use the credentials JSON to access and localize the CRAM prior to converting to BAM. This is done in a separate task because the disk size cannot be estimated without access to the CRAM, so the localization task should be as short as possible, and then the regular `RunCramToBam` task can use dynamically-allocated resources. This input was not added to any test JSONs as it is not the default behavior.
* Fix `Whamg` bug from #332. The file `/sys/fs/cgroup/memory/memory.stat` no longer exists, possibly due to a change to GCE VMs. The comment suggests it was only accessed for debugging purposes, so the line was deleted to prevent the error.

### Testing
* Successfully ran `GatherSampleEvidence` (including `CramToBam` and `Whamg`) on one sample with a `service_account_json` required to access the CRAM file.  Also ran on one sample without the `service_account_json` to make sure that typical processing is unaffected, and `CramToBam` completed successfully.


wdl/CramToBam.wdl
```diff
@@ -1,15 +1,20 @@
...
 
 import "Structs.wdl"
+import "Utils.wdl" as util
 
 workflow CramToBam {
   input {
     File cram_file
     File reference_fasta
     File? reference_index
+    String? service_account_json  # Provide path to service account credentials JSON if required to access CRAM file
```
mwalker174, line 11: Make a note that neither the json nor cram can be in requester pays buckets in this case

wdl/CramToBam.wdl
```diff
@@ -22,6 +27,17 @@ workflow CramToBam {
...
+      input:
+        cloud_file_path = cram_file,
+        service_account_json = select_first([service_account_json]),
+        disk_size = select_first([localize_cram_disk_size, 50]),
+        cloud_sdk_docker = select_first([cloud_sdk_docker]),
+        runtime_attr_override = runtime_attr_localize_cram
+    }
+  }
+
   if (requester_pays) {
```
mwalker174, line 41:
```suggestion
  if (requester_pays && !defined(service_account_json)) {
```

wdl/Utils.wdl
```diff
@@ -313,3 +313,47 @@ task SubsetPedFile {
...
+    String cloud_file_path
+    String service_account_json
+    Int disk_size
+    String cloud_sdk_docker
+    RuntimeAttr? runtime_attr_override
+  }
+
+  RuntimeAttr default_attr = object {
+    cpu_cores: 1,
+    mem_gb: 3.75,
```
mwalker174, line 328: Is 3.75 optimal? Can you use 0.9?

wdl/CramToBam.wdl
epiercehoffman: If the cram is in a requester pays bucket, I think it would fail before this point because there is no billing project provided to the `gsutil cp` command in the localization task

wdl/Utils.wdl
epiercehoffman: Just tested and LocalizeCram completed successfully with 0.9GB memory on a new sample with a similar runtime, so seems fine to set that as default

mwalker174 (approved):

PR #334, opened 2022-04-14
Update Whamg.wdl

riazgillani:
Updated Whamg.wdl to delete cat /sys/fs/cgroup/memory/memory.stat, to allow for continued use of v.0.9.1-beta-hotfix


epiercehoffman:
See #332 for discussion

PR #342, opened 2022-04-29
Use GATK tool SVAnnotate for functional consequence annotation

epiercehoffman:
### Updates
Replace `svtk annotate` with GATK tool `SVAnnotate` for functional consequence annotation during `AnnotateVcf.wdl`. 
* Create standalone workflow `AnnotateFunctionalConsequences.wdl` that runs `SVAnnotate` with a protein-coding GTF and optionally a BED file of noncoding elements, with optional tool parameters exposed. 
* Call `AnnotateFunctionalConsequences` as a subworkflow during `AnnotateVcf.wdl`. Scattering should not be necessary for a small / medium sized cohort as SVAnnotate runs in ~25-30 mins on an unfiltered VCF with ~2500 samples, so scattering was not implemented, but it may need to be run separately per contig for very large cohorts.
* Update single-sample WDL as needed
* Update JSON & Terra templates as needed
* Remove old annotation WDLs
* Change custom canonical protein-coding GTF to MANE Select Plus Clinical GTF

### Testing
* Ran `AnnotateVcf.wdl` successfully with 1KG ref panel with cromshell
* Validated all WDLs and JSONs with womtool (decrease to 27 test JSONs) & Terra validation script


wdl/AnnotateFunctionalConsequences.wdl
```diff
@@ -0,0 +1,95 @@
...
+
+  output {
+    File annotated_vcf = "~{outfile}"
+    File annotated_vcf_index = "~{outfile}.tbi"
+  }
+  command <<<
+
+     set -euo pipefail
+
+     gatk --java-options "-Xmx~{java_mem_mb}m" SVAnnotate \
```
mwalker174, line 77: What do you think about adding an `String additional_args` input that would let the user specify arbitrary arguments not built into the wdl (e.g. a requester pays project id)?

mwalker174 (approved):
Great work! Always happy to net 800 fewer lines and simplify the WDLs while fixing bugs and improving performance! A couple of minor suggestions.

epiercehoffman:
> Just checking: does the GATK docker need to be updated in `dockers.json`?

No, the current `gatk_docker` (the nightly snapshot) has SVAnnotate and that's what I used for testing.

Updated the way optional arguments are handled, added an optional additional_args parameter, validated with womtool, and tested successfully with default inputs (did not try with providing additional_args).

PR #348, opened 2022-05-24
Tb decrease docker count

### LINKED ISSUE #261: Consolidate dockers and update docker packages and Python
Many updates are in sv-pipeline-updates an need to be moved over. Also we should move exclusively to Python3 if possible (see #205).

TedBrookings:
Consolate dockers
    * Decreased number of dockers, merging most into sv-pipeline
    * Updated software package versions (e.g. numpy, bcftools)
    * Changes to WDLs and scripts to accommodate new behavior of updated packages and tools
    * Improvements to reproducibility and determinism of outputs
Closes #261


TedBrookings:
I've left this as several commits so that some of the logic / reasoning of the changes can be more clear. I'll leave a few high-level notes here.

1. The overall docker strategy is similar, in that changing code in the git repo should not require re-building python or R dependencies.
    a. Now only sv-pipeline depends on that code. All other downstream images have been combined into sv-pipeline.
    b. The gatk dependency is moved to sv-base so that it too can be updated without rebuilding python or R
    c. Most Docker images are slightly larger, due to larger updated packages, or combining packages into one image. They are not _significantly_ larger.
2. Packages were updated, and package version pinning is more effective
    a. Most packages and tools (python, R, bcftools, etc) were updated to their latest version
    b. The installed version of packages should now accurately reflect the version in Dockerfiles
    c. pysam was updated to an older version: the newest that is compatible with our current BND strategy.
3. Updating the packages and tools resulted in multiple errors due to default package / tool behavior changing
    a. Consequently multiple WDLs and scripts were changed: these are organized in additional commits
    b. In the course of debugging these problems, I encountered several unnecessary non-deterministic behaviors, and a section of svtk BAFpsyam that was not converging, and corrected those as well.
4. I've done a quick review of the output, compared to master. I also ran `svtest vcf` on my branch and on master. I've attached relevant files to this comment. To me it appears that the largest difference is that the new branch is passing around 5% more DUPs, predominantly > 5000bp and < 0.1 AF. My guess is this is due to fixing the convergence problem with the BAF calculation, although that's just a guess.

[master_qc_main_plots.tar.gz](https://github.com/broadinstitute/gatk-sv/files/8764822/master_qc_main_plots.tar.gz)
[branch_qc_main_plots.tar.gz](https://github.com/broadinstitute/gatk-sv/files/8764824/branch_qc_main_plots.tar.gz)
[side_by_side_sv_test_results.txt](https://github.com/broadinstitute/gatk-sv/files/8764828/side_by_side_sv_test_results.txt)

TedBrookings:
Just a note as to why there are a few post-PR pushes including a re-base: previously the docker build was failing due to the GitHub action's VM running out of disk. It was necessary to update build_docker.py (in another PR) to prune in between image builds in order to keep under the VM's small available disk limit. I also made a minor change to the location of the GATK jar build in order to decrease disk usage. Nothing should have substantially changed pipeline output.

wdl/TrainSRGenotyping.wdl
```diff
@@ -97,7 +97,7 @@ task GenotypeSRPart1 {
 
   RuntimeAttr default_attr = object {
     cpu_cores: 1, 
-    mem_gb: 3.75,
+    mem_gb: 6,
```
mwalker174, line 100: Assume this is necessary now, but any idea what's using more memory?

dockerfiles/sv-base-mini/Dockerfile
```diff
@@ -1,77 +1,102 @@
...
+ENV PATH=/opt/samtools/bin:/opt/bcftools/bin:/opt/vcftools/bin:/opt/bedtools/bin:/opt/bin:$PATH
+# note: need to specify path to perl libraries or tools from vcftools won't work
+RUN LIB_LOC=$(find /opt/vcftools/share/perl -mindepth 1 -type d) && \
+    ln -s $LIB_LOC /opt/vcftools/share/perl_lib
+ENV PERL5LIB=/opt/vcftools/share/perl_lib
 
-#### Basic utilities
+# establish file storing required apt packages (so subsequent stages know which packages can be removed and which can't)
+ARG APT_REQUIRED_PACKAGES
+ENV APT_REQUIRED_PACKAGES=$APT_REQUIRED_PACKAGES
```
mwalker174, line 76: A bit confusing - do you really need the ENV here? Could at least give them different names

mwalker174 (changes requested):
Thank you for all this work @TedBrookings! Mostly looks good, a few minor comments/questions. The BAF change seems to have increased CNV true positives as well against Sudmant et al. 

The only major request I have is to update the WDLs and input templates/values to reflect the new dockers (e.g. replacing all `sv_pipeline_rdtest_docker` with `sv_pipeline_docker`).

src/sv-pipeline/04_variant_resolution/scripts/pipeline_genotype_updated.sh
```diff
@@ -206,10 +206,18 @@ less cnv.copystate.multi.count.txt|awk '{if ($1>3) print $2}'|fgrep -wf - cnv.be
...
+  | sort -u \
+  > repeat.breakpoint.fail.ids.txt
 
 ##depth##
-cat /data/talkowski/rlc47/src/GRCh37.*.RMSK.merged.bed /data/talkowski/rlc47/src/GRch37.segdups_gaps_abParts_heterochrom.lumpy.exclude.bed|sort -k1,1 -k2,2n |coverageBed -a <(sort -k1,1 -k2,2n cnv.bed) -b -|awk '{if ($NF>0.3) print $4}' >repeat.depth.fail.ids.txt
+cat /data/talkowski/rlc47/src/GRCh37.*.RMSK.merged.bed /data/talkowski/rlc47/src/GRch37.segdups_gaps_abParts_heterochrom.lumpy.exclude.bed \
+| sort -k1,1 -k2,2n \
+| bedtools coverage -a <(sort -k1,1 -k2,2n cnv.bed) -b - \
+| awk '{if ($NF>0.3) print $4}' \
+> repeat.depth.fail.ids.txt
```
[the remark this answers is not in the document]
TedBrookings: Deleted

wdl/TrainSRGenotyping.wdl
TedBrookings: It's been a little while since I made this change, but I think I remember most of the details. I didn't investigate very fully, in part because it takes awhile to get to this part of the pipeline, so I tend to favor solutions that kill problems dead, rather than screwing around with long-shots.

1. This calls SR_genotype.opt_part1.sh, which has a few algorithms that could be implemented in a much more memory efficient manner. The location of the crash I was getting was the section on lines 144-152, which constructs a long list of patterns to search, and then does fgrep -wf to search them. I considered trying to rewrite this section to be more memory efficient, but it would alter the sv_pipeline docker (invalidating my call cache) and make reviewing harder. So I bumped up memory instead because it was a small task and didn't take all that long to complete (maybe around 5-10 minutes if I recall correctly).
2. As for why it's using more memory now than it used to, I'd guess we were previously pretty close to the line (there's no VM scaling in this task) and my fixes to the BAF processing earlier in the pipeline are retaining more variants.

mwalker174: As long as it doesn't increase cost substantially that's fine

mwalker174 (approved):
@TedBrookings Thanks for addressing the comments! Can you just open a ticket to remove the deprecated dockers from the WDLs and inputs?

TedBrookings:
It has been a LONG slog but I finally got the single-sample pipeline to run successfully. The complete workflow outputs are in gs://broad-dsde-methods-tbrookin/results/GATKSVPipelineSingleSample/branch/, and I've attached the final VCF:
[test_NA12878.annotated.final_cleanup.vcf.gz](https://github.com/broadinstitute/gatk-sv/files/8879064/test_NA12878.annotated.final_cleanup.vcf.gz)

Let me know if these results are acceptable and I will rebase and merge.

mwalker174 (approved):
@TedBrookings Single-sample metrics are identical to a previous run

PR #350, opened 2022-05-26
Update ExpansionHunter Dockerfile

VJalili:
- Update the Dockerfile to use a custom-built of ExpansionHunter;
- Update htslib, bcftools, and samtools to their current latest versions;
- Remove TRTools and GangSTR as they are not currently used in ExpansionHunter WDL (though GangSTR has its own WDL that is currently under development);
- Copy `src/str` to the VM as it contains scripts needed in ExpansionHunter WDL.


mwalker174 (changes requested):
This looks good. Can you actually move this to your other PR? It depends on having `src/str` which doesn't exist, so these two PRs should actually go in together.

PR #356, opened 2022-06-10
Refactor task input variable

VJalili:
The input and output variables are named the same. Even though it passes `womtool` validation and runs successfully on Cromwell, it fails on Terra. This PR renames the variable.


mwalker174 (approved):
Good to know, I wasn't aware of this particular issue

VJalili:
Thanks, @mwalker174! It was a tricky one since it worked as expected locally and on Cromwell but fail on Terra.

PR #359, opened 2022-06-17
Fix STRANDS and EV related errors in svtk vcfcluster

mwalker174:
Updates `svtk vcfcluster` to handle vcfs where:

- Any non-BND record types may omit the STRANDS info field
- EV format fields may have string list type


TedBrookings (approved):
This looks fine, other than the horror of base_record not being a record (which pre-dates this PR and probably would take a lot of work to fix). I made a few "pythonic" suggested changes, but nothing critical.

mwalker174:
Thanks for the feedback @TedBrookings. This is sort of time-sensitive so I'm going to go ahead and merge, but I learned a few things from your comments.

PR #362, opened 2022-06-23
svtk vcfcluster bug fixes

mwalker174:
-Allows no-call alleles (`.`) in genotype selection function `choose_best_genotype()`, which is now rewritten in a leaner, faster form
-Fix String INFO field merging issue that was causing some values to turn into comma-delimited lists of characters, e.g. `example` -> `e,x,a,m,p,l,e`


src/svtk/svtk/utils/genotype_merging.py
```diff
@@ -23,41 +23,31 @@ def choose_best_genotype(sample, records):
...
-            continue
-        # if new GT  = best_GT, while found a higher GQ, replace GQ
+    # Returns true if the GT is called and non-ref
+    # TODO: does not handle multi-allelic CNVs
+    def _is_non_ref(format_fields):
+        return 'GT' in format_fields and sum(g for g in format_fields['GT'] if g is not None) > 0
+
+    def _get_mingq_index(format_fields_list):
+        gqs = [f['GQ'] for f in format_fields_list]
+        min_gq = min([x for x in gqs if x is not None], default=None)
```
TedBrookings, line 33: No need to construct the list, min can operate on generator
```suggestion
        min_gq = min((x for x in gqs if x is not None), default=None)
```

src/svtk/svtk/svfile.py
```diff
@@ -282,10 +282,10 @@ def merge_record_infos(self, new_record, header):
         """
         PROTECTED_INFOS = ('SVTYPE CHR2 END STRANDS SVLEN ALGORITHMS CIPOS '
                            'CIEND RMSSTD MEMBERS').split()
-        records = [r.record for r in self.records]
         infos = defaultdict(list)
 
-        for record in records:
+        for sv_record in self.records:
+            record = sv_record.record
             for info, value in record.info.items():
```
TedBrookings, line 289: I know this predates your changes, but I kind of hate that the key is referred to as `info`, which just creates confusion. I'd suggest renaming it to `key` or `info_key`. (Both here and below)

src/svtk/svtk/utils/genotype_merging.py
```diff
@@ -23,41 +23,31 @@ def choose_best_genotype(sample, records):
...
-            best_GT = record.samples[sample]['GT']
-            best_GQ = record.samples[sample]['GQ']
-            best_record = record
-        elif record.samples[sample]['GT'] == (0, 0) and best_GT != (0, 0):
-            continue
-        # if new GT  = best_GT, while found a higher GQ, replace GQ
+    # Returns true if the GT is called and non-ref
+    # TODO: does not handle multi-allelic CNVs
+    def _is_non_ref(format_fields):
+        return 'GT' in format_fields and sum(g for g in format_fields['GT'] if g is not None) > 0
```
TedBrookings, line 29: I don't know how long this task takes to run. I've dealt with situations where a very large data set winds up having its allele count computed. In this situation, this multi-level looping (loop over GTs, within each GT sum over loop over allele) is quite slow in python. It can be faster if you store the results in a look-up map like so:
```suggestion
_ac_map = {}
    def _is_non_ref(format_fields):
        gt = format_fields.get('GT', None)
        if gt in _ac_map:
            return _ac_map[gt] > 0
        else:
            ac = 0 if gt is None else sum(1 for a in gt if a is not None)
            _ac_map[gt] = ac
            return ac > 0
```

TedBrookings (approved):
Looks good to me. I made a few suggestions that may not be hyper relevant if this task typically completes in fractions of a second, but could help if it's running for a long time.

src/svtk/svtk/utils/genotype_merging.py
mwalker174: Done

src/svtk/svtk/svfile.py
mwalker174: Done, and cleaned up a bit

src/svtk/svtk/utils/genotype_merging.py
mwalker174: Yes I've done this before for sample overlap comparisons. In this particular case, `_is_non_ref()` is only called once per record, so I don't think it saves time?

PR #364, opened 2022-06-29
KN Picard Metrics Aggregation Script

krithikanathamuni:


src/sv-pipeline/scripts/aggregate_picard.py
```diff
@@ -0,0 +1,276 @@
...
+    bddf6mean = dataframe6.mean()
+    
+    bddf1meanvalues = pd.DataFrame(bddf1mean).T.drop('CYCLE', axis = 1).add_suffix('_'+str(window))
+    bddf2meanvalues = pd.DataFrame(bddf2mean).T.drop('CYCLE', axis = 1).add_suffix('_'+str(window*2))
+    bddf3meanvalues = pd.DataFrame(bddf3mean).T.drop('CYCLE', axis = 1).add_suffix('_'+str(window*3))
+    bddf4meanvalues = pd.DataFrame(bddf4mean).T.drop('CYCLE', axis = 1).add_suffix('_'+str(window*4))
+    bddf5meanvalues = pd.DataFrame(bddf5mean).T.drop('CYCLE', axis = 1).add_suffix('_'+str(window*5))
+    bddf6meanvalues = pd.DataFrame(bddf6mean).T.drop('CYCLE', axis = 1).add_suffix('_'+str(window*6))
+    
+    BDMresult = pd.concat([bddf1meanvalues, bddf2meanvalues, bddf3meanvalues, bddf4meanvalues, bddf5meanvalues, bddf6meanvalues], axis = 1)
```
asmirnov239, line 67: Variable names (unless they are constants) should be lower case, separated by underscores.

src/sv-pipeline/scripts/aggregate_picard.py
```diff
@@ -0,0 +1,276 @@
...
+        onerow.loc[i] = nextrow
+    onerow = onerow.drop(0)
+    return onerow
+
+def concatrows(table, suffix):
+    newdata = pd.DataFrame()
+    for row in range(1, len(table) + 1):
+        addrow = pd.DataFrame(table.loc[row]).T.reset_index(drop=True)
+        addrow = addrow.add_suffix(suffix[row-1])
+        newdata = pd.concat([newdata, addrow], axis = 1)
```
asmirnov239, line 42: Dataframe concatenations as well as creation are expensive operations and should be avoided if possible. You should be able to rewrite this function to avoid these operations. Feel free to reach out if you need any help in this.

src/sv-pipeline/scripts/aggregate_picard.py
```diff
@@ -0,0 +1,276 @@
...
+    dataframe.columns = columnnames
+    dataframe_length = len(dataframe)
+    dataframe.loc[dataframe_length] = values
+    dataframe = dataframe.drop(0)
+    return dataframe
+
+def addrows(file_name, label):
+    onerow = newtable(file_name, label)
+    for i in range(len(read_in(file_name, label))):
+        nextrow = read_in(file_name, label)[i].split("\t")
```
asmirnov239, line 32: As above, here you are traversing the same file in each iteration - in general I/O is a bottleneck in many cases and we only want to read the files into memory once. 

You can rewrite this function to  read the file content into an array once and then fill in all the rows of the dataframe in one go.

src/sv-pipeline/scripts/aggregate_picard.py
```diff
@@ -0,0 +1,276 @@
...
+        newdata = pd.concat([newdata, addrow], axis = 1)
+    return newdata
+
+def window_metrics_mean(table, window):
+    dataframe1 = pd.DataFrame(table[0:window]).reset_index(drop = True).apply(pd.to_numeric, errors='ignore')
+    dataframe2 = pd.DataFrame(table[window:window*2]).reset_index(drop = True).apply(pd.to_numeric, errors='ignore')
+    dataframe3 = pd.DataFrame(table[window*2:window*3]).reset_index(drop = True).apply(pd.to_numeric, errors='ignore')
+    dataframe4 = pd.DataFrame(table[window*3:window*4]).reset_index(drop = True).apply(pd.to_numeric, errors='ignore')
+    dataframe5 = pd.DataFrame(table[window*4:window*5]).reset_index(drop = True).apply(pd.to_numeric, errors='ignore')
+    dataframe6 = pd.DataFrame(table[window*5:window*6]).reset_index(drop = True).apply(pd.to_numeric, errors='ignore')
```
asmirnov239, line 51: There is a lot of duplicated code here (and in other functions below), which is a sign that the code can benefit from refactoring (https://refactoring.guru/smells/duplicate-code).

One simple way to refactor it is to add an inner function inside the `window_metrics_mean` that creates the mean and constructs the dataframe and then invoke this function 6 times with different arguments.

src/sv-pipeline/scripts/aggregate_picard.py
```diff
@@ -0,0 +1,276 @@
...
+    bddf3meanvalues = pd.DataFrame(bddf3mean).T.drop('CYCLE', axis = 1).add_suffix('_'+str(window*3))
+    bddf4meanvalues = pd.DataFrame(bddf4mean).T.drop('CYCLE', axis = 1).add_suffix('_'+str(window*4))
+    bddf5meanvalues = pd.DataFrame(bddf5mean).T.drop('CYCLE', axis = 1).add_suffix('_'+str(window*5))
+    bddf6meanvalues = pd.DataFrame(bddf6mean).T.drop('CYCLE', axis = 1).add_suffix('_'+str(window*6))
+    
+    BDMresult = pd.concat([bddf1meanvalues, bddf2meanvalues, bddf3meanvalues, bddf4meanvalues, bddf5meanvalues, bddf6meanvalues], axis = 1)
+    return BDMresult
+
+def window_metrics_std(table, window):
+    mqdf1 = pd.DataFrame(table[0:window]).reset_index(drop = True).apply(pd.to_numeric, errors='ignore')
```
asmirnov239, line 71: Duplicate code here and below as well.

src/sv-pipeline/scripts/aggregate_picard.py
```diff
@@ -0,0 +1,276 @@
...
+    mqdf5stdtable = pd.DataFrame(mqdf5std).T.drop('CYCLE', axis = 1).add_suffix('_'+str(window*5))
+    mqdf6std = mqdf6.std().rename({'MEAN_QUALITY':'SD'})
+    mqdf6stdtable = pd.DataFrame(mqdf6std).T.drop('CYCLE', axis = 1).add_suffix('_'+str(window*6))
+
+    MQMresult = pd.concat([mqdf1stdtable, mqdf2stdtable, mqdf3stdtable, mqdf4stdtable, mqdf5stdtable, mqdf6stdtable], axis = 1)
+    return MQMresult
+
+def sequencing_artifact_metrics(table):
+    sadfone = pd.DataFrame(table.loc[1]).T.reset_index(drop = True)
+    sadf1 = sadfone.add_suffix("_1").drop(['SAMPLE_ALIAS_1', 'LIBRARY_1', 'WORST_CXT_1',
```
asmirnov239, line 96: This can also be refactored to get rid of the duplicate code, but also to store the strings `SAMPLE_ALIAS`, `SAMPLE_ALIAS_1` strings into variables. This way if the name of the Picard attribute changes you only need to change one variable - but also there is less chance of having a typo somewhere that will break the code.

src/sv-pipeline/scripts/aggregate_picard.py
```diff
@@ -0,0 +1,276 @@
...
+    sadftwelve = pd.DataFrame(table.loc[12]).T.reset_index(drop = True)
+    sadf12 = sadftwelve.add_suffix("_12").drop(['SAMPLE_ALIAS_12', 'LIBRARY_12', 
+                                            'WORST_CXT_12', 'WORST_CXT_QSCORE_12', 'WORST_PRE_CXT_12', 
+                                            'WORST_PRE_CXT_QSCORE_12', 'WORST_POST_CXT_12',
+                                       'WORST_POST_CXT_QSCORE_12'], axis = 1)
+    SAMresult = pd.concat([sadf1,sadf2, sadf3, sadf4, sadf5, sadf6, sadf7, sadf8, sadf9, sadf10, sadf11, sadf12 ], axis = 1)
+    return SAMresult
+
+def windows(table, window):
+    if len(table.columns) == 2:
```
asmirnov239, line 159: I think this is fine to use in the script, but in general having the output of the function be conditioned on the hardcoded variable like this is a sign that the code should be refactored. 

One way to refactor it is to rewrite this code in an object-oriented manner. Again, I don't think it is necessary here but for code that is more than few hundred lines long it's a good idea, and it will provide more readability and robustness.

src/sv-pipeline/scripts/aggregate_picard.py
```diff
@@ -0,0 +1,276 @@
...
+
+def mainfile(index):
+    allsamplecolumns = ['alignment_summary_metrics', 'base_distribution_by_cycle_table', 'gc_bias_summary_metrics', 
+                        'insert_size_metrics', 'mean_quality_by_cycle_table', 
+                        'sequencing_artifact_summary_metrics', 'quality_score_table', 'quality_yield_metrics', 
+                        'raw_wgs_metrics', 'wgs_metrics']
+    table_name = "sample"
+    samples = pd.read_csv(io.StringIO(fiss.fapi.get_entities_tsv(project, workspace, 'sample').text), sep='\t')
+    samples.rename(columns = {'entity:sample_id':'sample'}, inplace = True)
+#     specificcolumns = samples[['alignment_summary_metrics', 'base_distribution_by_cycle_table', 'gc_bias_summary_metrics',
```
asmirnov239, line 174: Remove old code here.

src/sv-pipeline/scripts/aggregate_picard.py
```diff
@@ -0,0 +1,276 @@
...
+    samples.rename(columns = {'entity:sample_id':'sample'}, inplace = True)
+#     specificcolumns = samples[['alignment_summary_metrics', 'base_distribution_by_cycle_table', 'gc_bias_summary_metrics', 
+#                         'insert_size_metrics', 'mean_quality_by_cycle_table', 
+#                         'sequencing_artifact_summary_metrics', 'quality_score_table', 'quality_yield_metrics', 
+#                         'raw_wgs_metrics', 'wgs_metrics']]
+#     dropemptyrows = specificcolumns.dropna(axis = 0)
+#     dropemptycolumns = dropemptyrows.dropna(axis = 1)
+#     files = ! ls
+    
+    Dict = {}
```
asmirnov239, line 182: Make the variable name more descriptive and lower case. `dict` is also name for a built-in type in Python and typing library has a class `Dict` so it's best to avoid using either.

src/sv-pipeline/scripts/aggregate_picard.py
```diff
@@ -0,0 +1,276 @@
...
+        
+                
+#         duplicates = find_duplicates(newrow)
+#         print(duplicates)
+        result = result.loc[:, ~result.columns.duplicated()].copy()
+        newrow = newrow.loc[:, ~newrow.columns.duplicated()].copy()
+        result = pd.concat([result, newrow], ignore_index = True)
+    return result
+        
+
```
asmirnov239, line 275: Extra spaces here.

asmirnov239 (commented):
@krithikanathamuni Thank you for putting this together. I made a few general comments on code structure and formatting, as well as more specific suggestions on code refactoring. Let me know if you have any questions!

src/sv-pipeline/scripts/aggregate_picard.py
mwalker174: +1. Reading files from disk is orders of magnitude slower than in-memory operations. I'd put my money on this being where you're seeing the biggest slowdowns.

PR #370, opened 2022-07-08
switch to new CondenseDepthEvidence gatk tool

tedsharpe:


mwalker174 (approved):
Thank you @tedsharpe! Outputs check out, and I think we're good to merge here.

PR #372, opened 2022-07-13
Enable docker build on m1 macs

TedBrookings:
Two minor changes, both important. Specifying platform in the docker build allows the mac to emulate an intel chip, and installing rosetta makes that emulation better (without that, java compilation via gradle seems to hang).

I successfully ran `./scripts/docker/build_docker.py --targets all --output-json /dev/null`

I tested my own docker (sv_utils_docker) because it contains a test suite. I ran in on my linux box and the test suite completed just fine.

For anyone attempting to build many images on the macbook pro, I recommend adding the option `--prune-after-each-image` to free up resources. I also set up my Docker Desktop to use: 5 CPUs, 20 GB of RAM, 2 GB of swap. Insufficient resources can result in crashes, or just hanging and failing to progress after a certain point.

### DIFF
--- a/scripts/docker/README.md
+++ b/scripts/docker/README.md
@@ -43,3 +43,5 @@ Conda packages `termcolor` and `pprint` are assumed to be available.
 * `test_build_docker.ipynb` is for debugging and testing purposes
 
 * Use `update_json_docker.sh` to update dockers tags across all json parameter files
+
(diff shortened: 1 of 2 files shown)


mwalker174 (approved):
Thanks, just a minor suggestion since not everyone might recognize "Apple silicon"

PR #381, opened 2022-08-05
Bug fix for ParseGenotyping

mwalker174:
Replaces a check for compressed inputs `process_posthoc_cpx_depth_regenotyping.sh` from GenotypeComplexVariants. The previous implementation used the `file` command to guess the file type, but it is not 100% reliable and recently misidentified a gzipped bed file as a boot sector.

This commit replaces the `file` check with a simple check for `.gz` extensions. Tested successfully on MakeCohortVcf.

### DIFF
--- a/dockerfiles/igv/makeigvpesr_cram.py
+++ b/dockerfiles/igv/makeigvpesr_cram.py
@@ -46,9 +46,9 @@ def ped_info_readin(ped_file):
         pin = line.strip().split()
         if not pin[1] in out.keys():
             out[pin[1]] = [pin[1]]
-        if not(pin[2]) == 0:
+        if not (pin[2]) == 0:
             out[pin[1]].append(pin[2])
-        if not(pin[3]) == 0:
+        if not (pin[3]) == 0:
             out[pin[1]].append(pin[3])
     fin.close()
     return out
@@ -62,7 +62,7 @@ def cram_info_readin(cram_file):
         if not pin[0] in out.keys():
             out[pin[0]] = pin[1:]
     fin.close()
-    return(out)
+    return (out)
 
 
 ped_info = ped_info_readin(args.ped)
--- a/dockerfiles/igv/makeigvpesr_trio.py
+++ b/dockerfiles/igv/makeigvpesr_trio.py
@@ -44,9 +44,9 @@ def ped_info_readin(ped_file):
         pin = line.strip().split()
         if not pin[1] in out.keys():
             out[pin[1]] = [pin[1]]
-        if not(pin[2]) == 0:
+        if not (pin[2]) == 0:
             out[pin[1]].append(pin[2])
-        if not(pin[3]) == 0:
+        if not (pin[3]) == 0:
             out[pin[1]].append(pin[3])
     fin.close()
     return out
@@ -60,7 +60,7 @@ def cram_info_readin(cram_file):
         if not pin[0] in out.keys():
             out[pin[0]] = pin[1:]
     fin.close()
-    return(out)
+    return (out)
 
 
 # ped_info = ped_info_readin(args.ped)
--- a/src/sv-pipeline/04_variant_resolution/scripts/process_posthoc_cpx_depth_regenotyping.sh
+++ b/src/sv-pipeline/04_variant_resolution/scripts/process_posthoc_cpx_depth_regenotyping.sh
@@ -103,7 +103,7 @@ if ! [ -s ${INVCF} ]; then
   usage
   exit 0
 fi
-if [ $( file ${INVCF} | fgrep "gzip" | wc -l ) -lt 1 ]; then
+if [[ ${INVCF} != *.gz ]]; then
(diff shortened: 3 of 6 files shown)


src/sv-pipeline/04_variant_resolution/scripts/process_posthoc_cpx_depth_regenotyping.sh
```diff
@@ -103,7 +103,7 @@ if ! [ -s ${INVCF} ]; then
   usage
   exit 0
 fi
-if [ $( file ${INVCF} | fgrep "gzip" | wc -l ) -lt 1 ]; then
+if [[ ${INVCF} != *.gz ]]; then
```
VJalili, line 106: I suppose this check can be "tricked" if one creates any file (compressed or uncompressed) and just adds `.gz` extension. Otherwise, I guess this check is good enough.

dockerfiles/igv/makeigvpesr_cram.py
```diff
@@ -46,9 +46,9 @@ def ped_info_readin(ped_file):
         pin = line.strip().split()
         if not pin[1] in out.keys():
             out[pin[1]] = [pin[1]]
-        if not(pin[2]) == 0:
+        if not (pin[2]) == 0:
```
TedBrookings, line 49: If you're going to adjust this for style, this line is very problematic. pin is an array of strings, so `not (pin[1]) == 0` always evaluates to True. If the goal is to only append non-trivial keys, it should be changed to:
```suggestion
        if pin[2] != "0":
```
Otherwise the if statement should be removed.
Similarly with pin[3]

dockerfiles/igv/makeigvpesr_cram.py
```diff
@@ -62,7 +62,7 @@ def cram_info_readin(cram_file):
         if not pin[0] in out.keys():
             out[pin[0]] = pin[1:]
     fin.close()
-    return(out)
+    return (out)
```
TedBrookings, line 65: Parenthesis have no effect
```suggestion
    return out
```

dockerfiles/igv/makeigvpesr_trio.py
```diff
@@ -44,9 +44,9 @@ def ped_info_readin(ped_file):
         pin = line.strip().split()
         if not pin[1] in out.keys():
             out[pin[1]] = [pin[1]]
-        if not(pin[2]) == 0:
+        if not (pin[2]) == 0:
```
TedBrookings, line 47: Same deal as in makeigvpesr_cram:
```suggestion
        if pin[2] != "0":
```

dockerfiles/igv/makeigvpesr_trio.py
```diff
@@ -60,7 +60,7 @@ def cram_info_readin(cram_file):
         if not pin[0] in out.keys():
             out[pin[0]] = pin[1:]
     fin.close()
-    return(out)
+    return (out)
```
TedBrookings, line 63: Parenthesis have no effect
```suggestion
    return out
```

src/sv-pipeline/scripts/downstream_analysis_and_filtering/apply_minGQ_filter.py
```diff
@@ -300,7 +300,7 @@ def _get_minGQ(record, minGQ_dict, SVLEN_table, AF_table, SVTYPE_table,
         for key in list(set(EV_table.values())):
             minGQ[key] = int(globalMin)
 
-    return(minGQ)
+    return (minGQ)
```
TedBrookings, line 303: Parenthesis have no effect
```suggestion
    return minGQ
```

src/svtk/svtk/pesr/pe_test.py
```diff
@@ -100,7 +100,7 @@ def _get_coords(pos, strand):
         i = 0
         for pair in pairs:
             pair = _DiscPair(*pair)
-            if(i > 1000000):
+            if (i > 1000000):
```
TedBrookings, line 103: Parenthesis have no effect
```suggestion
            if i > 1000000:
```

src/sv-pipeline/04_variant_resolution/scripts/process_posthoc_cpx_depth_regenotyping.sh
TedBrookings: In my opinion this is much more stable. When I updated our docker images, the output of `file` changed in a way that made the checks break for some files. If parsing the tool output causes breaks when you change Ubuntu versions, it seems better to me to just check the extension. Especially since this code will only be parsing output from previous scripts in our pipeline: if it breaks this way, it's our fault.

TedBrookings (approved):
In my opinion this is a substantial stability improvement.
However, I noticed several lines of code that you touched that don't do what they "seem" to do, and I believe are resulting in unintended behavior. I'd strongly suggest checking them out.

PR #385, opened 2022-08-08
Validate WDLs using miniwdl

VJalili:
This PR implements a python script that uses the miniwdl python bindings to validate (only errors) a list of given WDL files.


VJalili:
Linting errors are not related, so I'd rather leave fixing them for another PR.

scripts/test/miniwdl_validation.py
```diff
@@ -0,0 +1,61 @@
...
+    except WDL.Error.ImportError as e:
+        print(e)
+        print(format_error_message("Import Error", e, e.pos))
+    finally:
+        return False
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("files", nargs="+", help="The WDLs files to be checked.")
```
mwalker174, line 40:
```suggestion
    parser.add_argument("files", nargs="+", help="The WDL files to be checked.")
```

scripts/test/miniwdl_validation.py
```diff
@@ -0,0 +1,61 @@
...
+    args = parser.parse_args()
+
+    wdl_filenames = []
+    for filename in args.files:
+        if filename.endswith(".wdl"):
+            wdl_filenames.append(filename)
+    if len(wdl_filenames) != len(args.files):
+        print(f"Warning! Of {len(args.files)} provided files, {len(wdl_filenames)} are WDL.")
```
mwalker174, line 48: Is there a particular use case you're thinking of with this check? I could see this being useful if a directory were being searched, but since each file is specified I would simply expect validation to fail if one of them isn't a WDL.

VJalili: Would this be a useful check if the script is invoked as `python miniwdl_validation.py wdl/*.*` where the `wdl` directory contains files other than `*.wdl`?

mwalker174: I see yes that would be useful, although we usually don't need to validate every wdl, just ones that are part of the "canonical" pipeline i.e. have test inputs available (see validation.sh script for how we do that). Either way, fine to leave it like this.

VJalili: That makes sense. I removed this filtering to simplify the script.

mwalker174 (approved):

PR #387, opened 2022-08-17
bug fixed in concatinate script

xuefzhao:
put in a fix for the function that concatenate vapor results to add header to the output


epiercehoffman (approved):
Tested and header (with # in front) was retained in final concatenated output

Issue #44, opened 2020-09-08
Outlier Sample Removal in GenerateBatchMetrics and FilterBatchSites


hab45:
We have found certainly outlier samples with significantly more variants can influence variant metrics (module 2) and subsequent filtering (module 3). The challenge here is that given some family based studies have more tolerance for inclusion of these sample we are suggesting allowing the option of filtering out outliers for training purposes in module 2 and 3 but including them in the subsequent downstream steps. This could work as follows: At the end of module 1 generate a list of outlier samples akin to that performed after module 3 though likely more stringent. Exclude the samples from generation of variant metrics in module 2 unless a variant is comprised of only outlier samples. In that case generate variant metrics as normal. Exclude variants comprised of only outlier samples for training purposes in module 3 but include them when using the random forest for assessment.

PR #389, opened 2022-08-18
Plot SV counts at the end of the ClusterBatch module

VJalili:
The plots will be used for removing outliers before training the random forest model. This is the first step in resolving the issue #44.

### DIFF
--- a/inputs/templates/test/ClusterBatch/ClusterBatch.json.tmpl
+++ b/inputs/templates/test/ClusterBatch/ClusterBatch.json.tmpl
@@ -29,5 +29,7 @@
   "ClusterBatch.wham_vcf_tar": {{ test_batch.std_wham_vcf_tar | tojson }},
   "ClusterBatch.manta_vcf_tar": {{ test_batch.std_manta_vcf_tar | tojson }},
   "ClusterBatch.melt_vcf_tar": {{ test_batch.std_melt_vcf_tar | tojson }},
-  "ClusterBatch.ped_file": {{ test_batch.ped_file | tojson }}
+  "ClusterBatch.ped_file": {{ test_batch.ped_file | tojson }},
+
+  "ClusterBatch.outlier_cutoff_nIQR": "10000"
 }
--- a/wdl/ClusterBatch.wdl
+++ b/wdl/ClusterBatch.wdl
@@ -291,7 +291,7 @@ workflow ClusterBatch {
   call sv_counts.PlotSVCountsPerSample {
     input:
       prefix = batch,
-      vcfs = [select_first([ClusterPESR_manta.clustered_vcf]), select_first([ClusterPESR_wham.clustered_vcf]), select_first([ClusterPESR_melt.clustered_vcf]), select_first([ClusterPESR_scramble.clustered_vcf])],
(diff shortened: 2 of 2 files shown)


wdl/ClusterBatch.wdl
```diff
@@ -282,9 +288,22 @@ workflow ClusterBatch {
     }
   }
 
+  call sv_counts.PlotSVCountsPerSample {
+    input:
+      prefix = batch,
+      vcfs = [select_first([ClusterPESR_manta.clustered_vcf]), select_first([ClusterPESR_wham.clustered_vcf]), select_first([ClusterPESR_melt.clustered_vcf]), select_first([ClusterPESR_scramble.clustered_vcf])],
```
mwalker174, line 294:
```suggestion
      vcfs = select_all([ClusterDepth.clustered_vcf, ClusterPESR_manta.clustered_vcf, ClusterPESR_wham.clustered_vcf, ClusterPESR_melt.clustered_vcf, ClusterPESR_scramble.clustered_vcf]),
```
Note to add `ClusterDepth.clustered_vcf` in there as well

inputs/templates/test/ClusterBatch/ClusterBatch.json.tmpl
```diff
@@ -29,5 +29,7 @@
   "ClusterBatch.wham_vcf_tar": {{ test_batch.std_wham_vcf_tar | tojson }},
   "ClusterBatch.manta_vcf_tar": {{ test_batch.std_manta_vcf_tar | tojson }},
   "ClusterBatch.melt_vcf_tar": {{ test_batch.std_melt_vcf_tar | tojson }},
-  "ClusterBatch.ped_file": {{ test_batch.ped_file | tojson }}
+  "ClusterBatch.ped_file": {{ test_batch.ped_file | tojson }},
+
+  "ClusterBatch.outlier_cutoff_nIQR": "10000"
```
mwalker174, line 34:
```suggestion
  "ClusterBatch.outlier_cutoff_nIQR": "8"
```
10000 is used in FilterBatch to effectively disable sample filtering. Here we should make it something statistically reasonable, otherwise the cutoffs at +/- 10000 IQR make the plots unreadable.

wdl/ClusterBatch.wdl
```diff
@@ -65,6 +66,8 @@ workflow ClusterBatch {
 
     Float? java_mem_fraction
 
+    Int outlier_cutoff_nIQR
```
mwalker174, line 69: I'd suggest making this an optional. If it's provided, we run PlotSVCountsPerSample. If not, it skips the task (we don't need to run this for the single sample pipeline for example). Would need to add this input and the outputs to GATKSVPipelinePhase1 and GATKSVPipelineBatch.

mwalker174 (commented):
I think we should try and get this in since it is useful QC at this stage of the pipeline. Can you take it further and integrate with GATKSVPipelinePhase1 and GATKSVPipelineBatch, including updating json templates?

inputs/templates/test/ClusterBatch/ClusterBatch.json.tmpl
epiercehoffman: I've been using 6 for plotting

epiercehoffman:
This was completed as part of #567 so I would recommend closing this PR

PR #395, opened 2022-08-30
Log a warning when merging EH JSON files if a file does not contain LocusResult

VJalili:
This is related to a case where EH is run on male samples without specifying `--sex male`, hence reads on chrY are ignored. And if the sample only contains reads on chrY, not setting `--sex male` results in producing an output JSON that does not have `LocusResult` (see the following), which causes errors when merging the JSON files. 

```json
{
  "SampleParameters": {
    "SampleId": "sample_id",
    "Sex": "Female"
  }
}
```

### DIFF
--- a/src/str/combine_expansion_hunter_json_to_tsv.py
+++ b/src/str/combine_expansion_hunter_json_to_tsv.py
@@ -122,7 +122,14 @@ def main():
             try:
                 json_contents = parse_json_file(json_path)
             except Exception as e:
-                logging.info(f"Skipping {json_path}... Unable to parse json: {e}")
+                logging.warning(f"Skipping {json_path}... Unable to parse json: {e}")
+                continue
+
+            if "LocusResults" not in json_contents:
+                logging.warning(f"Skipping {json_path} as it does not contain `LocusResults`. "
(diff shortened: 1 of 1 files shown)


src/str/combine_expansion_hunter_json_to_tsv.py
```diff
@@ -122,7 +122,14 @@ def main():
             try:
                 json_contents = parse_json_file(json_path)
             except Exception as e:
-                logging.info(f"Skipping {json_path}... Unable to parse json: {e}")
+                logging.warning(f"Skipping {json_path}... Unable to parse json: {e}")
+                continue
```
mwalker174, line 126: Do you need the 2 `continue`'s here?

mwalker174 (approved):

src/str/combine_expansion_hunter_json_to_tsv.py
VJalili: I guess the first one `continue`s if the json is malformed (e.g., syntax or structural errors) and the second one `continue`s if the syntax is correct but it is missing a required field.

PR #398, opened 2022-09-09
Update BenchmarkGqFilter to pass miniwdl validation

VJalili:
Miniwdl complains about how `args_str` is defined and used; the simplest solution to me was to pass the array of arguments to the task and concatenate them in the script, instead of joining at wdl level. 

I am also changing the inline python code definition to match the more commonly used style in the pipeline.


VJalili:
Closed the PR since @TedBrookings has an ongoing overhaul of this WDL.

PR #399, opened 2022-09-09
Tb main vcf qc rename samples

TedBrookings:
Add utility to optionally rename input VCF sample IDs to conform to comparison data set.

### DIFF
--- a/wdl/MainVcfQc.wdl
+++ b/wdl/MainVcfQc.wdl
@@ -16,6 +16,7 @@ workflow MainVcfQc {
     String? bcftools_preprocessing_options
     File? ped_file
     File? list_of_samples_to_include
+    File? sample_renaming_tsv # File with mapping to rename per-sample benchmark sample IDs for compatibility with cohort
     Int max_trios = 1000
     String prefix
     Int sv_per_shard
@@ -40,6 +41,7 @@ workflow MainVcfQc {
 
     # overrides for MiniTasks or Utils
     RuntimeAttr? runtime_override_subset_vcf
+    RuntimeAttr? runtime_override_rename_vcf_samples
     RuntimeAttr? runtime_override_merge_vcfwide_stat_shards
     RuntimeAttr? runtime_override_merge_vcf_2_bed
 
@@ -141,7 +143,6 @@ workflow MainVcfQc {
   if (defined(site_level_comparison_datasets)) {
     scatter ( comparison_dataset_info in select_first([site_level_comparison_datasets, 
                                                        [[], []]]) ) {
-
       # Collect site-level external benchmarking data
       call sitebench.CollectSiteLevelBenchmarking {
         input:
@@ -182,7 +183,7 @@ workflow MainVcfQc {
   scatter ( shard in SplitSamplesList.shards ) {
     call persample.CollectQcPerSample {
       input:
-        vcfs=vcfs,
+        vcfs=vcfs_for_qc,
         vcf_format_has_cn=vcf_format_has_cn,
         samples_list=shard,
         prefix=prefix,
@@ -236,14 +237,14 @@ workflow MainVcfQc {
   if (defined(sample_level_comparison_datasets)) {
     scatter ( comparison_dataset_info in select_first([sample_level_comparison_datasets, 
                                                        [[], []]]) ) {
-
       # Collect per-sample external benchmarking data
       call samplebench.CollectPerSampleBenchmarking {
         input:
           vcf_stats=MergeVcfWideStats.merged_bed_file,
           samples_list=CollectQcVcfWide.samples_list[0],
           per_sample_tarball=TarShardVidLists.vid_lists,
-          comparison_tarball=select_first([comparison_dataset_info[1]]),
+          comparison_tarball=comparison_dataset_info[1],
+          sample_renaming_tsv=sample_renaming_tsv,
           prefix=prefix,
           contigs=contigs,
           comparison_set_name=comparison_dataset_info[0],
--- a/wdl/TasksMakeCohortVcf.wdl
+++ b/wdl/TasksMakeCohortVcf.wdl
@@ -991,14 +991,14 @@ task ScatterVcf {
   command <<<
     set -euo pipefail
     # in case the file is empty create an empty shard
-    bcftools view -h ~{vcf} | bgzip -c > ~{prefix}.0.vcf.gz
-    bcftools +scatter ~{vcf} -o . -O z -p ~{prefix}. --threads ~{threads} -n ~{records_per_shard} ~{"-r " + contig}
+    bcftools view -h ~{vcf} | bgzip -c > "~{prefix}.0.vcf.gz"
+    bcftools +scatter ~{vcf} -o . -O z -p "~{prefix}". --threads ~{threads} -n ~{records_per_shard} ~{"-r " + contig}
 
-    ls ~{prefix}.*.vcf.gz | sort -k1,1V > vcfs.list
+    ls "~{prefix}".*.vcf.gz | sort -k1,1V > vcfs.list
     i=0
-    while read vcf; do
+    while read VCF; do
       shard_no=`printf %06d $i`
-      mv ${vcf} ~{prefix}.shard_${shard_no}.vcf.gz
+      mv "$VCF" "~{prefix}.shard_${shard_no}.vcf.gz"
       i=$((i+1))
     done < vcfs.list
   >>>
--- a/inputs/templates/test/MainVcfQc/MainVcfQc.per_sample_benchmark.json.tmpl
+++ b/inputs/templates/test/MainVcfQc/MainVcfQc.per_sample_benchmark.json.tmpl
@@ -7,15 +7,27 @@
     {{ reference_resources.hgsv_byrska_bishop_site_level_benchmarking_dataset | tojson }},
     {{ reference_resources.thousand_genomes_site_level_benchmarking_dataset | tojson }}
   ],
+  "MainVcfQc.sample_level_comparison_datasets": [
+    {{ reference_resources.hgsv_byrska_bishop_sample_level_benchmarking_dataset | tojson }}
+  ],
+  "MainVcfQc.sample_renaming_tsv": {{ reference_resources.hgsv_byrska_bishop_sample_renaming_tsv | tojson }},
+
   "MainVcfQc.sv_pipeline_docker": {{ dockers.sv_pipeline_docker | tojson }},
   "MainVcfQc.sv_base_mini_docker":{{ dockers.sv_base_mini_docker | tojson }},
   "MainVcfQc.sv_pipeline_qc_docker": {{ dockers.sv_pipeline_qc_docker | tojson }},
 
-  "MainVcfQc.prefix": {{ test_batch.name | tojson }},
-  "MainVcfQc.ped_file": {{ test_batch.ped_file | tojson }},
+  "MainVcfQc.prefix": "HGDP_and_HGSV",
+  "MainVcfQc.ped_file": {{ test_batch.hgdp_1kgp_ped | tojson }},
 
-  "MainVcfQc.vcfs": [ {{ test_batch.clean_vcf | tojson }} ],
-
+  "MainVcfQc.vcfs": [ {{ test_batch.gq_filtered_vcf | tojson }} ],
   "MainVcfQc.sv_per_shard": 2500,
-  "MainVcfQc.samples_per_shard": 600
+  "MainVcfQc.samples_per_shard": 600,
+  "MainVcfQc.runtime_override_per_sample_benchmark_plot": {
+    "mem_gb": 30,
+    "disk_gb": 50
+  },
+  "MainVcfQc.runtime_override_plot_qc_per_family": {
+    "mem_gb": 15,
+    "disk_gb": 100
+  }
 }
--- a/inputs/values/dockers.json
+++ b/inputs/values/dockers.json
@@ -16,7 +16,7 @@
   "sv_pipeline_docker": "us.gcr.io/broad-dsde-methods/gatk-sv/sv-pipeline:2022-09-13-v0.25.1-beta-f6d0c4da",
   "sv_pipeline_hail_docker": "us.gcr.io/broad-dsde-methods/gatk-sv/sv-pipeline:2022-09-13-v0.25.1-beta-f6d0c4da",
   "sv_pipeline_updates_docker": "us.gcr.io/broad-dsde-methods/gatk-sv/sv-pipeline:2022-09-13-v0.25.1-beta-f6d0c4da",
-  "sv_pipeline_qc_docker": "us.gcr.io/broad-dsde-methods/gatk-sv/sv-pipeline:2022-09-13-v0.25.1-beta-f6d0c4da",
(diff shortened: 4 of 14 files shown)


TedBrookings:
I added an input file `MainVcfQc_gq_filter.json.tmpl`, which I used to run QC on my input data. I've left it in for now so that a) reviewers can see how the new arguments are intended to be filled out, and b) I can re-run the job if there are any problems. I think I should probably remove it before merging. I'm not sure if the altered values should be propagated into the main `MainVcfQc.json.tmpl`

inputs/templates/test/MainVcfQc/MainVcfQc_gq_filter.json.tmpl
```diff
@@ -0,0 +1,31 @@
...
+
+  "MainVcfQc.sv_pipeline_docker": {{ dockers.sv_pipeline_docker | tojson }},
+  "MainVcfQc.sv_base_mini_docker":{{ dockers.sv_base_mini_docker | tojson }},
+  "MainVcfQc.samtools_cloud_docker":{{ dockers.samtools_cloud_docker | tojson }},
+  "MainVcfQc.sv_pipeline_qc_docker": {{ dockers.sv_pipeline_qc_docker | tojson }},
+
+  "MainVcfQc.prefix": "HGDP_and_HGSV",
+  "MainVcfQc.ped_file": "gs://broad-dsde-methods-tbrookin/results/gq-filtering/inputs/HGDP_1KGP.ped",
+
+  "MainVcfQc.vcfs": [ "gs://broad-methods-cromwell-exec-bucket-instance-8/CrossValidateGqRecalibrator/749b2884-08c3-4b73-9c1e-da32c99c5d61/call-DirectGqRecalibrator/RecalibrateGq/50849ff4-3f30-4b02-829e-eb9b34ceea61/call-RecalibrateGqTask/cacheCopy/hgdp_and_hgsv.cleaned_fixed_annotated_gq_recalibrated.vcf.gz" ],
```
epiercehoffman, line 23: I think it's cleaner to have one JSON template in general, but if this one has trios then maybe we should keep it as an alternative template to test the full breadth of functionality of MainVcfQc. Is this data able to be shared publicly? If so, can you make this VCF & PED file public? And can you move the VCF to a bucket without a TTL?

Also, can you rename the file to something like `MainVcfQc.WithTrios.json.tmpl` to be descriptive of what it contains & follow the naming convention of `WDLName.Descriptor.json.tmpl` (like GATKSVPipelineBatch.FromSampleEvidence.json.tmpl)?

inputs/templates/test/MainVcfQc/MainVcfQc_gq_filter.json.tmpl
```diff
@@ -0,0 +1,31 @@
+{
+  "MainVcfQc.primary_contigs_fai": {{ reference_resources.primary_contigs_fai | tojson }},
+
+  "MainVcfQc.site_level_comparison_datasets": [
+    {{ reference_resources.ccdg_abel_site_level_benchmarking_dataset | tojson }},
+    {{ reference_resources.gnomad_v2_collins_site_level_benchmarking_dataset | tojson }},
+    {{ reference_resources.hgsv_byrska_bishop_site_level_benchmarking_dataset | tojson }},
+    {{ reference_resources.thousand_genomes_site_level_benchmarking_dataset | tojson }}
+  ],
+  "MainVcfQc.sample_level_comparison_datasets": [
```
epiercehoffman, line 10: Can you make this reference the values JSON - similar to the site-level ones above?

wdl/Utils.wdl
```diff
@@ -758,3 +758,84 @@ task TransferVcfAnnotations {
...
+with open("original_vcf_samples.list", 'r') as f_in, open("renamed_samples.list", 'w') as f_out:
+  for line in f_in:
+    sample_id = line.rstrip()
+    renamed_id = sample_ids_map.get(sample_id, sample_id)
+    f_out.write(renamed_id + "\n")
+____EoF
+
+    echo "use bcftools reheader to change the sample names"
+    bcftools reheader \
+      -s "renamed_samples.list" \
```
epiercehoffman, line 819: It seems like it would be simpler to just take this TSV sample ID map as the input instead - what do you think? Personally I prefer the TSV because I think it's easier to generate for an analyst, and it removes some computation from this task

wdl/TasksMakeCohortVcf.wdl
```diff
@@ -991,14 +991,14 @@ task ScatterVcf {
...
-    bcftools view -h ~{vcf} | bgzip -c > ~{prefix}.0.vcf.gz
-    bcftools +scatter ~{vcf} -o . -O z -p ~{prefix}. --threads ~{threads} -n ~{records_per_shard} ~{"-r " + contig}
+    bcftools view -h ~{vcf} | bgzip -c > "~{prefix}.0.vcf.gz"
+    bcftools +scatter ~{vcf} -o . -O z -p "~{prefix}". --threads ~{threads} -n ~{records_per_shard} ~{"-r " + contig}
 
-    ls ~{prefix}.*.vcf.gz | sort -k1,1V > vcfs.list
+    ls "~{prefix}".*.vcf.gz | sort -k1,1V > vcfs.list
     i=0
-    while read vcf; do
+    while read VCF; do
```
epiercehoffman, line 999: Definitely more readable with non-overlapping variable names this way

wdl/MainVcfQc.wdl
```diff
@@ -85,7 +88,20 @@ workflow MainVcfQc {
     }
   }
 
-  Array[File] vcfs_for_qc = select_first([SubsetVcfBySamplesList.vcf_subset, vcfs])
+  if (defined(sample_renaming_json)) {
```
epiercehoffman, line 91: Do you think there's a way to do the renaming for just the external per-sample benchmarking, but not for the rest of the per-sample QC collection? That might be too complicated, because there's a lot more files, but if it was able to just map between the current cohort's sample IDs and the external cohort's sample IDs within the per-sample comparison script that seems like it would be ideal

inputs/templates/terra_workspaces/cohort_mode/workflow_configurations/MainVcfQc.json.tmpl
```diff
@@ -10,6 +10,7 @@
 
   "MainVcfQc.sv_pipeline_docker": "${workspace.sv_pipeline_docker}",
   "MainVcfQc.sv_base_mini_docker": "${workspace.sv_base_mini_docker}",
+  "MainVcfQc.samtools_cloud_docker": "${workspace.samtools_cloud_docker}",
```
epiercehoffman, line 13: Does this input also need to be added to MakeCohortVcf, the Batch WDL, and the single-sample WDL? Or maybe just use sv_pipeline_docker to avoid having to do that?

wdl/MainVcfQc.wdl
```diff
@@ -16,6 +16,7 @@ workflow MainVcfQc {
     String? bcftools_preprocessing_options
     File? ped_file
     File? list_of_samples_to_include
+    File? sample_renaming_json
```
epiercehoffman, line 19: Should make this an option in all workflows that call MainVcfQc

epiercehoffman (changes requested):
Good idea to add a sample renaming function to the sample-level comparisons here - this is a pain otherwise. 

Most of my comments are more like suggestions/ideas and may turn out to not be practical, but I wanted to bring them up at least. The change that I definitely want to request before you merge is to move the VCF in the new JSON template to a bucket without a TTL.

wdl/Utils.wdl
TedBrookings: bcftools is taking as input just a list of the new sample IDs, which it will use based on position alone (no mapping involved). The issue I faced in designing this is that we have potentially multiple input VCFs, in which case I could either have multiple input sample-renaming files, or one master input file that we could subset. I decided the first approach was actually harder to deal with, so I went with the later. Both have their downsides though. That said, I think the computation is only adding mental overhead for understanding this task, I don't think it adds meaningfully to runtime.

I actually developed a script to produce this input file, which I considered adding to the PR, but I didn't call it in a WDL. I could add it to the `scripts` folder though.

wdl/MainVcfQc.wdl
TedBrookings: Honestly I think the thing to do is to rename the samples in the external comparison set: I went with this approach because it was suggested to me and I was too busy to sit and think through the implications. The problem with only renaming during the external per-sample benchmarking, is that it takes its data from the bed files produced by `CollectQcVcfWide`. It felt like I might have to intervene in names in multiple places, and maybe mess up somewhere.

If this is a significant problem, it could be a good reason to go back to the drawing board and instead rename the files in the external tarball.

inputs/templates/terra_workspaces/cohort_mode/workflow_configurations/MainVcfQc.json.tmpl
TedBrookings: We could use anything derived from `samtools_cloud_docker`, so `sv_pipeline_docker` would work. I went with `samtools_cloud_docker` because it's smaller, but I don't think that's a huge deal. I think the best way forward would be to keep `samtools_cloud_docker` as the input for the Task in `Utils.wdl`, but then just pass `sv_pipeline_docker` as the corresponding input from `MainVcfQc.wdl`. I'll make that change unless you object.

wdl/MainVcfQc.wdl
TedBrookings: Done. I omitted Module07MinGqStep2MergePCRStatus because a) I'm suspicious of it, and b) it doesn't have the option to run per-sample comparisons anyway.

inputs/templates/terra_workspaces/cohort_mode/workflow_configurations/MainVcfQc.json.tmpl
epiercehoffman: Sounds like a good plan to me

wdl/Utils.wdl
epiercehoffman: bcftools also has the option to take a whitespace-separated file with a column for the old ID & a column for the new ID, so that's what I was referring to - sorry for the confusion since you were using a different format here. If we stick with the VCF reheader approach, my personal preference would be to provide a TSV in that format as input, rather than a JSON, because it's super easy to generate for an analyst working in excel or with Terra data tables. 

My understanding is that the multiple input VCFs should contain the same samples, but they may be sharded by genomic position, so I think it's ok to have one reheader file for all VCFs.

wdl/MainVcfQc.wdl
epiercehoffman: Do you think it could work to pass a sample renaming map to this script and use the ID that matches the external benchmarking dataset here when creating the per-sample BED files in the tarball? https://github.com/broadinstitute/gatk-sv/blob/cd5ce7e0a890c4c5234341e9fc976b4d56d3ca60/src/sv-pipeline/scripts/vcf_qc/collectQC.perSample_benchmarking.sh#L158

mwalker174 (changes requested):
Thanks @TedBrookings I thought you could just do this as a one-off for the hgdp call set, but working it into the pipeline is really above and beyond. This will save us a lot of time in the future!

For general use, I think we should be careful since some call sets contain problematic external IDs, so we should stick to the internal IDs where we can. I also need to double-check that we can share the hgdp vcf.

inputs/templates/test/MainVcfQc/MainVcfQc_gq_filter.json.tmpl
TedBrookings: I renamed it to `MainVcfQc_per_sample_benchmark.json.tmpl`, because adding the per-sample benchmarking is the main difference. However, it's also using a different VCF and PED file than the other template too. Since I've had quite variable problems with memory and disk, I think it's possible the attribute overrides relate to those inputs.

inputs/templates/test/MainVcfQc/MainVcfQc_gq_filter.json.tmpl
TedBrookings: Done

wdl/Utils.wdl
TedBrookings: Used a totally different strategy (as discussed), but the input is now a TSV. The script that produces it is `gatk-sv/scripts/inputs/get_rename_benchmark_samples_map.py`

inputs/templates/terra_workspaces/cohort_mode/workflow_configurations/MainVcfQc.json.tmpl
TedBrookings: I followed this plan. However it's now moot: the task for renaming VCF samples is not called. I'm inclined to delete it, since it is not used and doesn't use the preferred TSV input format. However, I figured I would wait for reviewer confirmation.

wdl/Utils.wdl
```diff
@@ -758,3 +758,84 @@ task TransferVcfAnnotations {
...
+  RuntimeAttr runtime_attr = select_first([runtime_attr_override, default_attr])
+
+  command <<<
+    set -euo pipefail
+
+    # symlink vcf_index to current working dir
+    ln -s ~{vcf_idx} .
+
+    echo "Get original VCF samples"
+    export GCS_OAUTH_TOKEN=`gcloud auth application-default print-access-token`
```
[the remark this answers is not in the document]
TedBrookings: This is probably moot because I expect you will want the task removed, but I want to discuss for future reference. I wrote this in mind with local execution / non-gcp stuff in mind. The token would be empty in that case, but since the files are local, their path would be good and I expect the rest of the task would execute. Is this just too risky a strategy anyway?

inputs/templates/test/MainVcfQc/MainVcfQc_gq_filter.json.tmpl
epiercehoffman: Sure, per_sample_benchmark is fine with me as an identifier. Can you still rename it to `MainVcfQc.per_sample_benchmark.json.tmpl` to match the `.` format used in other JSON templates in our repo?

wdl/MainVcfQc.wdl
```diff
@@ -295,6 +307,81 @@ workflow MainVcfQc {
 }
 
 
+# rename samples in comparison benchmark to match the cohort VCF
+task RenameBenchmarkTarfileSamples {
```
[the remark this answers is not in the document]
TedBrookings: Actually I agree, and it wasn't a tough fix. I'm running a quick test to make sure that nothing crazy happens, but it shouldn't delay anything.

inputs/templates/terra_workspaces/cohort_mode/workflow_configurations/MainVcfQc.json.tmpl
epiercehoffman: Yes, please go ahead and delete this input as well as the old task

epiercehoffman (approved):
Thanks for making the requested changes! Approved pending removal of samtools_cloud_docker input from MainVcfQc Terra JSON template & old renaming task that is no longer used

wdl/Utils.wdl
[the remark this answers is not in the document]
mwalker174: That's a fair point that it would just ignore the token in local mode, so fine to leave in. I think we had the same discussion in another PR now that I think of it.

mwalker174 (approved):

Issue #400, opened 2022-09-12
Large CNV records missing depth genotypes causing failures in single sample pipeline


cwhelan:
The single sample pipeline is failing on some samples with this error in CleanVcf:

```
Traceback (most recent call last):
  File "/opt/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py", line 181, in <module>
    main()
  File "/opt/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py", line 126, in main
    if not sample_obj['GQ'] is None and sample_obj['RD_CN'] >= 2:
TypeError: '>=' not supported between instances of 'NoneType' and 'int'
```

This is caused by CNVs greater than 5kb in size that are missing RD_CN depth genotypes.

cwhelan:
I have a fix ready for this issue but wanted to create an issue for tracking.

epiercehoffman:
Observed in one record in one cohort run with GATK-SV cohort mode as well, in a common 50kb DEL originating from Manta with PE, SR, and RD evidence.

cwhelan:
I think this was fixed in https://github.com/broadinstitute/gatk-sv/pull/401. I may have been waiting for confirmation that it solved this issue for all users before closing. We should close if @epiercehoffman is no longer seeing the issue.

epiercehoffman:
I believe this was a very rare issue. I haven't heard of it occurring since January. I would support closing this issue under the assumption it's been fixed, as the rarity of the error makes it hard to verify in practice. If it happens again we can open a new issue

Issue #336, opened 2022-04-25
tracking variants that are filtered from raw output


ekarlins:
## Feature request

Is it possible to track variants that are in the raw calls but that don't make it into the final output? It would also be nice to have a reason why they were removed. I'm also wondering if there is any way to parameterize this so you can be stricter or more lenient on what is filtered.

Thanks!

mwalker174:
Hi @ekarlins thanks for your suggestions. 

It would be very helpful to track filtered variants. We could modify `FilterBatchSites` to also output vcfs of the hard-filtered variants in a straight-forward manner.  It's a little more difficult, however, to track variants that "morph" during the 2 clustering stages (ClusterBatch and CombineBatches) which can modify precise coordinates, including CNV defragmentation, as well as some variant types that are converted to BNDs during FilterBatch. Redundant calls sharing one endpoint are also collapsed during "breakpoint overlap" filtering in ResolveComplexVariants, but dropped records there are reported in the `breakpoint_overlap_dropped_record_vcfs` output.

We also don't have any mechanism for adjusting filtering stringency during FilterBatch. Those parameters are currently hard-coded in `src/svtk/svtk/adjudicate/labelers.py` but could be exposed.

PR #401, opened 2022-09-12
Single sample remove failed depth gt records

cwhelan:
Attempt to address https://github.com/broadinstitute/gatk-sv/issues/400 although I haven't been able to test a full run on a failing sample.

The offending records would likely have been filtered out in the early stages of the joint calling pipeline which are bypassed in the single sample pipeline, so I'm modifying the single-sample pipeline specific script that converts large CNVs without depth support into BNDs to also discard records with missing depth genotypes and no PESR support.


mwalker174 (approved):
Looks good to me. Is the error message important? A more transparent option would be to dump the filtered calls to another vcf and provide that in the outputs for manual review (#336).

cwhelan:
I thought about dumping these records into another output file, but since this appears to be an edge case for which the files would be empty most of the time, and for which there's no real interpretation possible for the record anyway, it seemed to me that doing so would just clutter up output directories and terra data tables without much benefit. Happy to change it if anyone feels strongly that they should be kept, though.

PR #402, opened 2022-09-12
Improvements to the script validating WDLs using miniwdl

VJalili:
- Take imports dir as an input argument;
- Allow setting max depth of import, and set it to max 2 (compared to max 10 of miniwdl default); # - Improve displaying messages on the console;
- Add tips for debugging errors.

### DIFF
--- a/scripts/test/miniwdl_validation.py
+++ b/scripts/test/miniwdl_validation.py
@@ -13,13 +13,14 @@ import sys
 import WDL
 
 
-def format_error_message(error_type, message, pos):
-    return f"{error_type}: '{message}' at {pos.abspath}, Line: {pos.line}, Col: {pos.column}"
+def format_error_message(error_type, message, pos, tips=None):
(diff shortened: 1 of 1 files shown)


VJalili:
Thanks for the review @mwalker174. [`10` is the default value](https://github.com/chanzuckerberg/miniwdl/blob/c337ac78ca18fdf74694dbc9d85e658486930c84/WDL/__init__.py#L42) for max import depth; I am changing the default to `2` mainly to simplify debugging/tracking import errors.

For instance, with max import depth set to `10`, the validation process throws an import error validating `wdl/AnnoRdPeSr.wdl` complaining `'Failed to import VaPoR.wdl'`, which implies the script cannot find `VaPoR.wdl` in the import path; while the root cause is in `wdl/TasksMakeCohortVcf.wdl` with the error `Expected Boolean instead of Boolean?`.

PR #404, opened 2022-09-13
Update README.md

shadizaheri:
to remove delly from the list of callers.


epiercehoffman (approved):
Thanks for catching this

shadizaheri:
Thanks for the prompt review

PR #407, opened 2022-09-15
Improve EH JSON to TSV conversion when the JSON is missing LocusResults

VJalili:
There are a few cases where ExpansionHunter (EH) may generate a JSON file that is missing the `LocusResults` key (see the following JSON file as an example). In general, this can happen when EH cannot process _any of the sites_ in the provided variant catalog, and one possible case is when the site is defined on a chromosome which the sample does not contain (e.g., asking for sites on `chrY` while studying female samples).

```json
{
  "SampleParameters": {
    "SampleId": "NA19238.final",
    "Sex": "Female"
  }
}
```

EH logs every site that it cannot process as the following, and if it fails on _all the sites_ specified in the variant catalog, then its generated output JSON file will be missing `LocusResults`. 

> Error on locus spec chrY-1377068-1377082-AC: Error loading locus chrY-1377068-1377082-AC: Flanks can contain at most 5 characters N but found 2000 Ns

The lack of `LocusResults` key causes a few corner case errors with the script used to convert JSON to TSV. This PR patches the script to only log a warning for such corner cases instead of the failing. 

PS: a minor correction on checking the value of an integer variable is also PR'd.

### DIFF
--- a/src/str/combine_expansion_hunter_json_to_tsv.py
+++ b/src/str/combine_expansion_hunter_json_to_tsv.py
@@ -106,7 +106,7 @@ def main():
     if args.sample_metadata:
         sample_metadata_df = pd.read_table(args.sample_metadata)
         sample_id_column_idx = get_sample_id_column_index(sample_metadata_df, column_name=args.sample_metadata_key)
-        if sample_id_column_idx is -1:
+        if sample_id_column_idx == -1:
             raise ValueError(f"'sample_id' column not found in sample metadata table. The columns found were: {sample_metadata_df.columns}")
         sample_id_column = sample_metadata_df.columns[sample_id_column_idx]
         for _, row in sample_metadata_df.iterrows():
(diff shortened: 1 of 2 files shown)


src/str/combine_expansion_hunter_json_to_tsv.py
```diff
@@ -14,6 +14,17 @@
...
 ALREADY_WARNED_ABOUT = set()  # used for logging
 
+MISSING_LOCUS_RESULTS_WARNING_MESSAGE = \
+    "Skipping {json_path} as it does not contain `LocusResults`. " \
+    "One possible cause is running ExpansionHunter on male samples with default " \
+    "value of the `--sex` argument, which defaults to female. " \
+    "Rerunning EH with a correct value of `--sex` argument may fix this issue. " \
+    "Another issue could be running ExpansionHunter on female samples with a variant " \
+    "catalog that only contains sites on ChrY; or in more general sense, running " \
+    "ExpansionHunter on a sample that contains no reads on none of the chromosomes " \
```
mwalker174, line 24:
```suggestion
    "ExpansionHunter on a sample that contains no reads on any of the chromosomes " \
```

mwalker174 (approved):
Looks good, just one minor suggestion.

VJalili:
Thanks @mwalker174!

PR #408, opened 2022-09-15
Remove extra referencing samtools docker image

VJalili:
Fixing a double-referencing of samtools docker image; identified using miniwdl validation.


TedBrookings (approved):
Good catch, surprised it didn't cause problems or get caught by womtool!

VJalili:
I am surprised too how womtool did not catch this. Thanks for the quick review!

PR #410, opened 2022-09-19
Use select_first when setting required vars using optional vars.

VJalili:
A couple of improvements on setting the value of required variables using optional variables in order to adhere with `miniwdl` requirements.


mwalker174 (approved):

VJalili:
Thank you @mwalker174!

PR #411, opened 2022-09-19
Fix overlapping var naming and duplicate references

VJalili:


mwalker174 (approved):
Thanks for fixing these!

VJalili:
Thanks for the review @mwalker174!

PR #416, opened 2022-09-23
an alpha version of a report generator that creats a report of the re…

shadizaheri:
This PR is about an alpha version of a report generator that I coded to generate a report using the plots generated in "*.pdf, *.jpg, and *.png" formats in a single doc format. 

The way that it works is explained below:
1- Download the zipped folder of the results you want to include in the plots, unzip, and un-tar the folder. Example: let's say using the provided gsutil from your specific workflow you will get the folder named as bwa_samples_aligning_with_dragen_SV_VCF_QC_output.wPlots.tar.gz
you want to go to bwa_samples_aligning_with_dragen_SV_VCF_QC_output folder.

2- to generate the word-doc at the first step, just copy the "alpha_**first**_time_report_generator.py" script to the folder containing the plots.
Following the same example: copy the "alpha_**first**_time_report_generator.py" script in the bwa_samples_aligning_with_dragen_SV_VCF_QC_output/plots/main_plot directory.

3- if you desire to add more plots to the existing document file, please run the "alpha_**second**_time_report_generator.py" script and import the name of the desired doc file that exists in the same directory as the plots, "alpha_second_time_report_generator.py" script, and the existing .doc file. 
Example: 
Let's say that in step 2 you generated a doc being called "bwa_report.doc". So copy the "bwa_report.doc" as well as the  "alpha_**second**_time_report_generator.py" script to the next directory that you want to include the plots in and then run the "alpha_**second**_time_report_generator.py"  script.

Future improvements that I'm working on:
1. Make it possible to use it on a Terra notebook (no need to download the files)
2. Provide only the gsutil of the results one wants to include (no need to manually unzip or untar the folder)
3. Merge the "alpha_**first**_time_report_generator.py" and "alpha_**second**_time_report_generator.py" python files for the first and second generators into one (with an if statement to check if the doc file exists, and if it does then it'll warn the user and vice versa)

### DIFF
--- a/scripts/qc/alpha_second_time_report_generator.py
+++ b/scripts/qc/alpha_second_time_report_generator.py
@@ -1,0 +1,89 @@
+import docx
+from docx.shared import Inches
+import os
+import glob
+from pdf2image import convert_from_path, convert_from_bytes
+from pdf2image.exceptions import (
+    PDFInfoNotInstalledError,
+    PDFPageCountError,
+    PDFSyntaxError
+)
+# Get the current working directory
+cwd = os.getcwd()
+output_file = input("This script assumes that there is already a .docx created \n if this is the first time runing this script please first run the alpha_first_time_report_generator.py \n Enter the name of the already existing output file: ")
+#output_file = "CleftPalete_WGS_GATK-SV"
+# Print the current working directory
+print("Current working directory: {0}".format(cwd))
+
+#path_to_files = r'/Users/szaheri/Documents/SV/GATK-SV/github_ryan_shared/bwa_dragen_sample_level_comparison/results/bwa/perFamily_plots_tarball/bwa_samples_aligning_with_dragen_perFamily_plots/supporting_plots/sv_inheritance_plots/'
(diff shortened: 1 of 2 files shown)


scripts/qc/alpha_second_time_report_generator.py
```diff
@@ -0,0 +1,89 @@
...
+import glob
+from pdf2image import convert_from_path, convert_from_bytes
+from pdf2image.exceptions import (
+    PDFInfoNotInstalledError,
+    PDFPageCountError,
+    PDFSyntaxError
+)
+# Get the current working directory
+cwd = os.getcwd()
+output_file = input("This script assumes that there is already a .docx created \n if this is the first time runing this script please first run the alpha_first_time_report_generator.py \n Enter the name of the already existing output file: ")
```
shadizaheri, line 13: output_file = input("This script assumes that there is already a .docx created \n if this is the first time runing this script please first run the alpha_first_time_report_generator.py \n Enter the name of the already existing output file (do not include the .docx part): ")

PR #423, opened 2022-09-27
Add VisualizeCnvs

mwalker174:
Adds a new WDL that generate depth plots for DEL/DUP variants from a VCF file using RdTest. Depth profiles across batches are plotted, and a min size cutoff parameter is included. Tested on the 1KGP reference panel with 50 kbp cutoff.

### DIFF
--- a/wdl/VisualizeCnvs.wdl
+++ b/wdl/VisualizeCnvs.wdl
@@ -6,8 +6,8 @@ import "Structs.wdl"
 
 workflow VisualizeCnvs {
   input{
-    File? vcf  # Option 1
-    File? bed  # Option 2, columns: chrom,start,end,name,svtype,samples
+    # Note vcf will be faster
+    File vcf_or_bed  # bed columns: chrom,start,end,name,svtype,samples
 
     String prefix
     Array[File] median_files
@@ -25,8 +25,7 @@ workflow VisualizeCnvs {
 
   call RdTestPlot {
     input:
-      vcf=vcf,
-      bed=bed,
+      vcf_or_bed=vcf_or_bed,
       median_files=median_files,
       ped_file=ped_file,
       rd_files=rd_files,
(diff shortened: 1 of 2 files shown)


wdl/VisualizeCnvs.wdl
```diff
@@ -0,0 +1,139 @@
...
+    File ped_file
+    String prefix
+    String sv_pipeline_docker
+    String flags
+    RuntimeAttr? runtime_attr_override
+  }
+  RuntimeAttr default_attr = object {
+                               cpu_cores: 1,
+                               mem_gb: 7.5,
+                               disk_gb: ceil(100 + size(vcf, "GB") * 10 + size(rd_files, "GB")),
```
VJalili, line 62: Include also `bed`?

wdl/VisualizeCnvs.wdl
```diff
@@ -0,0 +1,139 @@
+version 1.0
+
+import "Structs.wdl"
+
+# Plots CNV depth profiles across batches
+
+workflow VisualizeCnvs {
+  input{
+    File? vcf  # Option 1
+    File? bed  # Option 2, columns: chrom,start,end,name,svtype,samples
```
VJalili, line 10: Can both be given? if not, would not using one variable/arg and deciding based on the extension of the provided value whether is bed (compressed or decompressed) or VCF be simpler interface? 

On a similar note, since you are supporting zipped and unzipped `bed` would it be useful to support compressed and raw `vcf` as well?

VJalili (commented):
LGTM, thanks @mwalker174! I added a couple minor suggestions,.

wdl/VisualizeCnvs.wdl
mwalker174: Good catch

wdl/VisualizeCnvs.wdl
mwalker174: That's a good idea - I've implemented it as a single variable. Decompressed vcfs are very uncommon so I won't include support for those though.

wdl/VisualizeCnvs.wdl
VJalili: Thanks for correcting that!

VJalili (approved):
LGTM, thank you for addressing my comments.

PR #425, opened 2022-09-28
Redundancy update

xuefzhao:
added scripts to re-cluster the SVs of specific variant type, genomic context and size ranges, by requiring 10% sample overlap


mwalker174:
Closing in favor of #446

PR #428, opened 2022-09-30
Add support for the seeking option

VJalili:
EH has two options for reading a cram: seeking, streaming. I suppose the former is advantages when access a few sites (i.e., the variant catalog is covering only a small portion of the cram) and the later is a linear scan on the cram which can be advantageous when the variant catalog is covering a considerable portion of the cram. 

We are still experimenting with how these options can improve cost-efficiency of running EH in our application; regardless having this option in the WDL can improve its application for wider use-cases.


mwalker174 (approved):
👍

PR #429, opened 2022-10-03
Install git lfs and add tracking png and jpg files

VJalili:
Note that the link the template will be valid after the PR is merged, and should update to refer to a specific tag if not want to use the current latest.


VJalili:
Thank you @mwalker174!

PR #431, opened 2022-10-04
Fix a bug in generating realigned BAM in the output of EH

VJalili:
- [x] Fix a condition that was reversely applied;
- [x] Refactor from `overlapping_reads` to `realigned_bam` for clarity.


mwalker174 (approved):
👍

PR #433, opened 2022-10-04
Reformat BED file for VaPoR input

epiercehoffman:
### Updates
Based on Xuefang's branch (https://github.com/broadinstitute/gatk-sv/blob/xz_fixes_3/module10_benchmark/TasksBenchmark.wdl#L1478), when providing a BED file generated from svtk vcf2bed, make the following modifications to the BED file before running VaPoR:
* Split by contig
* Split by sample if sample ID provided and BED contains sample column
* Remove BND, CPX, and CNV events
* For INS events (including MEI/SVA/ALU), put INS_{SVLEN} in the SVTYPE/description column

The motivation is to allow users to provide the cohort-wide BED file from svtk vcf2bed from MainVcfQc to VaPoR without extra manual modifications.

This task should be able to handle BED files that are bgzipped or uncompressed, single-sample or multi-sample, and already modified with SVLEN for INS or not (as long as there is a header with an SVLEN column). I rewrote it in python for readability after it became unruly in bash due to the many alternate code paths.

### Testing
* Ran with manually created single-sample VaPoR-ready BED files to just split by contig and strip out BND/CNV/CPX. Confirmed the results were identical other than small change due to random sampling of high-coverage areas
* Ran with BED file from svtk vcf2bed and confirmed that INS events were properly formatted and evaluated by VaPoR


mwalker174 (changes requested):
Thanks for working all this out - this is going to be a critical piece of code going forward so I'm going to ask for some changes:
- I'd prefer to have this in a standalone script built into the docker to assist with debugging and allow for offline use.
- Rename `SplitBed` to `PreprocessBed` and move it out of the generic `TasksBenchmark.wdl` lib and into the vapor wdl, since it's really vapor-specific (there is also another `SplitBed` or two running around in other WDLs)
-

epiercehoffman:
Tested WDL & reran womtool validation with latest updates, including adding the preprocessing script to the docker and switching to the primary contigs list file instead of an array of contigs input. 

Note that I tested the WDL in Terra as the file in the VaPoR JSON template does not have the right permissions to use for testing. The template itself is correctly formatted, but I'm not sure if there is more widely available data that can be used for testing here.

PR #436, opened 2022-10-05
Add workflow to apply bcftools hard variant filter

epiercehoffman:
### Updates
* Add WDL to apply hard filter on Wham-only SR-only DELs <1kb
* Edited to make filter query customizable

### Testing 
* Validated with womtool
* Tested in Terra on AoU data
* Tested in cromshell on ref panel test data

### DIFF
--- a/wdl/ApplyHardFilter.wdl
+++ b/wdl/ApplyHardFilter.wdl
@@ -1,0 +1,81 @@
+version 1.0
+
+import "Structs.wdl"
+
+workflow ApplyHardFilter {
+  input {
+    String prefix
+    File vcf
+    String? filter_name
+
+    String sv_base_mini_docker
+    RuntimeAttr? runtime_attr_hard_filter_wham
+  }
+
+  call HardFilterWham {
+    input:
+      prefix = prefix,
+      vcf = vcf,
+      filter_name = filter_name,
+      sv_base_mini_docker = sv_base_mini_docker,
+      runtime_attr_override = runtime_attr_hard_filter_wham
+  }
+
+  output {
+    File hard_filtered_vcf = HardFilterWham.hard_filtered_vcf
+    File hard_filtered_vcf_index = HardFilterWham.hard_filtered_vcf_index
+  }
+}
+
+
+# Removes Wham-only, SR-only deletions <1kb
+task HardFilterWham {
+  input {
+    String prefix
+    File vcf
+    String? filter_name
+
+    String sv_base_mini_docker
+    RuntimeAttr? runtime_attr_hard_filter_wham
+  }
+
+  String filter_name_ = select_first([filter_name, "filter_wham_del_sr_1kb"])
+  String hard_filtered_vcf_name = "~{prefix}.~{filter_name_}.vcf.gz"
+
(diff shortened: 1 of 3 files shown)


wdl/ApplyHardFilter.wdl
```diff
@@ -0,0 +1,81 @@
...
+    max_retries: 1,
+    boot_disk_gb: 10
+  }
+  RuntimeAttr runtime_attr = select_first([runtime_attr_override, default_attr])
+
+  command <<<
+
+    set -euo pipefail
+
+    bcftools view -e 'SVTYPE=="DEL" && COUNT(ALGORITHMS)==1 && ALGORITHMS=="wham" && SVLEN<1000 && COUNT(EVIDENCE)==1 && EVIDENCE=="SR"' ~{vcf} -Oz -o "~{hard_filtered_vcf_name}"
```
mwalker174, line 61: Let's make this workflow reusable and let the user provide the filtering criteria as a String:
```suggestion
    bcftools view -e '~{filter}' ~{vcf} -Oz -o "~{hard_filtered_vcf_name}"
```

wdl/ApplyHardFilter.wdl
```diff
@@ -0,0 +1,81 @@
...
+
+    tabix -f -p vcf "~{hard_filtered_vcf_name}"
```
mwalker174, line 63:
```suggestion
    tabix "~{hard_filtered_vcf_name}"
```

wdl/ApplyHardFilter.wdl
```diff
@@ -0,0 +1,81 @@
...
+  input {
+    String prefix
+    File vcf
+    String? filter_name
+
+    String sv_base_mini_docker
+    RuntimeAttr? runtime_attr_hard_filter_wham
+  }
+
+  call HardFilterWham {
```
mwalker174, line 15:
```suggestion
  call HardFilterVcf {
```

mwalker174 (changes requested):
Thanks this is looking good. I think we should generalize this to allow custom filter queries as well and probably consider renaming the workflow, see comments below.

mwalker174:
Can you also include an input json with this? Use the cleaned vcf and supply the wham filter as the query.

wdl/ApplyManualVariantFilter.wdl
```diff
@@ -0,0 +1,85 @@
...
+    max_retries: 1,
+    boot_disk_gb: 10
+  }
+  RuntimeAttr runtime_attr = select_first([runtime_attr_override, default_attr])
+
+  command <<<
+
+    set -euo pipefail
+
+    bcftools view -e "~{bcftools_filter}" ~{vcf} -Oz -o "~{hard_filtered_vcf_name}"
```
mwalker174, line 65:
```suggestion
    bcftools view -e '~{bcftools_filter}' ~{vcf} -Oz -o "~{hard_filtered_vcf_name}"
```
Should use single quotes so that special characters such as `@` and `*` won't be interpreted by the shell. I believe then you also have to modify the filter expression in the input json to use double quotes for strings, escaped with a backslash i.e. `"SVTYPE==\"DEL\" …"`.

PR #437, opened 2022-10-05
Add wrapper workflow to call SubsetVcfBySamplesList task

epiercehoffman:
### Updates
Add wrapper WDL to call SubsetVcfBySamplesList task to be able to run it as a standalone workflow

### Testing
* Validated with womtool
* Tested in Terra on AoU data
* Tested in cromshell on ref panel test data


mwalker174 (changes requested):
Thanks, can you include a json template for this? Just use the ref panel cleaned vcf and make up a sample subset list and it in `gs://gatk-sv-ref-panel-1kg/test_inputs/`

epiercehoffman:
> Thanks, can you include a json template for this? Just use the ref panel cleaned vcf and make up a sample subset list and it in `gs://gatk-sv-ref-panel-1kg/test_inputs/`

I didn't see that subdirectory so I put it in `gs://gatk-sv-resources-public/hg38/v0/sv-resources/ref-panel/1KG/v1/`, but I can create the `test_inputs` subdirectory and move it if that's preferable

mwalker174:
@epiercehoffman  Yes you can just create the directory. We should keep the test data, which get frequently, in this separate bucket and reserve the resource bucket for permanent files.

epiercehoffman:
Updated to make SubsetVcfBySamplesList task flexible. Toggle remove_samples to either keep or remove the samples in the provided list, and remove_private_sites to set or not set --min-ac 1. 
* Accordingly update WDLs that use the task
* Use this task instead of ExcludeOutliers to remove outlier samples in FilterBatchSamples and FilterOutlierSamples

Testing:
* Tested SubsetVcfBySamplesList workflow with all 4 combinations of remove_samples and remove_private_sites settings and confirmed that the output VCF had a corresponding number of records and samples

mwalker174 (approved):
This is great! Good to merge I think.

PR #441, opened 2022-10-07
Create an empty timing file in case it is not created in EH

VJalili:
Creating an empty timing file in case EH fails to create this. This could be related to the cases where EH fails with errors such as the following and does not create the `sample_id_timing.tsv` file, while the delocalization script is trying to delocalize the file. 

> Error on locus spec chrY-57210762-57210771-AGA: Error loading locus chrY-57210762-57210771-AGA: Flanks can contain at most 5 characters N but found 2000 Ns

### DIFF
--- a/wdl/ExpansionHunter.wdl
+++ b/wdl/ExpansionHunter.wdl
@@ -144,6 +144,7 @@ task RunExpansionHunter {
             fi
         fi
 
+        touch ~{sample_id}_timing.tsv
         ExpansionHunter \
             --reads ~{bam_or_cram} \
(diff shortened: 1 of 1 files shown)


mwalker174:
Can you clarify why you need to have an empty timing file is there's an error occurring? If the tool fails then the job should fail anyway.

VJalili:
We are using a modified version of EH which seems to fail to generate the timing file when `--analysis-mode streaming` .

mwalker174 (approved):
Ok makes sense!

PR #442, opened 2022-10-07
Hook up WGD run attribute overrides

mwalker174:
Sets attribute overrides for the WGD workflow.


VJalili (approved):

VJalili:
LGTM, thanks for fixing this!

PR #444, opened 2022-10-11
Change the default max depth of inheritance & update help

VJalili:
Miniwdl implements max depth of import as how nested WDLs can be, and errors out if the WDL at the given max depth is not a leaf node in the inheritance tree. [Miniwdl's default for this parameter is `10`](https://github.com/chanzuckerberg/miniwdl/blob/06088e195a382684e5c6b1455d1753880de30127/WDL/__init__.py#L42), and I am increasing it to `15` to allow deeper nested imports.


mwalker174 (approved):
👍

VJalili:
Thanks @mwalker174!

PR #445, opened 2022-10-11
Fix referencing HailMerge task instead of the workflow.

VJalili:
In https://github.com/broadinstitute/gatk-sv/pull/440, some references to the `HailMerge` workflow were incorrectly replaced by a reference to the `HailMergeTask` task; this refactoring issue is fixed in this PR.


mwalker174 (approved):
Thank you @VJalili!

PR #449, opened 2022-10-14
Replace EH version with its current latest release 

VJalili:
Updates the ExpansionHunter (EH) workflow to use Illumina's current latest release, replacing an in-house built version. Specifically: 

- [x] Update the Dockerfile;
- [x] Add a `thread_count` workflow option since EH's current latest release is multi-threaded;
- [x] Remove `timings` since it the output of the in-house built;
- [x] Export all the JSON outputs of EH in addition to converting them to TSV and combining the output of multiple shards;
- [x] Change the default disk type from `HDD` to `SSD` since the process is I/O-bound. In a follow up PR, we can extend the `RuntimeAttr` so it takes the type of disk and applies accordingly.

### DIFF
--- a/wdl/ExpansionHunter.wdl
+++ b/wdl/ExpansionHunter.wdl
@@ -24,6 +24,7 @@ workflow ExpansionHunter {
         Boolean? generate_realigned_bam
         Boolean? generate_vcf
         Boolean? seeking_analysis_mode
+        Int? thread_count
         File? ped_file
         String expansion_hunter_docker
         String python_docker
@@ -47,6 +48,7 @@ workflow ExpansionHunter {
         reference_fasta_index,
         reference_fasta + ".fai"])
 
+    Int thread_count_ = select_first([thread_count, 2])
     Boolean generate_realigned_bam_ = select_first([generate_realigned_bam, false])
     Boolean generate_vcf_ = select_first([generate_vcf, false])
     Boolean seeking_analysis_mode_ = select_first([seeking_analysis_mode, true])
@@ -68,6 +70,7 @@ workflow ExpansionHunter {
                 generate_realigned_bam = generate_realigned_bam_,
                 generate_vcf = generate_vcf_,
                 analysis_mode = analysis_mode,
+                thread_count = thread_count_,
                 ped_file = ped_file,
                 expansion_hunter_docker = expansion_hunter_docker,
                 runtime_override = runtime_eh
@@ -80,7 +83,6 @@ workflow ExpansionHunter {
             variants_tsvs = RunExpansionHunter.variants_tsv,
             alleles_tsvs = RunExpansionHunter.alleles_tsv,
             realigned_bams = RunExpansionHunter.realigned_bam,
-            timings = RunExpansionHunter.timing,
             generate_realigned_bam = generate_realigned_bam_,
             generate_vcf = generate_vcf_,
             output_prefix = sample_id,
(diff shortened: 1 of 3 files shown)


wdl/ExpansionHunterScatter.wdl
```diff
@@ -56,20 +57,21 @@ workflow ExpansionHunterScatter {
...
-                sample_id=sample_id,
-                ped_file=ped_file,
-                generate_realigned_bam=generate_realigned_bam,
-                generate_vcf=generate_vcf,
-                seeking_analysis_mode=seeking_analysis_mode,
-                expansion_hunter_docker=expansion_hunter_docker,
-                python_docker=python_docker,
-                runtime_eh=runtime_eh,
-                runtime_concat=runtime_concat
+                bam_or_cram = bam_or_cram_,
```
cwhelan, line 60: I guess we have WDLs that format call parameter assignments differently scattered through the codebase (with varying amounts of whitespace around the =). I think no spaces might be the most common but I'm not sure about that. Fine to change it here for now if you prefer this style but we should probably eventually make a call on preferred style and format the whole codebase the same way. (this is just a note, not asking you to revert this change)

wdl/ExpansionHunter.wdl
```diff
@@ -178,19 +181,20 @@ task RunExpansionHunter {
...
+                reference_fasta_index], "GiB")))
     }
     RuntimeAttr runtime_attr = select_first([runtime_override, runtime_default])
 
     runtime {
         docker: expansion_hunter_docker
         cpu: select_first([runtime_attr.cpu_cores, runtime_default.cpu_cores])
         memory: select_first([runtime_attr.mem_gb, runtime_default.mem_gb]) + " GiB"
-        disks: "local-disk " + select_first([runtime_attr.disk_gb, runtime_default.disk_gb])  + " HDD"
+        disks: "local-disk " + select_first([runtime_attr.disk_gb, runtime_default.disk_gb])  + " SSD"
```
cwhelan, line 197: Agree with you that it would be good to make SSD vs HDD a part of RuntimeAttr so we can experiment with making this switch in other workflows, but that can be a separate PR.

cwhelan (approved):
This looks fine to me. Added a few notes that are more comments than requested changes. One more strikes me: if you've settled on one preferred analysis mode in the "seeking" vs "streaming" bake-off, you might want to consider removing it as an option in the workflow (especially if it won't work with the other default parameters you've settled on). But if you still want to leave open the possibility of experimentation feel free to leave it in there.

VJalili:
Thanks for reviewing this @cwhelan. `seeking` mode is advantageous when running on a small variant catalog, and `streaming` mode is faster when running on a larger catalog. So, I guess they are both useful for different use-cases.

cwhelan:
OK, sounds good to leave it in there!

wdl/ExpansionHunter.wdl
```diff
@@ -170,6 +171,8 @@ task RunExpansionHunter {
         python /opt/str/combine_expansion_hunter_json_to_tsv.py -o ~{sample_id} ~{sample_id}.json
         mv ~{sample_id}.*_json_files_alleles.tsv ~{sample_id}_alleles.tsv
         mv ~{sample_id}.*_json_files_variants.tsv ~{sample_id}_variants.tsv
+
+        bgzip ~{sample_id}.json
```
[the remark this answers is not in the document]
VJalili: Thanks! I changed it accordingly.

VJalili:
Thanks for the review @cwhelan!

PR #450, opened 2022-10-25
Add miniwdl validation to CI/CD and improve the validation results messages

VJalili:
- [x] Add a Github action that runs the miniwdl validation script if the PR is made or a commit is pushed to the `master`; the action will fail if the script fails, and succeeds if otherwise;
- [x] Change the color of the wdl filename message to default (previously yellow), and print `Fail` or `Pass` in front of the filename, and print the validation errors in yellow. These changes make it easier to read the logs, in particular on the Github actions.


mwalker174 (approved):
Thank you @VJalili - I tested and agreed it looks much better with the coloring. Glad to see everything is passing now!

cwhelan (approved):
This looks good to me, too!

VJalili:
Thank you @cwhelan and @mwalker174 for reviewing this!

PR #451, opened 2022-10-27
Make EvidenceQC table

VJalili:


wdl/EvidenceQC.wdl
```diff
@@ -169,5 +194,66 @@ workflow EvidenceQC {
...
+
+  command <<<
+    set -euxo pipefail
+
+    PLOIDY=""
+    if ~{defined(ploidy_plots)}; then
+      PLOIDY="--ploidy-plots ~{ploidy_plots}"
+
+    python sv-pipeline/scripts/make_evidence_qc_table.py \
+      $PLOIDY
```
epiercehoffman, line 236: You can do the argument setup for optional inputs in one line with `~{"--ploidy-plots " + ploidy_plots}`

wdl/EvidenceQC.wdl
```diff
@@ -169,5 +194,66 @@ workflow EvidenceQC {
...
+    File? wham_qc_high
+    File? scramble_qc_low
+    File? scramble_qc_high
+
+    String sv_pipeline_docker
+    RuntimeAttr? runtime_override
+  }
+
+  output {
+    File qc_table_csv = "${prefix}_table.csv"
```
epiercehoffman, line 225: TSV please

wdl/EvidenceQC.wdl
```diff
@@ -169,5 +194,66 @@ workflow EvidenceQC {
...
+  }
+
+  command <<<
+    set -euxo pipefail
+
+    PLOIDY=""
+    if ~{defined(ploidy_plots)}; then
+      PLOIDY="--ploidy-plots ~{ploidy_plots}"
```
epiercehoffman, line 233: I'd probably un-tar this tarball here outside of the python script

wdl/EvidenceQC.wdl
```diff
@@ -169,5 +194,66 @@ workflow EvidenceQC {
...
+      $PLOIDY
+  >>>
+
+  RuntimeAttr runtime_default = object {
+    cpu_cores: 1,
+    mem_gb: 4,
+    boot_disk_gb: 10,
+    preemptible_tries: 3,
+    max_retries: 1,
+    disk_gb: 100
```
epiercehoffman, line 245: Might make sense to scale disk based on inputs - although this may be hard to do for the tarball

wdl/EvidenceQC.wdl
epiercehoffman: Actually this table is going to be big, let's gzip it

VJalili: How big do you think this can be? @shadizaheri is currently using pandas dataframe for simplicity keeps the table in memory, if this will be big, probably we should use a different approach that does not keep the data in the memory.

epiercehoffman: At a batch level, I think it's fine to use pandas. Eventually on the cohort level (not within this workflow) we might run into issues with very large datasets.

epiercehoffman: I don't have a size estimate, so you could see how big it gets for 156 ref panel samples and then x5 for a larger batch

VJalili: I think @shadizaheri has a memory usage estimate for this.

wdl/EvidenceQC.wdl
```diff
@@ -46,6 +46,8 @@ workflow EvidenceQC {
 
     Boolean run_ploidy = true
 
+    Float? melt_insert_size
```
epiercehoffman, line 49: This would be an Array[Float?]? because it would be one insert size value for each sample

PR #452, opened 2022-10-31
GATK vcf formatting script fixes

mwalker174:
Fixes an issue where the `EV` FORMAT field was incorrectly removed from the header. Also adds `END2` tag propagation for CPX/CTX types.

### DIFF
--- a/src/sv-pipeline/scripts/format_svtk_vcf_for_gatk.py
+++ b/src/sv-pipeline/scripts/format_svtk_vcf_for_gatk.py
@@ -32,7 +32,7 @@ def _parse_bnd_ends(vcf_path: Text) -> Dict[Text, int]:
             columns = line.split('\t', 8)
             vid = columns[2]
             info = columns[7]
-            if 'SVTYPE=BND' not in info:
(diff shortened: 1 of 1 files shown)


src/sv-pipeline/scripts/format_svtk_vcf_for_gatk.py
```diff
@@ -32,7 +32,7 @@ def _parse_bnd_ends(vcf_path: Text) -> Dict[Text, int]:
             columns = line.split('\t', 8)
             vid = columns[2]
             info = columns[7]
-            if 'SVTYPE=BND' not in info:
+            if 'SVTYPE=BND' not in info and 'SVTYPE=CTX' not in info and 'SVTYPE=CPX' not in info:
```
TedBrookings, line 35: Totally fine with three things. If there's some possibility later on of having a lot more things that will be treated like BNDs / have END2 tags, you may want to consider storing these svtype declarations in a tuple and doing something like:
`if not any(svtype_declaration in info for svtype_declaration in svtype_declarations):`

TedBrookings (approved):
Sorry it took me awhile to get to this, I got a huge amount of email the day you requested and I just noticed it as I was clearing out my backlog.

PR #456, opened 2022-11-09
Extend the `testwdls` workflow to test WDLs using WOMtool

VJalili:
This PR extends the `testwdls` workflow to test WDLs using [`WOMtool`](https://cromwell.readthedocs.io/en/stable/WOMtool/) in addition to [`miniwdl`](https://github.com/chanzuckerberg/miniwdl). WOMtool and miniwdl tests are both run as different steps of one job that checks WDLs for their syntax issues. As per DM with @mwalker174, we decided to use `scripts/test/validate.sh` instead of directly calling `womtool`, mainly because this script performs a few convenience setup and can be extended for more comprehensive syntax tests, independent from the `testwdls` workflow.


mwalker174 (approved):
Thank you @VJalili also nice that this will also test the inputs build script as well.

VJalili:
Thanks, @mwalker174!  

> also nice that this will also test the inputs build script as well.

Agree!

PR #459, opened 2022-11-21
A py script to combine multiple outputs of Evidence QC in one table

shadizaheri:
This is the python script created to add some of the metrics resulting from the EvidenceQC step to the out put data table.

### DIFF
--- a/src/sv-pipeline/scripts/make_evidence_qc_table.py
+++ b/src/sv-pipeline/scripts/make_evidence_qc_table.py
@@ -6,7 +6,6 @@ This script creates a single table of sample QC & batching metrics from Evidence
 import argparse
 import pandas as pd
 from collections import Counter
-import numpy as np
 from numpy import array
 from functools import reduce
 
--- a/src/sv-pipeline/scripts/make_evidence_qc_table.py
+++ b/src/sv-pipeline/scripts/make_evidence_qc_table.py
@@ -90,9 +90,9 @@ def read_manta_outlier(filename: str) -> pd.DataFrame:
     if df_manta['Outlier_Sample'].empty:
         df_manta['#ID'] = df_manta.apply(lambda _: ' ', axis=0)
         outlier_manta_sample = df_manta.pivot_table(columns=['#ID'], aggfunc='size').astype(int)
-        status = 'DataFrame is empty'
+        #status = 'DataFrame is empty'
     else:
-        status = 'DataFrame is not empty'
+        #status = 'DataFrame is not empty'
         df_manta['#ID'] = df_manta['Outlier_Sample'].str.split('/', expand=True)[10].str.split('.', expand=True)[0]
         outlier_manta_sample = df_manta.pivot_table(columns=['#ID'], aggfunc='size').astype(int)
     outlier_manta_df = outlier_manta_sample.reset_index()
@@ -111,9 +111,9 @@ def read_melt_outlier(filename: str) -> pd.DataFrame:
     if df_melt['Outlier_Sample'].empty:
         df_melt['#ID'] = df_melt.apply(lambda _: ' ', axis=0)
         outlier_melt_sample = df_melt.pivot_table(columns=['#ID'], aggfunc='size').astype(int)
-        status = 'DataFrame is empty'
+        #status = 'DataFrame is empty'
     else:
-        status = 'DataFrame is not empty'
+        #status = 'DataFrame is not empty'
         df_melt['#ID'] = df_melt['Outlier_Sample'].str.split('/', expand=True)[10].str.split('.', expand=True)[0]
         outlier_melt_sample = df_melt.pivot_table(columns=['#ID'], aggfunc='size').astype(int)
     outlier_melt_df = outlier_melt_sample.reset_index()
@@ -132,9 +132,9 @@ def read_wham_outlier(filename: str) -> pd.DataFrame:
     if df_wham['Outlier_Sample'].empty:
         df_wham['#ID'] = df_wham.apply(lambda _: ' ', axis=0)
         outlier_wham_sample = df_wham.pivot_table(columns=['#ID'], aggfunc='size').astype(int)
-        status = 'DataFrame is empty'
+        #status = 'DataFrame is empty'
     else:
-        status = 'DataFrame is not empty'
+        #status = 'DataFrame is not empty'
         df_wham['#ID'] = df_wham['Outlier_Sample'].str.split('/', expand=True)[11].str.split('.', expand=True)[0]
         outlier_wham_sample = df_wham.pivot_table(columns=['#ID'], aggfunc='size').astype(int)
     outlier_wham_df = outlier_wham_sample.reset_index()
@@ -228,7 +228,6 @@ def main():
     parser.add_argument("Wham-QC-outlier-low-filename", help="...")
     parser.add_argument("--melt-insert-size-file", nargs=1)
     args = parser.parse_args()
-    prefix = args.prefix_test
     merge_evidence_qc_table(args.ploidy_test_filename, args.medianCov_transposed_filename, args.WGD_scores_filename,
                             args.binwise_CNV_qValues_filename, args.Manta_QC_outlier_high_filename,
                             args.Melt_QC_outlier_high_filename,
--- a/src/sv-pipeline/scripts/make_evidence_qc_table.py
+++ b/src/sv-pipeline/scripts/make_evidence_qc_table.py
@@ -2,232 +2,259 @@
 This script creates a single table of sample QC & batching metrics from EvidenceQC.
 """
 
-# Importing libraries
 import argparse
 import pandas as pd
 from collections import Counter
 from numpy import array
 from functools import reduce
+from pathlib import Path
+import numpy as np
+
+ID_COL = "#ID"
+EMPTY_OUTLIERS = "EMPTY_ROWS_DROP"
 
 
-###########################################################################
-# From chrX estimated CN, set to 1 (if round(chrX_CN) < 2) or 2 (if round(chrX_CN) >=2)
 def read_ploidy(filename: str) -> pd.DataFrame:
     """
-    input: estimated_copy_numbers in a tab separated format
-    output: 24 columns, one per chromosome. For chrX:
-        if round(chrX_CN) < 2: chrX estimated CN = 1
-        if round(chrX_CN) >=2: chrX estimated CN = 2
+    Args:
+        filename: A tab-delimited file containing estimated copy numbers.
+    Returns:
+        A pandas DataFrame containing the following columns:
+        [id, chr1_CopyNumber, ..., chr22_CopyNumber, chrX_CopyNumber, chrY_CopyNumber, chrX_CopyNumber_rounded].
     """
-    df_ploidy = pd.read_csv(filename, sep='\t')
-    df_ploidy.loc[df_ploidy['chrX_CopyNumber'] <= 2, 'chrX_CopyNumber_rounded'] = 1
-    df_ploidy.loc[df_ploidy['chrX_CopyNumber'] > 2, 'chrX_CopyNumber_rounded'] = 2
+    df_ploidy = pd.read_csv(filename, sep="\t")
+    df_ploidy.loc[round(df_ploidy["chrX_CopyNumber"]) < 2, "chrX_CopyNumber_rounded"] = 1
+    df_ploidy.loc[round(df_ploidy["chrX_CopyNumber"]) >= 2, "chrX_CopyNumber_rounded"] = 2
     return df_ploidy
 
 
-###########################################################################
-# Median coverage (bincov_median)
 def read_bincov_median(filename: str) -> pd.DataFrame:
     """
-    input:  bincov_median” output from EvidenceQC (medianCov.transposed) in a tab separated format
-    output: the transpose of the table ('#ID', 'median_coverage')
+    Median coverage (bincov_median)
+    Args:
+        filename: bincov_median output from EvidenceQC (medianCov.transposed) in a tab separated format
+    Returns:
+        a pandas DataFrame containing the transposed table given in the input file.
     """
-    df_median = pd.read_csv(filename, sep='\t').T
-    df_median = df_median.rename_axis('#ID').reset_index()
-    df_median.columns = ['#ID', 'median_coverage']
+    df_median = pd.read_csv(filename, sep="\t").T
+    df_median = df_median.rename_axis(ID_COL).reset_index()
+    df_median.columns = [ID_COL, "median_coverage"]
     df_median = df_median.reset_index(drop=True)
     return df_median
 
 
-###########################################################################
-# Wgd_scores
 def read_wgd_scores(filename: str) -> pd.DataFrame:
     """
-    input: WGD_scores in a tab separated format
-    output: WGD_scores in the table
+    Args:
+        filename: a tab-delimited file containing wgd scores.
+    Returns:
+        A pandas DataFrame containing wgd scores for each sample.
     """
-    return pd.read_csv(filename, sep='\t')
+    df_wgd_scores = pd.read_csv(filename, sep="\t")
+    df_wgd_scores.rename(columns={'score': 'wgd_score'}, inplace=True)
+    return df_wgd_scores
 
 
-###########################################################################
-# EvidenceQC → untar → /ploidy_est/binwise_CNV_qValues.bed.gz → count number of bins with q-value < 0.05 for each sample
 def read_non_diploid(filename: str) -> pd.DataFrame:
     """
-    input: binwise_CNV_qValues in a tab separated format
-    output: the counts for number of bins with q-value < 0.05 for each sample
+    EvidenceQC → untar → /ploidy_est/binwise_CNV_qValues.bed.gz →
+    count number of bins with q-value < 0.05 for each sample
+    Args:
+        filename: a tab-delimited file containing binwise CNV qValues.
+    Returns:
+        A pandas DataFrame containing the number of bins with q-value < 0.05 for each sample.
     """
-    df_non_diploid = pd.read_csv(filename, sep='\t').T
-    df_non_diploid = df_non_diploid.T.iloc[0:, 3:]  # .rename_axis('#ID').reset_index()
-    new_c = df_non_diploid.iloc[0:, 0:]
-    nondiploid_counts = new_c[new_c < 0.05].count()
-    nondiploid_counts_df = pd.DataFrame(nondiploid_counts, columns=['nondiploid_counts']).rename_axis(
-        '#ID').reset_index()
+    df_non_diploid = pd.read_csv(filename, sep="\t").T
+    df_non_diploid = df_non_diploid.T.iloc[0:, 3:]
+    nondiploid_counts = df_non_diploid[df_non_diploid < 0.05].count()
+    nondiploid_counts_df = pd.DataFrame(nondiploid_counts, columns=["nondiploid_bins"]).rename_axis(
+        ID_COL).reset_index()
     return nondiploid_counts_df
 
 
-###########################################################################
-# Insert size
 def read_melt_insert_size(filename: str) -> array:
     """
-    input: melt_insert_size in sample data table
-    output: Array[Float] of melt_insert_size for each sample
+    Reads MELT insert size into an array.
+    Args:
+        filename: melt_insert_size in sample data table
+    Returns:
+        Array[Float] of melt_insert_size for each sample
     """
-    insert_size = pd.read_csv(filename, sep='\t')
-    melt_insert_size_array = pd.DataFrame(insert_size, columns=['melt_insert_size']).astype(float)
+    insert_size = pd.read_csv(filename, sep="\t")
+    melt_insert_size_array = pd.DataFrame(insert_size, columns=["mean_insert_size"]).astype(float)
     return melt_insert_size_array
 
 
-###########################################################################
-# Outliers from Manta, Melt, and Wham:
-# Manta:
-def read_manta_outlier(filename: str) -> pd.DataFrame:
+def get_col_name(caller: str, outlier_type: str) -> str:
+    return f"{caller}_{outlier_type}_outlier"
+
+
+def read_outlier(filename: str, outlier_col_label: str) -> pd.DataFrame:
     """
-    input:  manta_qc_low or high from EvidenceQC in a tab separated format
-    output: number of times sample appears in manta_qc_low or/and high
+    Args:
+        filename: A tab-delimited file containing the output of
+        Evidence QC on a caller (e.g., Manta) with low/high.
+        outlier_col_label: Column header for the outliers.
+    Returns:
+        A pandas DataFrame containing the number of times a sample
+        appears in the QC of a caller output.
     """
-    outlier_manta = filename.split('.')[-4] + '_' + filename.split('.')[-1]  # + '_' + filename.split('.')[-3]
-    df_manta = pd.read_csv(filename, sep='\t')
-    if df_manta['Outlier_Sample'].empty:
-        df_manta['#ID'] = df_manta.apply(lambda _: ' ', axis=0)
-        outlier_manta_sample = df_manta.pivot_table(columns=['#ID'], aggfunc='size').astype(int)
+    df = pd.read_csv(filename, sep="\t")
+    if df["Outlier_Sample"].empty:
+        df[ID_COL] = df.apply(lambda _: EMPTY_OUTLIERS, axis=0)
+        outlier_sample = df.pivot_table(columns=[ID_COL], aggfunc="size").astype(int)
     else:
-        df_manta['#ID'] = df_manta['Outlier_Sample'].str.split('/', expand=True)[10].str.split('.', expand=True)[0]
-        outlier_manta_sample = df_manta.pivot_table(columns=['#ID'], aggfunc='size').astype(int)
-    outlier_manta_df = outlier_manta_sample.reset_index()
-    outlier_manta_df.columns = ['#ID', str(outlier_manta) + '_outlier']
-    return outlier_manta_df
-
-
-# Melt:
-def read_melt_outlier(filename: str) -> pd.DataFrame:
-    """
-    input:   melt_qc_low or high output from EvidenceQC in a tab separated format
-    output:  number of times sample appears in melt_qc_low or/and high
-    """
-    outlier_melt = filename.split('.')[-4] + '_' + filename.split('.')[-1]
-    df_melt = pd.read_csv(filename, sep='\t')
-    if df_melt['Outlier_Sample'].empty:
-        df_melt['#ID'] = df_melt.apply(lambda _: ' ', axis=0)
-        outlier_melt_sample = df_melt.pivot_table(columns=['#ID'], aggfunc='size').astype(int)
-    else:
-        df_melt['#ID'] = df_melt['Outlier_Sample'].str.split('/', expand=True)[10].str.split('.', expand=True)[0]
-        outlier_melt_sample = df_melt.pivot_table(columns=['#ID'], aggfunc='size').astype(int)
-    outlier_melt_df = outlier_melt_sample.reset_index()
-    outlier_melt_df.columns = ['#ID', str(outlier_melt) + '_outlier']
-    return outlier_melt_df
-
-
-# Wham:
-def read_wham_outlier(filename: str) -> pd.DataFrame:
-    """
-    input:  wham_qc_low or high output from EvidenceQC in a tab separated format
-    output:  number of times sample appears in wham_qc_low or high outputs
-    """
-    outlier_wham = filename.split('.')[-4] + '_' + filename.split('.')[-1]  # + '_' + filename.split('.')[-3]
-    df_wham = pd.read_csv(filename, sep='\t')
-    if df_wham['Outlier_Sample'].empty:
-        df_wham['#ID'] = df_wham.apply(lambda _: ' ', axis=0)
-        outlier_wham_sample = df_wham.pivot_table(columns=['#ID'], aggfunc='size').astype(int)
-    else:
-        df_wham['#ID'] = df_wham['Outlier_Sample'].str.split('/', expand=True)[11].str.split('.', expand=True)[0]
-        outlier_wham_sample = df_wham.pivot_table(columns=['#ID'], aggfunc='size').astype(int)
-    outlier_wham_df = outlier_wham_sample.reset_index()
-    outlier_wham_df.columns = ['#ID', str(outlier_wham) + '_outlier']
-    return outlier_wham_df
+        df['Outlier_Sample'] = df['Outlier_Sample'].apply(Path)
+        df[ID_COL] = df['Outlier_Sample'].apply(lambda x: x.stem if '.' in x.suffix  else np.nan)
+        outlier_sample = df.pivot_table(columns=[ID_COL], aggfunc="size").astype(int)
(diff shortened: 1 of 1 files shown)


VJalili:
Thank you, @shadizaheri!!! It reads well, I made a few changes, mostly styling and abstracting methods, but please feel free to revert any if you prefer otherwise. Specifically, I made the following changes.

- There were multiple methods for reading the outputs of different methods (one for each of melt, manta, and wham). These methods were mostly implementing a similar functionality except for one array index. I refactored this into one method and made the array index as an input argument that can be adjusted w.r.t. to the method.

- You are composing some column names using filenames; e.g., from `bwa.Manta.QC.outlier.high` (filename), you compose `Manta_high` (column name). I think setting column names in the code w.r.t to the methods you call is easier and less error-prone than relying on the template of the filename. So, I made some changes similar to the following. 

```python
# Example 1
# from:
outlier_manta = filename_manta.split(".")[-4] + "_" + filename_manta.split(".")[-1]

# to
col_name = get_col_name("manta", outlier_type)

def get_col_name(caller: str, outlier_type: str) -> str:
    return f"{caller}_{outlier_type}_outlier"

# Example 2
# from
read_wham_outlier(filename_high_wham)
def read_wham_outlier(filename):
    outlier_wham = filename.split(".")[-4] + "_" + filename.split(".")[-1]
    ...
    outlier_wham_df.columns = ["#ID", str(outlier_wham) + "_outlier"]

# to
read_outlier(filename_high_wham, get_col_name("wham", "high"))
def read_outlier(filename, col_name):
    ...
    outlier_df.columns = ["#ID", col_name]
```

- I defined constant variable with its value set to `"#ID"`, and replaced all the references of `"#ID"` with the const var. This makes it less error-prone. 

- I fixed some issues with the command line args (optional vs. required, and the `-` to `_` conversion), and added a simple help message. Please feel free to extend the help messages if needed.

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+    Returns:
+        A pandas DataFrame containing the number of times a sample
+        appears in the QC of a caller output.
+    """
+    df = pd.read_csv(filename, sep="\t")
+    if df["Outlier_Sample"].empty:
+        df[ID_COL] = df.apply(lambda _: "NA", axis=0)
+        outlier_sample = df.pivot_table(columns=[ID_COL], aggfunc="size").astype(int)
+    else:
+        df[ID_COL] = df["Outlier_Sample"].str.split("/", expand=True)[idx].str.split(".", expand=True)[0]
```
VJalili, line 109: I am not sure how extracting sample ID from its URI is portable/reliable. But I'll leave it to you to decide how best the sample ID can be extracted.

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+
+        idx: The index of the cell in the array containing sample
+        URI split by `/` that contains sample ID.
+    Returns:
+        A pandas DataFrame containing the number of times a sample
+        appears in the QC of a caller output.
+    """
+    df = pd.read_csv(filename, sep="\t")
+    if df["Outlier_Sample"].empty:
+        df[ID_COL] = df.apply(lambda _: "NA", axis=0)
```
VJalili, line 106: You had set the ID to empty string, I have seen issues with that in particular when a line is trimmed before split by a delimiter. So, I changed the empty string with `NA`; however, I'm not sure what is the best approach here. Do you need this row?

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+    # all data frames
+    dfs = [df_ploidy, df_bincov_median, df_wgd_scores, df_non_diploid, df_manta_high_outlier,
+           df_melt_high_outlier, df_wham_high_outlier, df_total_high_outliers,
+           df_manta_low_outlier, df_melt_low_outlier, df_wham_low_outlier, df_total_low_outliers]
+    output_df = reduce(
+        lambda left, right: pd.merge(left, right, on=ID_COL, how="outer"), dfs
+    ).fillna(0)  # "none"
+
+    # save the file
+    return output_df.to_csv(f"{output_prefix}evidence_qc_table.tsv", sep="\t", header=True, index=False)
```
VJalili, line 193: what is returned here? I guess it is `None`, if so, you may remove `return`.

shadizaheri (commented):
Thank you so much, @VJalili  for your comments and suggestions for the script's style. Very helpful.

src/sv-pipeline/scripts/make_evidence_qc_table.py
shadizaheri: @VJalili This is supposed to return the final merged DataFrames in a "tsv" format.

src/sv-pipeline/scripts/make_evidence_qc_table.py
shadizaheri: @VJalili I think we need this row so that it can return `'0'` for the empty files (the files that we did not have any types of outliers).

src/sv-pipeline/scripts/make_evidence_qc_table.py
shadizaheri: @VJalili using `os.path.split()` could also be an option, but it was not working here.

VJalili: @shadizaheri do you have sample id in a different column of your data table? if so, that could be a better alternative, if not, it should be find to leave it as is.

src/sv-pipeline/scripts/make_evidence_qc_table.py
VJalili: Thanks! That makes sense to me.

src/sv-pipeline/scripts/make_evidence_qc_table.py
VJalili: Do you mean the filename?! I think since you're not using the returned value, it should be fine to remove `return`, but it should also be fine to leave it as is, it is up to you.

VJalili (approved):
Thank you, @shadizaheri!!! The script reads very well. I tested it and it runs as we discussed; though I would leave it to @mwalker174 or @epiercehoffman to check if it implements the logic we discussed.

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+    col_name = get_col_name("wham", outlier_type)
+    outlier_wham_df = read_outlier(filename_wham, col_name, 11)
+    dict_wham = dict(list(zip(outlier_wham_df[ID_COL], outlier_wham_df[col_name])))
+
+    # merging all the dictionaries
+    outlier_dicts = [dict_manta, dict_melt, dict_wham]
+    merged_dicts = Counter()
+    for counted in outlier_dicts:
+        merged_dicts.update(counted)
+    all_outliers = dict(merged_dicts)
```
epiercehoffman, line 149: When I did this I called `Counter()` on the concatenated sample ID columns from all the high outliers files, which seemed a little simpler/marginally more efficient because there were no dictionaries and you could count all of the callers at once

src/sv-pipeline/scripts/make_evidence_qc_table.py
epiercehoffman: Agreed, no need for a `return` since it is unused. Make sure to update the WDL with the expected output file name so that they match

epiercehoffman, line 193: Let's just make the prefix required so we can add a delimiter here
```suggestion
    return output_df.to_csv(f"{output_prefix}.evidence_qc_table.tsv", sep="\t", header=True, index=False)
```

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+
+    # save the file
+    return output_df.to_csv(f"{output_prefix}evidence_qc_table.tsv", sep="\t", header=True, index=False)
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "ploidy_test_filename",
+        help="Sets the filename containing ploidy tests.")
```
epiercehoffman, line 200:
```suggestion
        "estimated_copy_number_filename",
        help="Sets the filename containing estimated copy numbers per contig.")
```

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+
+    parser.add_argument(
+        "median_cov_transposed_filename",
```
epiercehoffman, line 203:
```suggestion
        "median_cov_filename",
```

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+        "wham_qc_outlier_low_filename",
+        help="Sets the filename containing Wham QC outlier low.")
+
+    parser.add_argument(
+        "melt_insert_size_file",
+        help="Sets the filename containing Melt insert size.")
+
+    parser.add_argument(
+        "--output-prefix",
+        default="",
```
epiercehoffman, line 244:
```suggestion
```

epiercehoffman (commented):
Great, this is looking very close to ready! I think it will be really helpful. Thanks @shadizaheri for your work on this. I think overall the code looks organized, readable, well-documented, and 99% correct (though I haven't had a chance to do a test run).

There's only one real change to the actual output values that I've requested, which is related to the rounded chrX copy number used for batching. My other suggestions are for code style / simplicity / efficiency.

I'd also like to see an example of the output. Can you share an example output table and the input files used for the test? I'd also like to test the script as part of the EvidenceQc workflow because there is some preparation of the inputs involved - let me know what your plan is for finalizing the WDL and integrating the changes across branches.

src/sv-pipeline/scripts/make_evidence_qc_table.py
shadizaheri: @epiercehoffman It is therefore necessary to check if the file is blank. Empty files should also include outlier columns. Even empty files need IDs when merging based on the `'#ID'` key at the end.  I think it will be more clear once you run the code using the empty and non-empty files.

src/sv-pipeline/scripts/make_evidence_qc_table.py
VJalili: That makes sense, thanks!!

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Args:
+        filename: a tab-delimited file containing binwise CNV qValues.
+
+    Returns:
+        A pandas DataFrame containing the number of bins with q-value < 0.05 for each sample.
+    """
+    df_non_diploid = pd.read_csv(filename, sep="\t").T
+    df_non_diploid = df_non_diploid.T.iloc[0:, 3:]  # .rename_axis("#ID").reset_index()
+    new_c = df_non_diploid.iloc[0:, 0:]
```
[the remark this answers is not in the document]
shadizaheri: This is an interesting perspective. I don't see a big difference between using 'sum' or 'counts'. I think having the code as it is (check below) can make further possible modifications easier.

```
    df_non_diploid = df_non_diploid.T.iloc[0:, 3:]
    new_c = df_non_diploid.iloc[0:, 0:]
    nondiploid_counts = new_c[new_c < 0.05].count()
```

src/sv-pipeline/scripts/make_evidence_qc_table.py
shadizaheri: @epiercehoffman I'm not sure if I understand the suggestion here. Can you please explain the suggestion, here?

src/sv-pipeline/scripts/make_evidence_qc_table.py
shadizaheri: Thank you @VJalili and @epiercehoffman ! This is a valid point. The only reason that I had the `'return'` here, was to test the script by outputting a tsv file. Once we test the script and have the wdl written, we can remove the `'return'` line from here.

src/sv-pipeline/scripts/make_evidence_qc_table.py
epiercehoffman: I think the prefix should be a required argument, so it does not need a default value

src/sv-pipeline/scripts/make_evidence_qc_table.py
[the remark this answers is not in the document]
epiercehoffman: I don't really care about sum vs. counts, but I think it's best to do this without making unnecessary copies of the dataframe. Why do you need new_c if it copies all the rows and columns?

src/sv-pipeline/scripts/make_evidence_qc_table.py
epiercehoffman: The return has no impact on whether the TSV file is written. You can delete the `return` keyword and keep the rest of the line and it will write the TSV file - for testing & for use in the workflow.

src/sv-pipeline/scripts/make_evidence_qc_table.py
epiercehoffman: Do you end up with an extra row for `NA` in the case of an empty file due to the outer join later? I see an extra row with no ID in your example output, which we would want to remove

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+    """
+    df_median = pd.read_csv(filename, sep="\t").T
+    df_median = df_median.rename_axis(ID_COL).reset_index()
+    df_median.columns = [ID_COL, "median_coverage"]
+    df_median = df_median.reset_index(drop=True)
+    return df_median
+
+
+def read_wgd_scores(filename: str) -> pd.DataFrame:
+    return pd.read_csv(filename, sep="\t")
```
epiercehoffman, line 48: Can you name the output column `wgd_score`? Currently it is named `score` in the original WGD file, but that is ambiguous in the unified output table

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+        filename: a tab-delimited file containing binwise CNV qValues.
+
+    Returns:
+        A pandas DataFrame containing the number of bins with q-value < 0.05 for each sample.
+    """
+    df_non_diploid = pd.read_csv(filename, sep="\t").T
+    df_non_diploid = df_non_diploid.T.iloc[0:, 3:]
+    new_c = df_non_diploid.iloc[0:, 0:]
+    nondiploid_counts = new_c[new_c < 0.05].count()
+    nondiploid_counts_df = pd.DataFrame(nondiploid_counts, columns=["nondiploid_counts"]).rename_axis(
```
epiercehoffman, line 66: Let's name this column `nondiploid_bins` to indicate where it came from
```suggestion
    nondiploid_counts_df = pd.DataFrame(nondiploid_counts, columns=["nondiploid_bins"]).rename_axis(
```

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+    ).fillna(0)  # "none"
+
+    # save the file
+    return output_df.to_csv(f"{output_prefix}.evidence_qc_table.tsv", sep="\t", header=True, index=False)
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "estimated_copy_number_filename",
```
epiercehoffman, line 199: Dashes are best practice for command-line arguments, rather than underscores. When you access these options later as variables the dashes will be automatically converted to underscores. \

I think since there are so many arguments they should be named rather than positional. You can add a short argument name if you want. 

Some of the files are marked optional in the WDL task - make sure that is handled on the script level.
```suggestion
        "--estimated-copy-number-filename",
```

src/sv-pipeline/scripts/make_evidence_qc_table.py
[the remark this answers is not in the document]
shadizaheri: I used `new_c`  before printing out/testing the code. We can easily delete `new_c` and modify as:

`    nondiploid_counts = df_non_diploid[df_non_diploid < 0.05].count()
`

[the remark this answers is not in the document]
epiercehoffman: Sounds good, let's go with that

src/sv-pipeline/scripts/make_evidence_qc_table.py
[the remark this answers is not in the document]
VJalili: Agree. I am more inclined to use [`.stem`](https://stackoverflow.com/a/47496703/947889), which should handle cases where you have multiple `.` in the filename, but `.split` should also be fine.

src/sv-pipeline/scripts/make_evidence_qc_table.py
shadizaheri: @epiercehoffman , this is a valid point, thanks for sharing. I also think that since these are required arguments, not optional, it might be better/easier if we go with underscore here. You can read more [here](https://stackoverflow.com/questions/12834785/having-options-in-argparse-with-a-dash)

epiercehoffman: You can make them required but still use `--`. I think that for this quantity of arguments they shouldn't all be positional.

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Returns:
+        A pandas DataFrame containing the number of bins with q-value < 0.05 for each sample.
+    """
+    df_non_diploid = pd.read_csv(filename, sep="\t").T
+    df_non_diploid = df_non_diploid.T.iloc[0:, 3:]  # .rename_axis("#ID").reset_index()
+    new_c = df_non_diploid.iloc[0:, 0:]
+    nondiploid_counts = new_c[new_c < 0.05].count()
+    nondiploid_counts_df = pd.DataFrame(nondiploid_counts, columns=["nondiploid_counts"]).rename_axis(
+        "#ID").reset_index()
```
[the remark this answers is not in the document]
shadizaheri: I see! I updated the script with ID_COL and checked to be sure that it's consistent all over the script.

src/sv-pipeline/scripts/make_evidence_qc_table.py
shadizaheri: @epiercehoffman , I think there could be positional for the python script, at the same time I agree that it's better to not have them as positional! Thanks for the comment. I applied your comment in [this](https://github.com/broadinstitute/gatk-sv/commit/0086235f678fb29d32c88aa9fdd15ba7a76a71b8) commit.

src/sv-pipeline/scripts/make_evidence_qc_table.py
shadizaheri: aha, now I see what you mean. Thanks for the comment and explanation. I removed `return`

src/sv-pipeline/scripts/make_evidence_qc_table.py
shadizaheri: Renamed to `wgd_score`.

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+def read_all_outlier(filename_manta: str, filename_melt: str, filename_wham: str, outlier_type: str) -> pd.DataFrame:
+    """
+
+    Args:
+        filename_manta: Outliers determined in EvidenceQC for Manta.
+        filename_melt: Outliers determined in EvidenceQC for MELT.
+        filename_wham: Outliers determined in EvidenceQC for Wham.
+        outlier_type:
```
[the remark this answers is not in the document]
shadizaheri: thanks for the comment! This got modified [here](https://github.com/broadinstitute/gatk-sv/blob/sz_evidence_qc_python_script/src/sv-pipeline/scripts/make_evidence_qc_table.py#:~:text=outlier_type%3A%20low%20or%20high)

shadizaheri (commented):
thanks for the comment! This got modified [here](https://github.com/broadinstitute/gatk-sv/blob/sz_evidence_qc_python_script/src/sv-pipeline/scripts/make_evidence_qc_table.py#:~:text=outlier_type%3A%20low%20or%20high)

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,264 @@
...
+    serialized to the given output filename.
+    """
+    df_ploidy = read_ploidy(filename_estimated_cn)
+    df_bincov_median = read_bincov_median(filename_mediancov)
+    df_wgd_scores = read_wgd_scores(filename_wgd)
+    df_non_diploid = read_non_diploid(filename_cnv_qvalues)
+    df_manta_high_outlier = read_outlier(filename_high_manta, get_col_name("manta", "high"))
+    df_melt_high_outlier = read_outlier(filename_high_melt, get_col_name("melt", "high"))
+    df_wham_high_outlier = read_outlier(filename_high_wham, get_col_name("wham", "high"), 11)
+    df_total_high_outliers = read_all_outlier(filename_high_manta, filename_high_melt, filename_high_wham, "high")
```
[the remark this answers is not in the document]
epiercehoffman: It is less efficient to read the file again from disk. Using the outputs of the previous function would decrease the number of times that you read from disk as well as the number of dataframe operations. You directly call and use read_outlier within read_all_outlier, so it would be very simple to make those dataframes inputs to read_all_outlier instead of calling the function again.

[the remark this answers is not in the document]
shadizaheri: We are in some sense using the output of other functions (`read_outlier`) inside the `read_all_outlier`. I don't think that making this change is necessary at this stage as I don't see much of a difference here.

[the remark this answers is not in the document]
epiercehoffman: Let's touch base offline, I think it will be easier to explain!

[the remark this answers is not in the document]
shadizaheri: Sure! Meantime I thank you for your precise comment and conversation!
I think, for now, this is not going to cause any problems for us as I have checked the Pandas DataFrame memory usage for a different number of rows, columns, and data types. Even for the number of rows (in our case number of samples) in the order of 5e6 and the number of columns in the order of 40, we are not going to pass a couple of megabytes. 

So, I won't expect this to cause a problem for us, at the same time I definitely consider your comment for the future.

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,274 @@
...
+    parser.add_argument(
+        "-a", "--manta-qc-outlier-low-filename",
+        help="Sets the filename containing Manta QC outlier low.", nargs=1)
+
+    parser.add_argument(
+        "-s", "--melt-qc-outlier-low-filename",
+        help="Sets the filename containing Melt QC outlier low.", nargs=1)
+
+    parser.add_argument(
+        "-w", "--wham-qc-outlier-low-filename",
```
epiercehoffman, line 246: -w is used twice as a short argument name, here and for wham high outlier
```suggestion
        "-q", "--wham-qc-outlier-low-filename",
```

shadizaheri: very nice catch! I fixed it [here](https://github.com/broadinstitute/gatk-sv/blob/sz_evidence_qc_python_script/src/sv-pipeline/scripts/make_evidence_qc_table.py#:~:text=%22%2Dr%22%2C%20%22%2D%2Dwham%2Dqc%2Doutlier%2Dlow%2Dfilename%22%2C).

shadizaheri (commented):
Thanks for your work on this!

src/sv-pipeline/scripts/make_evidence_qc_table.py
shadizaheri: The files have been set as optional.

shadizaheri:
@epiercehoffman and @VJalili thank you both for your very great comments and help on this PR! 
I believe I have addressed all the comments here.
You can test the script by running 
`python make_evidence_qc_table.py -y estimated_copy_numbers.txt -d bwr_medianCov.transposed.bed -g bwr_WGD_scores.txt -b binwise_CNV_qValues.bed -z bwr.Manta.QC.outlier.high -e bwr.Melt.QC.outlier.high -w bwr.Wham.QC.outlier.high -a bwr.Manta.QC.outlier.low -s bwr.Melt.QC.outlier.low -r bwr.Wham.QC.outlier.low -m melt_insert_size.tsv -o testing` command line. I am also attaching a zip file containing the input files as and a sample of the output file to this comment.
Please note that `melt_insert_size.tsv` is a tsv file that I exported from the data table to test the python script. Because the information for 'melt_insert_size.tsv' is already included in the data table as an output from the GatherSampleEvidence step, most likely we will not need this once we write the wdl as we can read this info from the data table in Terra.
[table_EvidenceQC_test.zip](https://github.com/broadinstitute/gatk-sv/files/10143461/table_EvidenceQC_test.zip)

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,269 @@
...
+    parser.add_argument(
+        "-s", "--melt-qc-outlier-low-filename",
+        help="Sets the filename containing Melt QC outlier low.")
+
+    parser.add_argument(
+        "-r", "--wham-qc-outlier-low-filename",
+        help="Sets the filename containing Wham QC outlier low.")
+
+    parser.add_argument(
+        "-m", "--melt-insert-size-filename",
```
epiercehoffman, line 240: Not necessary to merge.
Why: I think we should update every reference to this file in this script to call it "mean" insert size instead of "Melt". We currently calculate the insert size during MELT, but as we work towards deprecating MELT, that will likely change, so we should be more flexible/accurate with the naming here.
Also: "melt" > "mean" elsewhere in script
```suggestion
        "-m", "--mean-insert-size-filename",
```

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,269 @@
...
+    parser.add_argument(
+        "-g", "--wgd-scores-filename",
+        help="Sets the filename containing WGD scores.")
+
+    parser.add_argument(
+        "-b", "--binwise-cnv-qvalues-filename",
+        help="Sets the filename containing bin-wise CNV q-values.")
+
+    parser.add_argument(
+        "-z", "--manta-qc-outlier-high-filename",
```
epiercehoffman, line 216: Not necessary to merge: Scramble outlier files are missing here, so we should make sure to come back and update this script to handle them later. 
In the mean time, please remove the scramble inputs in the WDL task since the script does not use them.

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,269 @@
...
+    df_non_diploid = read_non_diploid(filename_cnv_qvalues)
+    df_manta_high_outlier = read_outlier(filename_high_manta, get_col_name("manta", "high"))
+    df_melt_high_outlier = read_outlier(filename_high_melt, get_col_name("melt", "high"))
+    df_wham_high_outlier = read_outlier(filename_high_wham, get_col_name("wham", "high"))
+    df_total_high_outliers = read_all_outlier(df_manta_high_outlier, df_melt_high_outlier, df_wham_high_outlier, "high")
+    df_manta_low_outlier = read_outlier(filename_low_manta, get_col_name("manta", "low"))
+    df_melt_low_outlier = read_outlier(filename_low_melt, get_col_name("melt", "low"))
+    df_wham_low_outlier = read_outlier(filename_low_wham, get_col_name("wham", "low"))
+    df_total_low_outliers = read_all_outlier(df_manta_low_outlier, df_melt_low_outlier, df_wham_low_outlier, "low")
+    df_melt_insert_size = read_melt_insert_size(filename_melt_insert_size)
```
epiercehoffman, line 183: Not necessary to merge / question of preference. When I tested without providing one of the outlier files, I noticed that it produced a column of NA values. I think I might prefer to omit the column if the corresponding file is not provided, but would be open to other views.

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -0,0 +1,269 @@
...
+    if df["Outlier_Sample"].empty:
+        df[ID_COL] = df.apply(lambda _: EMPTY_OUTLIERS, axis=0)
+        outlier_sample = df.pivot_table(columns=[ID_COL], aggfunc="size").astype(int)
+    else:
+        df['Outlier_Sample'] = df['Outlier_Sample'].apply(Path)
+        df[ID_COL] = df['Outlier_Sample'].apply(lambda x: x.stem if '.' in x.suffix else np.nan)
+        outlier_sample = df.pivot_table(columns=[ID_COL], aggfunc="size").astype(int)
+    outlier_df = outlier_sample.reset_index()
+    outlier_df.columns = [ID_COL, outlier_col_label]
+    return outlier_df
```
epiercehoffman, line 117: Not necessary to merge / question of preference. I think my preference would be to fill cells with 0, rather than `np.nan`, if the sample does not appear as an outlier at all in a specified category. This would allow the outlier columns to be stored as ints and may make downstream computation a bit simpler because users won't have to account for NaN values. But, I recognize that it's easier to have NaN values due to the dataframe merge later on, and I'm open to other views.

epiercehoffman (approved):
Thanks @shadizaheri for all the updates you've made this week and to both @shadizaheri and @VJalili for all your work on this! I think this is in good shape and can be merged - I've left a few comments of varying degrees of importance but I'm ok with merging as-is and returning to those later (more details below). I've tested it locally and it appears to work as expected. I'll include the details of my testing below for transparency.

Those are the important bits, the rest is details.

I've made a few comments, mainly related to required inputs/outputs. I have marked them as either "Address soon" if I believe they should ideally be addressed before merging the script (but ok if soon after) or "Not necessary to merge" if I think it's fine to come back to them in the future (without urgency). 

I approved the PR as-is because I think that when run as part of EvidenceQC in the way we typically expect (ie. with all of the inputs provided, except optionally omitting MELT outlier files), the script will perform as desired. So, the "Address soon" reflects that the script does not fully handle required/optional outputs correctly, but given the desire to get this PR in quickly, I would approve a merge as-is as long as we come back and fix these items soon to provide more flexibility with inputs. (Some comments relate to the WDL task as well so it would be good to read them over for those details even if we merge as-is.)

Testing details:
- Ran with all input files and checked: number of rows in output, no empty sample IDs, all values in expected range for metric, outlier counts add up across rows, rounded chrX as expected based on chrX CN (spot-checked), column names are good
- Spot-checked a couple of columns against the input files to make sure nothing got scrambled
- Ran with a gzipped estimated CN file and checked that the output is identical, since the provided test files were not gzipped but in the WDL several of the input files will be gzipped
- Ran removing one command line argument each time and checked for expected behavior. Worked fine to omit MELT outlier files, but I've left a few comments about optional/required inputs in other cases

PR #465, opened 2022-11-29
Build, push, and track docker images on GCR

VJalili:
This PR makes the following changes related to maintaining replicas of docker images on GCP and Azure.

- Adds `dockers_gcp.json` that contains only the docker images pushed to GCP;
- Updates the GitHub action to set the related arguments of the `build_docker.py` script to point to `dockers_gcp.json`;
- Replaces the deprecated `--gcr-project` argument of `build_docker.py` used in the GitHub actions with `--docker-repo`.


TedBrookings:
I'm not sure the intended outcome of this PR, so I'll need some clarification before I can review.

As it stands, this will cause the GitHub action to read from and update the new file dockers_gcp.json, but people working locally are likely to still use the default dockers.json. If the desire is to somehow make room for dockers_azure.json, or something similar, why not just have people working with azure specify that file? This seems likely to lead to confusion. Someone working locally on azure could update dockers.json, push changes to it, then someone else working on gcp my also make local changes to the same file and mess things up.

Like I said, I may just not understand the desired outcome.

VJalili:
You guessed the flow correctly; I can add a few clarifications, specifically:

- `Dockers_gcp.json` and `Dockers_azure.json` will be in complete sync, i.e., containing the references to identical images cloned on the two registries;
- Maybe drop `Dockers.json` as I am not sure what use-case it will serve;
- `Dockers_*.json` is modified by the `gatk-sv-bot` only and will contain references to images stored on Broad-managed buckets. This serves two purposes. First, building and pushing to two or more container registries will require the developer to set up accounts on all the registries, which increases the bar for setup. Second, not every contributor will have push access to Broad's buckets on these container registries. `Dockers_*.json` containing images pushed by the bot only ensures all images are stored in one bucket with a common access policy (e.g., all publicly available and stored in multi-region buckets).
- If a developer needs to push docker images to a different registry/bucket, they could use a separate `Dockers_*.json` file that is stored on their branch/fork but not pushed to the `main` branch of `Broadinstitute/gatk-sv`. 

This PR adds `Dockers_gcp.json`, and in the follow-up PRs, we can implement changes related to the above points and adjust them as needed.

TedBrookings:
I would then request changes. build_docker.py will still default to updating the still-existing dockers.json, which will simply be a source of confusion and collisions between automated and local docker updates. I think there are two reasonable plans:
1. Also change build_docker.py to default to dockers_gcp.json, and remove dockers.json.
2. Wait until you have a plan to update dockers_azure.json, and make the above changes plus put in place auto-build dockers_azure.json. I would also say as a bonus, having the dockers JSON still be selectable, but having the default dockers JSON be selected based on the docker repo would be a nice bonus for people working locally (not relied upon by automation though).

Of those two, I would definitely prefer approach 2. Approach 1 makes changes that don't change the functionality of the project ("trust us, this will be worth it later"). I don't think approach 2 is a particularly large or complex PR either. It's just an added build action.

TedBrookings:
My personal feeling is that a PR should improve the project. In this case, approach 1 only just barely qualifies. It effectively renames dockers.json to dockers_gcp.json without introducing bugs or confusion points. That is likely to be useful later, but is not really a huge win. I'm not the dictator of this project, so if you go with approach 1, I will approve (assuming there are no other issues). HOWEVER, I personally feel like if it was me, I'd hold off until building dockers for azure was a thing (i.e. either go with approach 2, or just not do a PR now).
As to making "building dockers for azure" a thing, I would note that the dockers don't have to actually work on azure! If you have a file dockers_azure.json and an azure docker repo, you could make a build action that builds dockers for that repo and maintains them in dockers_azure.json. Then you've effectively made a space for someone to go experiment at getting those dockers to work on azure. In my mind that's a win, because that person only has to worry about getting the dockers to work, not hooking up all the plumbing.

VJalili:
I agree with you, @TedBrookings. We can keep this PR open or close it and get back to it once we have concrete plans for Azure.

TedBrookings:
I would keep it open. I think it will be a small (in terms of lines of code) change when it's ready, and I'm happy to review it again then.

VJalili:
Closing since this functionality is implemented as part of other PRs.

PR #468, opened 2022-12-01
eliminate cram to bam conversion when possible

tedsharpe:


mwalker174 (approved):
Thank you @tedsharpe for devising and implementing this - we're looking at savings of ~$0.50 per sample! It has taken me a while to review since this required thorough testing and evaluations. 

One change I needed to revert was with using only `chr1` for the `-c` option (specifying contigs to estimate read statistics from) in wham rather than the full genome. It turns out that `chr1` is not a representative portion of the genome to use and this dramatically altered the read statistics in some cases, leading to substantially fewer calls. The magnitude of this effect was surprising, but luckily this does not affect run time much.

Otherwise, I just did some minor tidying up and retained a few features we need, like the "ReviseBase" version of CramToBam, which may be needed occasionally. It also seems that gCNV postprocessing memory requirements have gone up a tad during my testing, so I bumped that up as well.

The CI build is failing due to the MELT, but this is a known issue right now, so it is still OK to merge.

PR #470, opened 2022-12-06
Patch to remove duplicate sites from VCF

epiercehoffman:
### Updates
Add patch to remove duplicate sites from VCF. Coordinates, SV type, and many INFO fields are compared to find identical events represented more than once in the VCF, and then the duplicated records are merged. A de-duplicated VCF is produced. A table of duplicated events is also produced to match Xuefang's approach to allow for comparison.

Merging duplicate records:
* The first variant ID is kept
* EVIDENCE and ALGORITHMS fields are merged
* FILTER is chosen based on `filter_precedence = ['PASS', 'UNRESOLVED', 'PESR_GT_OVERDISPERSION', 'BOTHSIDES_SUPPORT']`, where the highest precedence status is kept if present. If none of the listed statuses appear, the union of FILTER statuses across records is kept
* The genotypes are merged (the best-quality non-ref GT is kept, or a null GT if present, and the corresponding GT fields are also kept)

### Testing
Tested on 1KG and AoU (2000 samples). Examined rate of GT mismatch. Spot-checked some events to check EVIDENCE, ALGORITHMS, FILTER, and GTs were merged as expected. Checked number of records removed corresponds to number of duplicates found.

WDL validated with womtool. Will test full workflow after feedback on the approach.

### DIFF
--- a/src/sv-pipeline/scripts/remove_duplicates.py
+++ b/src/sv-pipeline/scripts/remove_duplicates.py
@@ -1,0 +1,136 @@
+import pysam
+import argparse
+
+
+def records_match(record, other):
+    """
+    Test if two records are same SV: check chromosome, position, stop, SVTYPE, SVLEN (for insertions),
+    STRANDS (for BNDS and INVs), CPX_TYPE and CPX_INTERVALS (for CPX), and CHR2/END2 (for multi-chromosomal events)
+    """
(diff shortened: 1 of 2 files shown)


epiercehoffman:
Incorporated into ManualReview.wdl on xz_fixes_3 with updates

PR #471, opened 2022-12-09
Fix a Typo colums to clumns

shadizaheri:


epiercehoffman (approved):
Looks good to me. Ran a quick test and confirmed that the script now succeeds if the insert size input is not provided.

PR #472, opened 2022-12-13
CleanVcfPart5 fix

mwalker174:
Addresses `None` type comparison errors in clean vcf part 5. Tested manually in a Docker container for correctness on ref panel test files and that the offending file from gnomAD no longer causes an error.

### DIFF
--- a/src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
+++ b/src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
@@ -123,18 +123,22 @@ def main():
 
                 if gt5kb_del:
                     for sample_obj in record.samples.itervalues():
-                        if not sample_obj['GQ'] is None and sample_obj['RD_CN'] >= 2:
+                        if not sample_obj['GQ'] is None and \
+                                (sample_obj['RD_CN'] is not None and sample_obj['RD_CN'] >= 2):
                             sample_obj['GT'] = (0, 0)
-                        elif not sample_obj['GQ'] is None and sample_obj['RD_CN'] == 1:
+                        elif not sample_obj['GQ'] is None and \
+                                (sample_obj['RD_CN'] is not None and sample_obj['RD_CN'] == 1):
                             sample_obj['GT'] = (0, 1)
                         elif not sample_obj['GQ'] is None:
                             sample_obj['GT'] = (1, 1)  # RD_CN 0 DEL
 
                 if gt5kb_dup:
                     for sample_obj in record.samples.itervalues():
-                        if not sample_obj['GQ'] is None and sample_obj['RD_CN'] <= 2:
+                        if not sample_obj['GQ'] is None and \
+                                (sample_obj['RD_CN'] is not None and sample_obj['RD_CN'] <= 2):
                             sample_obj['GT'] = (0, 0)
-                        elif not sample_obj['GQ'] is None and sample_obj['RD_CN'] == 3:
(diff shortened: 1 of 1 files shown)


src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -123,18 +123,22 @@ def main():
 
                 if gt5kb_del:
                     for sample_obj in record.samples.itervalues():
-                        if not sample_obj['GQ'] is None and sample_obj['RD_CN'] >= 2:
+                        if not sample_obj['GQ'] is None and \
```
VJalili, line 126: Would `sample_obj['GQ'] is not None` be better?

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -123,18 +123,22 @@ def main():
...
+                                (sample_obj['RD_CN'] is not None and sample_obj['RD_CN'] >= 2):
                             sample_obj['GT'] = (0, 0)
-                        elif not sample_obj['GQ'] is None and sample_obj['RD_CN'] == 1:
+                        elif not sample_obj['GQ'] is None and \
+                                (sample_obj['RD_CN'] is not None and sample_obj['RD_CN'] == 1):
```
VJalili, line 130: Would the following syntax be a good alternative?

```python
int(sample_obj['RD_CN'] or 0) == 1
```

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
```diff
@@ -160,8 +164,9 @@ def main():
                 if record.id in sexchr_revise:
                     for sample in record.samples:
                         if sample in male_samples:
-                            cn = int(record.samples[sample]['RD_CN'])
-                            if cn is not None and cn > 0:
+                            cn = record.samples[sample]['RD_CN']
+                            if cn is not None and int(cn) > 0:
+                                cn = int(cn)
```
VJalili, line 169: Are you separating `int` casting because you expect `record.samples[sample]['RD_CN']` containing values not convertible to the `int` type?

VJalili (commented):
LGTM, thank you, @mwalker174! I left a few syntax suggestions; I have not tested the suggestions, so please feel free to keep your code unchanged if you see functionality differences.

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
mwalker174: Yes definitely more pythonic/readable. According to [this](https://docs.python.org/3/reference/expressions.html#operator-precedence) the order of operations is `is` > `not` > `and` so the two are equivalent though.

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
mwalker174: Oh that's nifty!

src/sv-pipeline/04_variant_resolution/scripts/clean_vcf_part5_update_records.py
mwalker174: Yes it could be `None` which causes an error. Is there a better way?

VJalili (approved):

PR #475, opened 2022-12-16
Add TSV rename map option to RenameVcfSamples

epiercehoffman:
### Updates
Revamp RenameVcfSamples
* Add option to provide File input to map from old to new sample IDs instead of string arrays
* Make check for whether all samples have a new ID provided optional (slower)
* Fix disk size determination to use GB :) 
* Add test JSON

### Testing
* Tested WDL with string arrays & checking that all samples have a new ID specified (test JSON)
* Tested WDL with File rename map input for a subset of the samples, without the check that all samples have a new ID specified


wdl/RenameVcfSamples.wdl
```diff
@@ -62,9 +66,10 @@ task RenameVcfSamplesTask {
 
   command <<<
     set -euo pipefail
-    paste ~{current_sample_list} ~{new_sample_list} > dict.tsv
-    /opt/sv-pipeline/scripts/vcf_replacement_samples.py --vcf ~{vcf} --dict dict.tsv > reheader.list
-    bcftools reheader --samples reheader.list -o ~{prefix}.vcf.gz ~{vcf}
+    if ~{check_rename_all_samples}; then
+      python /opt/sv-pipeline/scripts/vcf_replacement_samples.py --vcf ~{vcf} --dict ~{sample_id_rename_map} > reheader.list
+    fi
```
mwalker174, line 71: I'd feel better requiring this check - it should be fast no?

mwalker174 (changes requested):

wdl/RenameVcfSamples.wdl
epiercehoffman: Currently the default behavior is to perform the check (`check_rename_all_samples = true`), but if a map to rename only a subset of samples (instead of providing an old & new sample ID for all samples in the VCF) is provided, the check would fail, so I wanted to provide a way to turn it off in that situation. Does that seem reasonable?

mwalker174: What's the behavior of bcftools when a sample name is invalid? I think it would be better to either rely on bcftools catching such errors or, if it doesn't, having a simple bash command to do the check rather than a script.

epiercehoffman: bcftools does not raise an error when there is a sample ID in the renaming map that doesn't appear in the header. 

I didn't see a good way to do the check in bash, but I replaced the script with a very short inline python block (the script wrote an output file that is no longer needed). By default the check is enabled, so it will catch incorrect IDs when it checks that every sample in the header appears in the renaming map. If a user wishes to rename only a subset of samples, they can either 1) disable the check or 2) include all samples in the renaming map anyway.

I tested this with incorrect sample IDs (failed as expected when check=true), missing sample IDs (failed as expected when check=true and succeeds as expected when check=false), and extra sample IDs (succeeds as expected) in the renaming map.

inputs/templates/test/RenameVcfSamples/RenameVcfSamples.json.tmpl
```diff
@@ -0,0 +1,321 @@
+{
+  "RenameVcfSamples.vcf": {{ test_batch.clean_vcf | tojson }},
+  "RenameVcfSamples.prefix": {{ test_batch.name | tojson }},
+  "RenameVcfSamples.current_sample_ids": [
```
mwalker174, line 4: Just thinking about this again. In order to be consistent with other templates we should use `{{ test_batch.samples }}` here and create a new `{{ test_batch.alternate_ids }}` or the like for the new ids.

mwalker174 (approved):
Thanks! One last minor suggestion here.

PR #476, opened 2022-12-19
be more permissive with baf calculation

tedsharpe:


mwalker174 (approved):
Thank you @tedsharpe looks good here.

PR #489, opened 2023-01-26
Fix a few build errors of docker images

VJalili:
This PR fixes the following errors in building docker images.

- In `samtool-cloud/Dockerfile`, update the Google Cloud SDK version to its current latest, and install it using `apt` instead of building from the source code.
- The GATK GitHub repository stores `~15GB` data via LFS, which are fetched while cloning the repository for building the `sv-base` docker file. Downloading `~15GB` data is a slow process on the GitHub Action runners, leading to out-of-disk issues. The LFS-hosted content for GATK does not seem to be needed, so this PR disables downloading/cloning them.
- Add missing dependencies to [sv-pipeline-virtual-env](https://github.com/broadinstitute/gatk-sv/compare/main...VJalili:update-dockers?expand=1#diff-e303d776623b30eaa600e3dedc6f1fa991bc890219adf96b75f46c96ebba3119) and [dockerfiles/sv-utils-env/Dockerfile](https://github.com/broadinstitute/gatk-sv/compare/main...VJalili:update-dockers?expand=1#diff-f40cc8b63fc3dafa83f5f6776c01d9e4362cf3aa375ab8b645f2e9b4037e9d70).


VJalili:
The following error can be fixed by removing `--freeze-installed` from [this line](https://github.com/VJalili/gatk-sv/blob/318a49e189657a022704afaed4a7c5f3fdebf553/dockerfiles/sv-pipeline-virtual-env/Dockerfile#L24), though not sure if that defeats the purpose of using `--freeze-installed`.


> package matplotlib-3.5.1-py310hff52083_0 requires pyqt, but none of the providers can be installed

dockerfiles/sv-pipeline-virtual-env/Dockerfile
```diff
@@ -49,7 +49,7 @@ RUN export SETUPTOOLS_VERSION=$(python -c 'import setuptools; print(setuptools._
 
 # pybedtools needs to be installed via pip because it doesn't like the updated python
 # hail's latest version is only available via pip or local build
-ARG PIP_PKGS="pybedtools==0.9.0 hail==0.2.93"
+ARG PIP_PKGS="pybedtools==0.9.0 hail==0.2.93 more-itertools==9.0.0"
```
mwalker174, line 52: Did you pick this change from my branch? Shouldn't be necessary for building, I just added it for a new script.

mwalker174 (approved):
@VJalili great thank you for working through all the errors. One change in there that I think you don't need, but otherwise looks great!

dockerfiles/sv-pipeline-virtual-env/Dockerfile
VJalili: Yes, it was from your branch! Good to know it was for testing purposes only; I'll revert it.

VJalili:
@mwalker174 Thanks for the prompt review and approval. I can reproduce the failing test locally, so I suppose it is unrelated to the runner's environment. Any thoughts on how that could be fixed? I can successfully build the image locally when I remove --freeze-installed from [this line](https://github.com/VJalili/gatk-sv/blob/318a49e189657a022704afaed4a7c5f3fdebf553/dockerfiles/sv-pipeline-virtual-env/Dockerfile#L24), though I am not sure if I should remove that flag. I also tried adding the missing dependency, but that led to a decent number of missing chains of dependencies. So, I am guessing there should be a different reason why this image fails the build. Any thoughts?

PR #490, opened 2023-01-29
Fix single sample pipeline REF_PANEL_GENOTYPES filter bug

cwhelan:
The single sample pipeline was overfiltering variants because of a bug in how it was comparing the variant sample lists for CVNs between calls in the single sample VCF and calls in the reference panel VCF. This turned out to be due to bedtools truncating that field of the bed file during an intersection (see https://github.com/arq5x/bedtools2/issues/919). I've modified the script with a slightly ugly workaround -- it now saves the pre-intersection files to temporary files, loads the reference panel intervals file into a dict, does the intersection, and then looks up the sample list from the dict. Unfiltered variants from the single sample pipeline on NA12878 rise from 4328 to 5850. Due to the nature of the bug the incorrectly filtered variants in were likely all very common deletions and duplications. 

We should probably consider re-writing this script to not be so dependent on a complicated bedtools pipeline.


mwalker174 (approved):
Thanks @cwhelan, looks good from here. Concerning that bedtools is doing this since it's so pervasive in the pipeline, so we should make sure to understand the underlying issue. Have you tested this in a fresh run?

cwhelan:
Yes, sorry, I should have documented my test run more fully here. Tested in the Terra workspace: `CW_Test_GATK-Structural-Variants-Single-Sample_2023-01-15-00-48-02_test_fix`

After completion of a fresh run with an updated docker file the pass variant counts from the metrics file are:

```
final_vcf_DEL_pass_count	3222
final_vcf_DUP_pass_count	735
final_vcf_INS_pass_count	1451
final_vcf_INV_pass_count	13
final_vcf_CTX_pass_count	0
final_vcf_CNV_pass_count	116
final_vcf_CPX_pass_count	112
```

For a total of 5,649.

inputs/values/dockers.json
```diff
@@ -13,12 +13,12 @@
...
-  "sv_pipeline_hail_docker": "us.gcr.io/broad-dsde-methods/gatk-sv/sv-pipeline:2023-01-19-v0.26.8-beta-5d1d5e0c",
-  "sv_pipeline_updates_docker": "us.gcr.io/broad-dsde-methods/gatk-sv/sv-pipeline:2023-01-19-v0.26.8-beta-5d1d5e0c",
-  "sv_pipeline_qc_docker": "us.gcr.io/broad-dsde-methods/gatk-sv/sv-pipeline:2023-01-19-v0.26.8-beta-5d1d5e0c",
-  "sv_pipeline_rdtest_docker": "us.gcr.io/broad-dsde-methods/gatk-sv/sv-pipeline:2023-01-19-v0.26.8-beta-5d1d5e0c",
+  "sv_pipeline_base_docker": "sv-pipeline:cw-fix-refpanelgt-filter-c2bca8",
+  "sv_pipeline_docker": "sv-pipeline:cw-fix-refpanelgt-filter-c2bca8",
+  "sv_pipeline_hail_docker": "sv-pipeline:cw-fix-refpanelgt-filter-c2bca8",
+  "sv_pipeline_updates_docker": "sv-pipeline:cw-fix-refpanelgt-filter-c2bca8",
+  "sv_pipeline_qc_docker": "sv-pipeline:cw-fix-refpanelgt-filter-c2bca8",
+  "sv_pipeline_rdtest_docker": "sv-pipeline:cw-fix-refpanelgt-filter-c2bca8",
```
VJalili, line 21: Would this tag not mean the image should be pulled from the docker hub and not gcr?

cwhelan:
Won't the github action on merge update this with the right gcr pointers? Or am I misunderstanding how that's supposed to work?

VJalili:
Chris and I discussed this case via DM, possible build issues will most probably be resolved when https://github.com/broadinstitute/gatk-sv/pull/489 is merged.

PR #492, opened 2023-02-01
Update the style of `build_docker.py`

VJalili:
This PR implements a few style updates to `build_docker.py` and replaces `termcolor` with built-in features. Specifically:

- [x] Consistent use of line breaks for each argument;
- [x] Consistent use of indentation before arguments;
- [x] Consistent use of double and single quotations for strings;
- [x] Consistent use of line break before closing parenthesis;
- [x] Use the backtick for quoting a substring;
- [x] Use the number of spaces for indentation instead of giving a string containing the number of spaces intended;
- [x] `termcolor` is currently the only non-standard dependency of this script. This PR implements a method that provides the same functionality with built-in features. This would simplify running this script in another environment without installing additional packages.


mwalker174 (approved):

VJalili:
Thanks @mwalker174!

PR #493, opened 2023-02-05
Remove depth vcf from pesr merging in FilterBatchSamples

mwalker174:
Addresses a bug where depth calls were being included in the merged PESR calls vcf.


epiercehoffman (approved):
Good catch

PR #499, opened 2023-02-24
New RdTest mode for very large variants

cwhelan:
This PR introduces a new RdTest genotyping mode intended to be applied to very large variants.

RdTest currently genotypes events larger than 1MB by only looking at depth information from the bins covering the middle 1MB of the event. In some cases, especially for very large events, the middle 1MB can fall into regions of the genome for which we do not collect coverage (eg centromeres) which can cause the depth to appear to be zero and make every sample genotype as a homozygous deletion. In other case it may cause incorrect results (eg if the middle of the event is covered by segmental duplications or a nested CNV). This behavior is currently controlled by the `-s, --sizefilter` parameter. 

This change adds a new genotyping method that avoids this problem. In the new method, RdTest loads depth from `m` windows of size `n` with centers that are evenly spaced across the event. Based on testing I have set the number of windows `m` to 500 and the window size `n` to 2000 -- this results in 1MB of depth data being loaded as before. For each window the median depth is calculated. The resulting medians are then used as inputs to the regular RdTest algorithms. Events over a given size threshold (set to 2.5MB by default) can be genotyped according to this method. This behavior can be controlled by the `-x, --verylargeeventsize`, `--verylargevariantpoints`, and `--verylargevariantwindows` parameters.

Note that I have tried to be conservative about changing RdTest in other scenarios. This means that the old "middle megabase" method is still used for events between 1MB and 2.5MB in size.

I've tested this method on a number of events and data from the gnomad and SSC cohorts. I've also run the single sample pipeline through with this version (I think this mode was triggered a few times). 

One note is that loading depth data from these windows is slower than loading a contiguous 1MB chunk of data. We should assess how long it takes in a full run on a large test cohort batch to make sure that the performance is still acceptable (I don't have my exact timing data available anymore but I believe that genotyping a large event took a few times longer in a gnomad batch than using the previous method).

This PR also includes a few other changes to RdTest, including renaming the sample include and exclude lists, refactoring some chunks of code into methods for readability, and adding automatic failure on warnings and printing of tracebacks in case of errors.


mwalker174:
@cwhelan looks great! Have you done a test run with `inputs/build/ref_panel_1kg/test/GATKSVPipelineBatch/GATKSVPipelineBatch.FromSampleEvidence.json`? I'd want to see that that runs ok and look at the cleaned vcf.

cwhelan:
I've completed a run with `inputs/build/ref_panel_1kg/test/GATKSVPipelineBatch/GATKSVPipelineBatch.FromSampleEvidence.json`

The clean vcf is here: `gs://broad-methods-cromwell-exec-bucket-instance-8/GATKSVPipelineBatch/ba3adf13-b406-4577-84b4-9e03644b7ff0/call-MakeCohortVcf/MakeCohortVcf/7c7a73d5-15dc-44b9-8545-0fb104d57672/call-CleanVcf/CleanVcf/a780c30e-6d96-4c49-828d-b77cc3475ab0/call-ConcatCleanedVcfs/ref_panel_1kg.cleaned.vcf.gz`

So far by scouring the logs I've found a few sites where the new genotyping mode was invoked: 

`chr18:68537528-80263001` in GenotypeDepthPart2 / RdTestGenotypeOver5kb
`chr14:80522636-106847714` in TrainRDGenotyping
`chr6:153431038-160785696` in CPX genotyping

There may be more, I'll keep looking through the logs.

mwalker174 (approved):
@cwhelan Overall call set looks good, so no regressions on the reference panel. I looked at those regions and there are events in those regions but the genotypes look the same against my latest run as of BAF-from-CRAM updates.

PR #502, opened 2023-03-07
Build and publish images to multiple registries in one job, & update the dependencies of the action

VJalili:
To build and publish Docker images to GCR and ACR, previously, we needed to run `build_docker.py` once for each of the registries; hence, the `Docker Images` workflow was implemented accordingly. However, we observed a few ephemeral errors with this approach where with the changes to the git status in the first run of `build_docker.py`, the second run cannot read git HEAD. We could solve this issue by re-running the failed action.  

To improve the reliability of the action and decrease its runtime, I added multi-registry support to `build_docker.py` in https://github.com/broadinstitute/gatk-sv/pull/507, which allows building and publishing Docker images to multiple container registries in one run.

This PR updates the `Docker Images` workflow to build and publish images to GCR and ACR in one _step_; specifically:

- [x] Login to Azure and GCP at the same time. The login order, first Azure, then GCP, is important since the Azure login process first logs out a previous login session; hence, if login with GCP first, then Azure, it results in logging out from GCP.
- [x] Build images, tag them with both ACR and GCR labels, push to both container registries, and update `dockers_gcp.json` and `dockers_azure.json` in one _step_. In addition to the advantages and motivations discussed above, this approach also makes the runtime faster since the images are built only once. Note that while `build_docker.py` can reuse images between multiple runs, we need to prune after each run because we have cases where we run out of disk space by not deleting images after they are built and pushed. 
- [x] Force-push the commit made by the bot to the _main_ branch, since we have seen cases where `git push` was rejected because of the changes on the branch after `git rebase`. (See the comment in the action file for details.) 

Additionally, this PR updates a few dependencies of the GitHub actions to address the deprecation warnings. Specifically, the following changes are implemented. 

- [x] Updates the checkout action to their current latest version to solve the node version deprecation warning; 
- [x] Updates setting output variables by replacing `set-output` with `GITHUB_OUTPUT`;
- [x] Remove the actions for installing `termcolor` as the current version of `build_docker.py` does not need this package;
- [x] Update the version of the python setup action to its current latest version. 

Some test runs of this branch on a separate fork are: 
- [`sv_pipeline_docker`](https://github.com/Genometric/gatk-sv/actions/runs/4349148966);
- [`testwdls`](https://github.com/Genometric/gatk-sv/actions/runs/4349148987);
- [`pytest`](https://github.com/Genometric/gatk-sv/actions/runs/4349148958).

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -11,9 +11,9 @@ jobs:
     name: Linting
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Set up Python 3.8
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v3
       with:
         python-version: 3.8
     - name: Install dependencies
--- a/.github/workflows/sv_pipeline_docker.yml
+++ b/.github/workflows/sv_pipeline_docker.yml
@@ -30,7 +30,7 @@ jobs:
       image_tag: ${{ steps.image_tag.outputs.IMAGE_TAG }}
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
         with:
           # By default, this checks out only the current commit;
           # however, since a diff between the current commit and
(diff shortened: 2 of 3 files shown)


mwalker174 (approved):
This looks great, thank you @VJalili

VJalili:
Thank you, @mwalker174, for the review! I hope we will not observe the ephemeral error anymore!

PR #505, opened 2023-03-08
Add option to subset to primary contigs during BatchEvidenceMerging

mwalker174:
This is a feature for legacy PE/SR files generated with a version of the pipeline that did not properly sort them. The files, which included non-primary contigs (e.g. HLA sequences), were sorted with `sort -k1,1V` which results in ordering inconsistent with the reference dictionary. This causes a GATK error in `PrintSVEvidence`. 

To correct the issue, it is faster to just subset the files rather than try to sort them in dictionary order since the pipeline does not evaluate evidence on non-primary contigs.

Tested on GatherBatchEvidence with the 1kgp reference panel.

Update: now also includes the option to update sample IDs in PE/SR/BAF files. This functionality was previously lost with the introduction of merging with PrintSVEvidence.

### DIFF
--- a/wdl/BatchEvidenceMerging.wdl
+++ b/wdl/BatchEvidenceMerging.wdl
@@ -13,6 +13,7 @@ workflow BatchEvidenceMerging {
     File reference_dict
     File primary_contigs_fai
     Boolean subset_primary_contigs  # If true, input PE/SR/BAF files will be subsetted to primary contigs only
+    Boolean rename_samples  # If true, rename samples to IDs in the input array
     String batch
     String gatk_docker
     RuntimeAttr? runtime_attr_override
@@ -27,6 +28,7 @@ workflow BatchEvidenceMerging {
       reference_dict = reference_dict,
       primary_contigs_fai=primary_contigs_fai,
       subset_primary_contigs=subset_primary_contigs,
+      rename_samples=rename_samples,
       gatk_docker = gatk_docker,
       runtime_attr_override = runtime_attr_override
   }
@@ -40,6 +42,7 @@ workflow BatchEvidenceMerging {
       reference_dict = reference_dict,
       primary_contigs_fai=primary_contigs_fai,
       subset_primary_contigs=subset_primary_contigs,
+      rename_samples=rename_samples,
       gatk_docker = gatk_docker,
       runtime_attr_override = runtime_attr_override
   }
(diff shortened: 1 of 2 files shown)


wdl/GatherBatchEvidence.wdl
```diff
@@ -44,6 +44,8 @@ workflow GatherBatchEvidence {
     File? ref_panel_bincov_matrix
     File? bincov_matrix
     File? bincov_matrix_index
+    Boolean subset_primary_contigs = false  # If true, input PE/SR/BAF files will be subsetted to primary contigs only
```
epiercehoffman, line 47: Do you want to clarify that this is only necessary for legacy files?

wdl/BatchEvidenceMerging.wdl
```diff
@@ -110,6 +127,34 @@ task MergeEvidence {
...
+            | bgzip > $OUT
+        elif [[ "~{evidence}" == "sr" ]]; then
+          zcat $fil \
+            | awk -F'\t' -v OFS='\t' -v SAMPLE="$sample" '!second_file{chroms[$1]; next} {~{if subset_primary_contigs then "if ($1 in chroms)" else ""} print ~{if rename_samples then "$1,$2,$3,$4,SAMPLE" else "$0"} }' contigs.list second_file=1 - \
+            | bgzip > $OUT
+        else
+          # baf - also uniquify records from old files
+          zcat $fil \
+            | awk -F'\t' -v OFS='\t' -v SAMPLE="$sample" '!second_file{chroms[$1]; next} {~{if subset_primary_contigs then "if ($1 in chroms)" else ""} print ~{if rename_samples then "$1,$2,$3,SAMPLE" else "$0"} }' contigs.list second_file=1 - \
+            | awk -F'\t' -v OFS='\t' '!_[$1"_"$2]++' - \
```
cwhelan, line 150: I don't know the background on this but this seems like a bit of a hidden thing to have happen if one of these two options is specified. I could imagine someone getting pretty confused if they ever had to debug an issue that revolved around duplicate records. Could we always do this if BAF is specified? Or make an additional option to do it?

cwhelan: Or at least put in an explicit comment / warning somewhere, if we don't want to go back and re-test?

cwhelan (approved):
This looks fine to me -- I was surprised by the deduplication behavior in BAF files so you might want to make that more explicit somehow (and/or always do it?)

The data hoarder in me hates to throw away data that someone might potentially want to look at for an off-label research use of the pipeline or something, but given storage costs we probably don't want to keep these files around that long anyway so I won't object.

wdl/BatchEvidenceMerging.wdl
mwalker174: Added another comment at the top about this

PR #511, opened 2023-03-16
Change ref allele to N if unsupported during vcf standardization

mwalker174:
Addresses a rare issue with vcfs having a non-ACGTN reference allele. These are not currently supported in gatk, thus causing an error during SVCluster in ClusterBatch.

This adds a simple check for unsupported reference alleles in `svtk standardize` and changes to an `N` if it's invalid.

Tested locally on a problematic WHAM vcf from AoU-PhaseII.


cwhelan (approved):
LGTM

PR #514, opened 2023-03-23
Add missing space in functional annotation

RCollins13:
This is an ultra-simple PR that adds a missing space to the `SVAnnotate` task in `AnnotateFunctionalConsequences.wdl`.

The current version in the main branch includes the following line (on line 86):
```
~{"--max-breakend-as-cnv-length" + max_breakend_as_cnv_length} \
```

This causes an error when the optional `max_breakend_as_cnv_length` input is supplied because there is no space between `"--max-breakend-as-cnv-length"` and the user input value.


mwalker174 (approved):
Thanks @RCollins13, looks good!

PR #516, opened 2023-03-27
Extend STR workflow to collect additional locus-level metrics

VJalili:
This PR extends the STR workflow in two areas: (a) compressing and indexing the outputs; and (b) extends the workflow by adding the steps to generate images for every locus it assesses and extracts a few metrics by processing the images, which will be used for training a filtering model. Specifically, this PR implements the following changes. 

- [x] Add [Illumina/REViewer](https://github.com/Illumina/REViewer) to STR docker file;
- [x] Fix a bug where the EH workflow fails if the optional PED file is not given. This was related to a bug in the command section where a variable inside `if` is evaluated even if the condition was not met.
- [x] _Sort_ and _index_ the realigned BAM file that EH generates. 
- [x] Compress (`gzip`) the TSV files containing alleles and variants. 
- [x] Add scripts to measure a few metrics from the SVG images REViewer generates. It would be ideal if EH would extract these metrics, and until EH is extended to output these metrics, @Nehir291 developed scripts (all the scripts under `src/str/` directory added in this PR) that extract the measurements from the images. The metrics are as the following. AlleleDepth; Total interrupting nucleotides upper; Median upper; Total interrupting nucleotides bottom; Median bottom; A upper Orange; T upper Orange; G upper Orange; C upper Orange; Total upper Orange; Median orange upper; Total read upper; Count Orange Upper reads with interruptions; A bottom Orange; T bottom Orange; G bottom Orange; C bottom Orange; Total bottom Orange; Median orange bottom; Total read bottom; Count Orange Bottom reads with interruptions. `Upper` and `bottom` refer to the haplotypes in the images. 
- [x] Adds a script that merges all metrics collected per site across batches in one file.
- [x] REViewer outputs a few three files per site (SVG, metrics.tsv, and phasing.tsv) which are needed to output. Outputing files separately will result in `Array[Array[Array[File]]]` that could contain millions of files per sample (e.g., running on catalog with `2M` sites, it will result in `3 * 2M` files per sample), which will be more files than Terra or Cromwell can store per variable. Therefore, the WDLs first put all the files of all the sites of a shard in a `.tar.gz` file, then combine all the `.tar.gz` of different shards into one `.tar.gz` file, resulting in one file per sample.


wdl/ExpansionHunter.wdl
```diff
@@ -278,3 +339,67 @@ task ConcatEHOutputs {
...
+        File? reviewer_images_gz = "reviewer_images.tar.gz"
+        File? sample_metrics = "${sample_id}_metrics.csv"
+    }
+
+    command <<<
+        set -euxo pipefail
+
+        REF="$(basename "~{reference_fasta}")"
+        mv ~{reference_fasta} $REF
+        mv ~{reference_fasta_index} $REF.fai
```
mwalker174, line 366: Usually we try to soft-link files rather than use `mv` in case the user is on a shared file system.

wdl/ExpansionHunter.wdl
```diff
@@ -75,6 +78,21 @@ workflow ExpansionHunter {
                 expansion_hunter_docker = expansion_hunter_docker,
                 runtime_override = runtime_eh
         }
+
+        if (generate_reviewer_images_ == true && generate_realigned_bam_ == true && generate_vcf_ == true) {
```
mwalker174, line 82:
```suggestion
        if (generate_reviewer_images_ && generate_realigned_bam_ && generate_vcf_) {
```

wdl/ExpansionHunter.wdl
```diff
@@ -251,6 +297,21 @@ task ConcatEHOutputs {
...
+            --input-filename ~{write_lines(sample_metrics_)} \
+            --metrics metrics.csv \
+            --missing-metrics missing_metrics.csv
+
+        # Combine multiple archives into one archive,
+        # by first extracting the files then re-archiving it.
+        # This allows preserving a flat structure with one archive,
+        # otherwise, it would create an archive containing archives.
+        mkdir temp_dir
+        cat ~{write_lines(reviewer_images_gz_)} | xargs -I{} tar -xzvf {} -C temp_dir/ && tar -czvf reviewer_images.tar.gz -C temp_dir/ .
```
mwalker174, line 314: I may be mistaken here, but I think you need to delete the extracted files in `temp_dir/` for each line in `reviewer_images_gz_`? It seems like the temp dir will accumulate files and keep re-tarring all previous files? You may need to use a loop here.

mwalker174: Actually in my experience, it's been fastest to just extract everything into one temp dir, then do one big `tar -c` to create a new archive rather than adding files incrementally.

wdl/ExpansionHunter.wdl
```diff
@@ -278,3 +339,67 @@ task ConcatEHOutputs {
...
+
+        python /opt/str/get_locus_metrics.py \
+            --sample-id ~{sample_id} \
+            --realigned-bam ~{realigned_bam} \
+            --reference $REF \
+            --genotypes genotypes.vcf \
+            --variant-catalog ~{variant_catalog_json} \
+            --output ~{sample_id}_metrics.csv
+
+        tar -czvf reviewer_images.tar.gz ~{sample_id}_*.svg
```
mwalker174, line 378:
```suggestion
        tar -czf reviewer_images.tar.gz ~{sample_id}_*.svg
```
To cut down on logs

wdl/ExpansionHunter.wdl
```diff
@@ -278,3 +339,67 @@ task ConcatEHOutputs {
...
+    RuntimeAttr runtime_default = object {
+        cpu_cores: 1,
+        mem_gb: 4,
+        boot_disk_gb: 10,
+        preemptible_tries: 3,
+        max_retries: 1,
+        disk_gb: 20 +
+            (2 * ceil(
+                size(realigned_bam, "GiB") +
+                size(vcf_gz, "GiB") +
```
mwalker174, line 390: I think your vcfs are small, but since you are decompressing it we usually use a decompression factor of 5-10 for vcfs, although the relationship is generally non-linear, particularly for multi-sample VCFs.

mwalker174 (approved):
Looks good, @VJalili. Understood that this is still a prototype, but it looks organized enough to go in. I just have a few suggestions for your WDLs.

wdl/ExpansionHunter.wdl
VJalili: That is a better alternative, thanks!

wdl/ExpansionHunter.wdl
VJalili: Thanks!

wdl/ExpansionHunter.wdl
```diff
@@ -215,12 +251,19 @@ task ConcatEHOutputs {
...
-        File variants_tsv = "${output_prefix}_variants.tsv"
-        File alleles_tsv = "${output_prefix}_alleles.tsv"
+        File variants_tsv_gz = "${output_prefix}_variants.tsv.gz"
+        File alleles_tsv_gz = "${output_prefix}_alleles.tsv.gz"
         File vcf_gz = "${output_prefix}.vcf.gz"
         File realigned_bam = "${output_prefix}.bam"
+        File realigned_bam_index = "${output_prefix}.bam.bai"
+        File reviewer_metrics = "metrics.csv"
+        File reviewer_missing_metrics = "missing_metrics.csv"
+        File reviewer_images_gz = "reviewer_images.tar.gz"
```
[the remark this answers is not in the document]
VJalili: Agree; that would make all the outputs named with a common prefix. I updated accordingly.

wdl/ExpansionHunter.wdl
VJalili: Good suggestion, thanks! I updated it accordingly.

wdl/ExpansionHunter.wdl
VJalili: Thanks!

wdl/ExpansionHunter.wdl
VJalili: That is a good point; I added the decompression factor `10` to be on the safer side.

wdl/ExpansionHunter.wdl
```diff
@@ -278,3 +339,67 @@ task ConcatEHOutputs {
...
+                size(reference_fasta, "GiB") +
+                size(variant_catalog_json, "GiB")))
+    }
+    RuntimeAttr runtime_attr = select_first([runtime_override, runtime_default])
+
+    runtime {
+        docker: expansion_hunter_docker
+        cpu: select_first([runtime_attr.cpu_cores, runtime_default.cpu_cores])
+        memory: select_first([runtime_attr.mem_gb, runtime_default.mem_gb]) + " GiB"
+        disks: "local-disk " + select_first([runtime_attr.disk_gb, runtime_default.disk_gb]) + " SSD"
```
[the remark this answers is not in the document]
VJalili: Good catch, I changed it to HDD; we do not need SSD for this.

PR #517, opened 2023-03-29
Inputs for MakeCohortVcf subworkflows

mwalker174:
- Adds input files to `ref_panel_1kg` test batch so that json files can be generated for `ResolveComplexVariants`, `GenotypeComplexVariants`, and `CleanVcf`.
- Removed unneeded square brackets for array inputs in the templates themselves.
- Adds corresponding outputs to `GATKSVPipelineBatch` so they can be updated in future runs (using `scripts/inputs/create_test_batch.py`).


epiercehoffman (approved):
Looks good - I tested the JSON build & validation on your branch and all the subworkflow JSONs were able to be built and validated by womtool. This will be nice to have done.

PR #518, opened 2023-03-29
Remove vcf header contig sorting in CleanVcf5

mwalker174:
Fixes an issue in CleanVcf5 where the use of `sort` results in re-ordering of contigs in the vcf header (i.e. to chr1, chr10, chr11, chr12, etc. instead of keeping chr1, chr2, chr3, …). This causes problems with GATK tools downstream in particular.

Also rewrites parts of the reheadering commands to be easier to understand.


cwhelan (approved):
This looks fine to me. It's a few passes through the header but it's so small it doesn't make much of a difference how you do it as long as it works. Would changing the original to a `sort -k1,1V` have worked? 

Also there is a GATK tool `UpdateVCFSequenceDictionary` I just remembered exists: https://gatk.broadinstitute.org/hc/en-us/articles/360037268111-UpdateVCFSequenceDictionary

I don't remember why we are stripping all of these lines from the VCF header here. 

This is fine as-is, though.

PR #519, opened 2023-03-31
Add sample renaming for SD files in GatherBatchEvidence

mwalker174:
- Adds optional renaming of sample IDs for BAF generation from SD files in GatherBatchEvidence. This is often useful when incorrect, invalid, or unsafe IDs were used during GatherSampleEvidence. This functionality exists for PE, SR, and RD files already.
- Rescale disk size in SDtoBAF
- Bump up max_retries to 1 in SDtoBaf, in case of transient failures
- Minor fix to disk scaling for MergeEvidence task

### DIFF
--- a/wdl/BatchEvidenceMerging.wdl
+++ b/wdl/BatchEvidenceMerging.wdl
@@ -102,7 +102,7 @@ task MergeEvidence {
 
   Float file_size = size(files, "GiB")
   Float subset_disk = if (subset_primary_contigs || rename_samples) then file_size else 0
-  Int disk_size = 10 + ceil(file_size * 2 + subset_disk)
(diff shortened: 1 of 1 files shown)


epiercehoffman (approved):
LGTM

shadizaheri:
My suggestion would be to hold off on merging this particular PR at the moment. Although the workflow appears to have been successful with the changes made, a couple of issues must be addressed. The `ClusterBatch` is currently failing and the `GenerateBatchMetrics` is also encountering issues with the `RDTestAllosome` step. It might be more effective to address the necessary fixes for these problems in a separate PR. In the meantime, documenting these issues within this PR would be a good idea.

Issue #432, opened 2022-10-04
Generate Terra workspace data tsv files from transposed tables


cwhelan:
See the suggestions in https://github.com/broadinstitute/gatk-sv/pull/413

This would make it much easier to edit, view diffs, and track changes in git, which is currently very cumbersome with this two-row wide tsv format.

As noted on the ticket above @epiercehoffman has also entered a feature request for Terra to change the format it expects this file in: https://support.terra.bio/hc/en-us/community/posts/9319574881819-Make-workspace-tsv-format-edit-friendly-git-friendly

shadizaheri:
This issue was resolved by merging [this pull request](https://github.com/broadinstitute/gatk-sv/pull/523).   Therefore, I go ahead and close this issue.

PR #523, opened 2023-04-04
Fix workspace format ( Generate Terra workspace tsv files from transposed tables)

shadizaheri:
**Description**
This pull request addresses the issue of workspace_tsv files being in a two-row-wide format.
The changes made in this PR will simplify editing, viewing diffs, and tracking changes in Git, as the current two-row-wide TSV format can be cumbersome and challenging to work with.

Initially, I transformed the `workspace.tsv.tmpl` file located in the templates folder into a two-column key-value format. Subsequently, I updated the `build_inputs.py` script to generate `workspace.tsv` files in the correct format for Terra, and save them in the appropriate directories within the build folder."

To test my changes, I produced workspace files on Terra and verified that they could be easily uploaded without any errors.

**Changes Made**
Transposed the `workspace.tsv.tmpl` file to key, value format with two columns
Modified the `build_inputs.py` scripts to output `workspace.tsv` files in the correct format
Tested workspace files on Terra

**Related Issues**
Successfully merging this pull request will resolve issue [#432](https://github.com/broadinstitute/gatk-sv/issues/432)."


scripts/inputs/build_inputs.py
```diff
@@ -165,9 +165,32 @@ def process_directory(input_dict, template_dir, target_directory):
...
+                groups[i].append(column)
+    # write the transposed data to a new file
+    with open(input_file, 'w') as f:
+        for i in sorted(groups.keys()):
+            f.write('\t'.join(groups[i]) + '\n')
+
+
 def process_file(input_dict, template_subdir, template_file, target_subdir):
+    # transpose files that starts with workspace for Terra processing
+    if template_file.startswith("workspace"):
```
cwhelan, line 190: I'd move this processing to after we check for the `.tmpl` extension

scripts/inputs/build_inputs.py
```diff
@@ -165,9 +165,32 @@ def process_directory(input_dict, template_dir, target_directory):
...
+        template_file_path = os.sep.join([template_subdir, template_file])
+        transpose_tsv(template_file_path, template_file_path)
```
cwhelan, line 192: I don't really love rewriting and then reverting the template files, which are tracked in git. (For example, if something goes wrong with the process people will end up with git differences that they weren't expecting.) What if you just rendered the non-transposed template, and then after writing out the processed file you transposed the result? That would leave an extra non-transposed output file in the build directory but that wouldn't be so bad. Or, you could transpose the file contents in-memory after processing the template but before writing out the contents, like:

```
processed_content = env.get_template(template_file).render(input_dict)
if template_file.startswith("workspace"):
   processed_content = transpose_tsv(processed_content)
```
You'd have to rework your `transpose_tsv` method to work on a string instead of file paths. But it might be cleaner and more efficient.

cwhelan (changes requested):
Looks good! 

I have a couple of suggestions of alternate ways that you could do this transpose without having to re-write and then revert a git-tracked file. 

Also, I'd be sure to add a comment to the documentation of `build_inputs.py` describing what happens with the transpose and what triggers it (having a template filename that  begins with `workspace`) since this behavior could be a little surprising to someone not expecting it.

scripts/inputs/build_inputs.py
shadizaheri: Thank you for your feed back, this request has been addressed [here](https://github.com/broadinstitute/gatk-sv/pull/523/commits/aabccf74c38db6c85598567a8ecb399cfdd96f7e).

scripts/inputs/build_inputs.py
shadizaheri: Thank you for your feed back, this request has been addressed [here](https://github.com/broadinstitute/gatk-sv/pull/523/commits/aabccf74c38db6c85598567a8ecb399cfdd96f7e).

shadizaheri:
@cwhelan Thank you for reviewing this PR and providing your valuable suggestions. I have addressed your suggestions in commits: [Modify the description of transpose tsv](https://github.com/broadinstitute/gatk-sv/pull/523/commits/653e444cfedb96f65fee046babf05b3dfd847081), [Avoid unnecessary re-writing](https://github.com/broadinstitute/gatk-sv/pull/523/commits/aabccf74c38db6c85598567a8ecb399cfdd96f7e), [Change transpose_tsv to accept a string input](https://github.com/broadinstitute/gatk-sv/pull/523/commits/1f1f899ec26e2e1ac88b064b8ca8ba1763c3536a), and [address the lint issue](https://github.com/broadinstitute/gatk-sv/pull/523/commits/83458591e891ef1e22257e7d3e61ade8cd579971). In addition, I have included the corresponding commit as a response to each requested change.

scripts/inputs/build_inputs.py
```diff
@@ -50,6 +50,10 @@
 # the run based on test_batch_small does not currently produce results for module03 and beyond because those files depend
 # resources not defined for the small batch (but defined for the large batch).
 #
+# The 'transpose_tsv' function simplifies the editing, viewing of diffs, and tracking of changes in Git for
```
cwhelan, line 53: I would just add an explicit sentence at the start of this paragraph describing the new behavior, something like "Any template files with a filename that begins with the string 'workspace' will be transposed before the final inputs are written. This functionality allows us to simplify the editing, viewing of diffs, and tracking of changes in Git of workspace TSV files that Terra expects in a two-row format."

cwhelan (changes requested):
Hi Shadi, looks good! I just have one minor comment about the functionality and a suggestion for how to make the script documentation clearer.

scripts/inputs/build_inputs.py
```diff
@@ -144,6 +148,29 @@ def main():
...
+            for i, column in enumerate(columns):
+                if i not in groups:
+                    groups[i] = []
+                groups[i].append(column)
+
+    # Transpose the groups and write the result to a new string
+    transposed_groups = []
+    for i in sorted(groups.keys()):
+        transposed_groups.append('\t'.join(groups[i]))
+    transposed_str = '\n'.join(transposed_groups)
```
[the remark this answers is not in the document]
shadizaheri: Hi Chris! Thank you for your comment! I addressed this [here](https://github.com/broadinstitute/gatk-sv/pull/523/commits/1c51a9d08bf8e64785ed2e488b2987a5c35f0b49). However, there is no extra line at the end of the new version of the new `workspace.tsv.tmpl` which is the transposed version of the previous one (please see [here](https://github.com/broadinstitute/gatk-sv/pull/523/commits/459a64425d56000dc0a984475e9f037962428d52#diff-4f3b416ffc1d0ba83f662aa4f54eabf363b8a069c990217384dfe9603eaec4b3) ). I tried testing the final generated `workspace.tsv` file for the cohort mode by uploading the `workspace.tsv` to the Terra workspace, and it uploaded the information correctly.

scripts/inputs/build_inputs.py
shadizaheri: @cwhelan Thank you for your comment! Following your suggestion, I updated the description in [this commit](https://github.com/broadinstitute/gatk-sv/pull/523/commits/40565a248f0df80ae2968945be8b0fe31149c817).

cwhelan (approved):
Great, looks good to merge now!

Issue #520, opened 2023-03-31
IntegrateGQ error in GenotypePESRPart2 in single sample pipeline


cwhelan:
A user has run into an error in in the single sample pipeline, version 0.26.9-beta, involving a failure of one shard of the `IntegrateGQUnder5kb` task in `GenotypePESRPart2`. The task failed without any error messages written to the log file. This appears to be a rare issue as many other samples were processed successfully by the same version of the pipeline by the same user.

Further investigation shows that this is due to a very small shard on chromosome Y containing only one variant, which is causing grep statements within the script to fail to find any matches and return failure exit codes that break their bash pipelines.

cwhelan:
The errors occur in `src/sv-pipeline/04_variant_resolution/scripts/IntegrateGQ.sh`

mwalker174:
Addressed in #524

PR #524, opened 2023-04-07
Fix for tiny shard of IntegrateGQ in single sample pipeline

cwhelan:
Fixes an error in processing with the single sample pipeline in the IntegrateGQUnder5Kb task of GenotypePESRPart2: https://github.com/broadinstitute/gatk-sv/issues/520

The shard being processed contained only one variant (on chrY) which was causing a `grep` statement to fail to find any matches and return a failure exit code, which broke the bash pipeline due to `pipefail` and caused the script to exit.

I've wrapped the grep statements in the script with `{ || true }` to cause them to not fail even if they don't find any matches. 

The user who encountered the error reports tested with a patched docker image and reports that this caused the sample to be processed successfully. She also re-processed 9 other samples that had previously been run successfully with the new docker and says that the results were identical to the previous version.

### DIFF
--- a/src/sv-pipeline/04_variant_resolution/scripts/IntegrateGQ.sh
+++ b/src/sv-pipeline/04_variant_resolution/scripts/IntegrateGQ.sh
@@ -33,14 +33,14 @@ zcat $RD_melted_genotypes \
 ##Deletions, need to PE-SR genotypes to match RD format (2==ref)##
 ##PE##
 zcat $pegeno_indiv_file \
-  |fgrep -wf <(awk '{if ($5=="DEL") print $4}' int.bed) || true  \
+  | { fgrep -wf <(awk '{if ($5=="DEL") print $4}' int.bed) || true; }  \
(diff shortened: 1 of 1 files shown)


VJalili:
Docker build fails building melt, which is expected given the licensing issues. However, that failure will cause dockers.json not to be updated with the latest images, so the referenced image(s) will not contain the updated script.

mwalker174 (approved):
Thank you @cwhelan this looks fine to me. I've been trying to rewrite `grep -f` statements with `awk` instead. FYI the safest way to do this I've used in BatchEvidenceMerging:

```
awk -F'\t' -v OFS='\t' -v SAMPLE="$sample" '!second_file{chroms[$1]; next} {~{if subset_primary_contigs then "if ($1 in chroms)" else ""} print ~{if rename_samples then "$1,$2,$3,SAMPLE" else "$0"} }' contigs.list second_file=1 -
```
which doesn't rely on `NR==FNR` (breaks with empty files) or using `ARGIND` (isn't supported by all versions of awk).

PR #525, opened 2023-04-13
Remove non-public images from the git-sha-based target determination

VJalili:
This PR excludes the non-public docker images (`melt` specifically`) from the git-sha-based target images determination. 

These images can still be built if listed explicitly via the `--targets` argument.

### DIFF
--- a/scripts/docker/build_docker.py
+++ b/scripts/docker/build_docker.py
@@ -323,7 +323,7 @@ class ProjectBuilder:
                         for image, dependencies in self.dependencies.items()
                         if dependencies.depends_on(new_targets_to_build) and
                         not ImageBuilder(image, self).do_not_rebuild
-                    }.difference(targets_to_build)
+                    }.difference(targets_to_build, self.non_public_images)
                 )
 
         # noinspection PyTypeChecker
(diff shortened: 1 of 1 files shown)


VJalili:
Thank you for reviewing this, @cwhelan! I agree printing a warning message would be helpful. I updated accordingly.

cwhelan (approved):
LGTM now.

VJalili:
I updated the script to the non-public images (`melt` specifically) that are also skipped from the dependency tree. Also, I added print statements that indicate the images are omitted. 

One way to test this would be the following command, and you can check if the non-pub images are to be built by considering the shown print statement. 

```
$ python scripts/docker/build_docker.py --targets sv-base-virtual-env --disable-git-protect
...
Building the following targets in order:
['sv-base-virtual-env', 'cnmops-virtual-env', 'sv-pipeline-virtual-env', 'sv-base', 'cnmops', 'sv-pipeline']
```

Note that `melt` is not included in the above list.

cwhelan:
Looks good to me.

VJalili:
Thanks, @cwhelan!

PR #527, opened 2023-04-27
Fix issue in Genotype batch script with low PE_log_pval causing zero …

shadizaheri:
This commit addresses a problem where batches with low PE_log_pval were resulting in PE_count of zero, which could cause issues. To fix this, the commit sets the minimum count value to max(1, count-1), ensuring that a count of at least 1 is always returned.

The changes made in this PR were tested using the batch that was holding a PE_count = 0, which was causing the GenotypePESRPart1 task to fail. The result of the test was that the task succeeded, confirming that the fix is effective and resolving the issue. A screenshot of the test result is attached for reference.

Screen Shot 2023-05-02 at 8 24 21 AM

### DIFF
--- a/src/sv-pipeline/04_variant_resolution/scripts/convert_poisson_p.py
+++ b/src/sv-pipeline/04_variant_resolution/scripts/convert_poisson_p.py
@@ -22,7 +22,7 @@ def convert_poisson_p(log_pval):
         pval_cmp = -np.log10(ss.poisson.cdf(0, count))
 
         if pval_cmp > log_pval:
-            return count - 1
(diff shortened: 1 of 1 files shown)


shadizaheri:
@epiercehoffman and @mwalker174  thank you for your feedback. The whole test succeeded ([here](https://app.terra.bio/#workspaces/allofus-drc-wgs-sv-prod/AoU_DRC_WGS_GATK-SV-Phase2/job_history/7e440e69-1e32-40f1-a90b-aded731f37f5)).

mwalker174 (approved):

PR #529, opened 2023-05-09
make the RunMELT task a little more robust

tedsharpe:
I won't say that this is any less obscure, but at least some of the obscurity is documented.

### DIFF
--- a/wdl/MELT.wdl
+++ b/wdl/MELT.wdl
@@ -496,30 +496,15 @@ task RunMELT {
 
     # create a bam file from the input bam or cram, filtering out high coverage intervals
     # and ambiguous bases
-
-    # use a fifo as the output, because PrintReads doesn't know how to write to stdout
-    mkfifo tmp.sam
     java -Xmx~{java_mem_mb}m -jar /opt/gatk.jar PrintReads \
-      -XL highCountIntervals.bed \
-      --interval-exclusion-padding ~{interval_padding} \
-      -I ~{bam_or_cram_file} \
-      -R ~{reference_fasta} \
-      -O tmp.sam &
-
-    # a harmless command that will eventually open and close the fifo without writing anything
-    # this is a failsafe so that the awk command that follows will not hang forever in the
-    #   event that the previous command fails without opening its output file
-    # the amount of sleeping isn't critical -- it has to be long enough that PrintReads will have
-    #   opened its output file (if it's ever going to do so)
-    ( sleep 60; cat /dev/null > tmp.sam ) &
-
-    # read from the fifo and replace ambiguous bases with 'N'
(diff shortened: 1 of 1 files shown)


wdl/MELT.wdl
```diff
@@ -496,15 +496,31 @@ task RunMELT {
...
+    #   opened its output file (if it's ever going to do so)
+    ( sleep 60; cat /dev/null > tmp.sam ) &
+
+    # read from the fifo and replace ambiguous bases with 'N'
+    awk 'BEGIN{FS=OFS="\t"}{gsub(/[BDHVRYKMSW]/,"N",$10);print}' < tmp.sam | \
       samtools view -b1 > ~{sample_id}.bam
+
+    # the wait command's return code is the return code of the PrintReads command
+    # so if PrintReads fails, the script will fail
+    wait %1
```
mwalker174, line 522: This is a slight concern - could you wait on the specific PID for PrintReads instead? I'm not very familiar with shell job management but Cromwell executes these command sections by basically copy+pasting inside a larger script. If Cromwell (future versions eg Cromwell on Azure) decides to add a background process first, could that cause a problem here?

wdl/MELT.wdl
```diff
@@ -496,15 +496,31 @@ task RunMELT {
...
       -R ~{reference_fasta} \
       -O tmp.sam &
-    awk 'BEGIN{FS=OFS="\t"}{gsub(/[BDHVRYKMSW]/,"N",$10);print}' tmp.sam | \
+
+    # a harmless command that will eventually open and close the fifo without writing anything
+    # this is a failsafe so that the awk command that follows will not hang forever in the
+    #   event that the previous command fails without opening its output file
+    # the amount of sleeping isn't critical -- it has to be long enough that PrintReads will have
+    #   opened its output file (if it's ever going to do so)
+    ( sleep 60; cat /dev/null > tmp.sam ) &
```
mwalker174, line 514: So if PrintReads fails immediately, I can see how this will work. But if PrintReads runs for >60 seconds, successfully opens the output stream, but fails afterwards, we are still going to miss the error right?

wdl/MELT.wdl
tedsharpe: Yeah.  I think that's a valid concern.  I'll try to figure out a workaround.

tedsharpe:
Giving up on the fifo.  It was clever and goofy at the same time.  Adopting Chris' suggestion.  A little extra compression and decompression, but much more straightforward.

tedsharpe:
This has now been tested.  (Sample size = 1.)

mwalker174 (approved):
Thanks @tedsharpe LGTM

PR #534, opened 2023-05-11
Manually install MASS R package in sv-pipeline-virtual-env

mwalker174:
The latest MASS package now requires a newer version of R and is breaking docker builds. This PR manually installs MASS with a previous version that is compatible with our version.


VJalili (approved):
LGTM, thank you!!

PR #535, opened 2023-05-11
Update copy_outputs.py

mwalker174:
This utility script was very out of date and hard-coded to work with specific workflows. This PR simplifies it to copy down all output files from a Cromwell metadata into a flat GCR location.


VJalili (approved):
The updated code reads much cleaner and more concise; thank you!! 

I think a good follow-up PR on this could be abstracting the backend-specific code, in order to make it applicable to a wider range of backends, such as Azure.

PR #539, opened 2023-05-23
Improve STR workflow

VJalili:
This PR improves the STR workflow in the following areas: 

- [x] Add a flag to enable including _all_ the ExpansionHunter generated metrics when converting its JSON output to a TSV file using a utility script; defaulted to `false` (i.e., not including);
- [x] Improve on generating metrics from the REViewer images: 
   - Bug fix in the utility script addressing a case where the file containing the list of image files may have invalid file names (e.g., blank space or lines containing only white space). This bug is fixed by considering only lines referencing a valid file.  
   - Simplify the interface so that only a prefix is provided instead of separate output filenames.
   - Run REViewer-related steps only if the `generate_reviewer_images` flag is set; otherwise, generate empty files for the workflow expected outputs.

### DIFF
--- a/src/str/merge_csv_files.py
+++ b/src/str/merge_csv_files.py
@@ -1,10 +1,12 @@
 import argparse
+import os.path
 import pandas as pd
 
 
 def merge_files(input_filename: str, metrics_filename: str, missing_metrics_filename: str):
     with open(input_filename, "r") as fin:
-        files = [line.rstrip() for line in fin]
+        lines = fin.read().splitlines()
+        files = [x for x in lines if os.path.isfile(x)]
 
     dfs = []
     for filename in files:
--- a/src/str/merge_csv_files.py
+++ b/src/str/merge_csv_files.py
@@ -6,7 +6,12 @@ import pandas as pd
 def merge_files(input_filename: str, metrics_filename: str, missing_metrics_filename: str):
     with open(input_filename, "r") as fin:
         lines = fin.read().splitlines()
-        files = [x for x in lines if os.path.isfile(x)]
+        files = []
+        for x in lines:
+            if os.path.isfile(x):
+                files.append(x)
+            else:
+                print(f"Skipping the invalid filename `{x}`.")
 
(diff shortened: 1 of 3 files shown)


wdl/ExpansionHunter.wdl
```diff
@@ -200,7 +205,9 @@ task RunExpansionHunter {
             touch ~{sample_id}.vcf.gz
         fi
 
-        python /opt/str/combine_expansion_hunter_json_to_tsv.py -o ~{sample_id} ~{sample_id}.json
+        python /opt/str/combine_expansion_hunter_json_to_tsv.py \
+            ~{if (defined(include_all_fields)) then "--include-all-fields " else ""} \
```
mwalker174, line 209:
```suggestion
            ~{if include_all_fields then "--include-all-fields " else ""} \
```
Is what I think you want, or 
```suggestion
            ~{true="--include-all-fields" false="" include_all_fields} \
```

src/str/merge_csv_files.py
```diff
@@ -1,10 +1,12 @@
 import argparse
+import os.path
 import pandas as pd
 
 
 def merge_files(input_filename: str, metrics_filename: str, missing_metrics_filename: str):
     with open(input_filename, "r") as fin:
-        files = [line.rstrip() for line in fin]
+        lines = fin.read().splitlines()
+        files = [x for x in lines if os.path.isfile(x)]
```
mwalker174, line 9: I think you'd want to emit an error or at least a warning if one of the paths is not found (thinking of the case where there's a typo in a path)

mwalker174 (changes requested):
Looks good! Just a couple of comments.

src/str/merge_csv_files.py
VJalili: Good suggestion!! I updated to code accordingly.

VJalili:
Thanks, @mwalker174, for the prompt review!!

mwalker174 (approved):

PR #548, opened 2023-06-13
Remove DeleteIntermediaFiles task.

VJalili:
The `DeleteIntermediateFiles` uses `gsutil` to delete intermediate files, which is not portable and not compatible with Cromwell on Azure. As per discussion with @mwalker174, we decided to delete this task.

I tested the workflow after deleting this task, and the submission finished successfully. The submission id is `bf0093fd-9952-4af9-bccf-507f7a54e8a6`


VJalili:
Thank you, @mwalker174, for a prompt review!!

PR #549, opened 2023-06-14
Add high-level description of docker images and their build process

VJalili:
This PR extends the website by adding documentation about the docker images, their relationship, the motivation behind building them into multiple images, and a high-level description of their build, test, and publish process. 

The documentation on building, pushing, developing, and updating CI/CD for other container registries will be added in follow-up PRs.


website/static/img/docker_hierarchy.png
mwalker174, line 1: This is great! Can you label the latest release in one of the corners, just in case we update the hierarchy in the future and forget to update this figure.

website/docs/docker/images.md
```diff
@@ -0,0 +1,91 @@
...
+GATK-SV pipeline, written in WDL, are organized into separate Docker 
+containers. This modular approach ensures that each container 
+contains only the necessary tools for its specific task, 
+resulting in smaller image sizes. This design choice simplifies 
+the definition of Dockerfiles and facilitates easier maintenance. 
+Moreover, the smaller image sizes contribute to reduced disk 
+usage and lower workflow execution costs.
+
+
+The figure below illustrates the relationships between the GATK-SV Docker images.
```
mwalker174, line 27: I think you need to explain the figure a bit. It wouldn't be clear to a naive reader what the arrows mean.

website/docs/docker/deploy/manual.md
```diff
@@ -0,0 +1,16 @@
...
+
+If you are contributing to the GATK-SV codebase, specifically focusing on 
+enhancing tools, configuring dependencies in Dockerfiles, or modifying GATK-SV scripts 
+within the Docker images, it is important to build and test the Docker images locally. 
+This ensures that the images are successfully built and function as intended. 
+Additionally, if you wish to host the images in your own container registry, 
+you will need to follow these steps. 
+To simplify the build process, we have developed a Python script 
+that automates the image building, and publishing to your container registry. 
+This section provides a detailed guide on using this script.
```
mwalker174, line 15: Looks like this needs to be added still?

website/docs/docker/index.md
```diff
@@ -0,0 +1,66 @@
...
+  sources={{
+    light: useBaseUrl('/img/docker_infra_diagram.png'),
+    dark: useBaseUrl('/img/docker_infra_diagram.png'),
+  }}
+/>
+
+
+The GATK-SV Docker setup is organized as follows:
+
+ - **Dockerfile**: There are 21 Dockerfiles available in the GitHub repository.
```
mwalker174, line 48:
```suggestion
 - **Dockerfile**: 
```
I don't think it's necessary to cite the specific number here. Also it could change and then we'd need to remember to change it here.

mwalker174 (changes requested):
Thank you for putting this together. The docker system has become a bit complex over the years and this adds a lot of clarity as to how the build system works.

I have a few minor suggestions and also it looks like the manual docker publishing section is missing some content. 

In addition, the documentation is very technical and I think we want to make it clear that this is intended for developers and not users (who we don't want to scare away ;)). One way to do that would be place the "Docker Images" as a subsection under a "Development" or "For Developers" section. Not sure how that will look visually since there's already a few levels of subsections, so you could try to just add a clear disclaimer to the Overview.

On a related note, I think it would be helpful to have a small section under "Getting Started" about dockers that is intended for users. A quick description of where to find `dockers.json` and that the images there are updated with workflows and other code.

website/docs/docker/images.md
VJalili: Agree; I added a brief description.

website/docs/docker/deploy/manual.md
VJalili: I removed the last sentence since I will extend this file by adding the related guides in a follow-up PR.

website/static/img/docker_hierarchy.png
VJalili: Following our offline discussion, we planned on adopting a website versioning approach that can bind the docs in their entirety to a particular release. We discussed enabling this feature as part of our release strategy planning.

VJalili:
Thanks for the review, @mwalker174! I implemented the suggested changes; I agree it reads better now; thank you! I would appreciate it if you could re-review it.

website/docs/advanced/docker/images.md
```diff
@@ -0,0 +1,99 @@
...
+sidebar_position: 1
+---
+
+import useBaseUrl from '@docusaurus/useBaseUrl';
+import ThemedImage from '@theme/ThemedImage';
+
+:::info
+This page provides a detailed explanation of Docker 
+images and their hierarchy. For information on the process 
+of building these images, please refer to [this section](/docs/advanced/docker/deploy/index).
```
mwalker174, line 13: Link's broken

website/docs/advanced/docker/deploy/manual.md
```diff
@@ -0,0 +1,16 @@
...
+
+If you are contributing to the GATK-SV codebase, specifically focusing on 
+enhancing tools, configuring dependencies in Dockerfiles, or modifying GATK-SV scripts 
+within the Docker images, it is important to build and test the Docker images locally. 
+This ensures that the images are successfully built and function as intended. 
+Additionally, if you wish to host the images in your own container registry, 
+you will need to follow these steps. 
+To simplify the build process, we have developed a Python script 
+that automates the image building, and publishing to your container registry. 
+This section provides a detailed guide on using this script.
```
mwalker174, line 15: Looks like this sentence is still here?

mwalker174 (approved):
Thanks! Just a couple last comments but on the whole good to go.

website/docs/advanced/docker/images.md
VJalili: Thanks! I fixed it.

website/docs/advanced/docker/deploy/manual.md
VJalili: Thanks! I think it is because the commit was not pushed.

VJalili:
Thanks for the review, @mwalker174!

PR #551, opened 2023-06-16
Add support for incomplete workflows in get_inputs_outputs.py

mwalker174:
Quick update to a utility script that traverses a Cromwell metadata file and generates input/output jsons for all subworkflows. 

In particular, this update enables the script to emit the input parameters to incomplete/failed workflows, which is sometimes useful (before the workflow needed both inputs and outputs to be available). 

Also some minor formatting changes.


epiercehoffman (approved):
Looks good. These are some nice dictionary comprehensions to replace the awkward copying code

PR #552, opened 2023-06-16
Remove max GQ filtering from per sample QC

mwalker174:
Removes the upper-bound on GQ from the MainVcfQc per-sample analysis script. There were a few bugs with the implementation and it was just easier to remove this constraint since it isn't practically necessary.

Note that the max-gq parameter is still used for GQ binning.


epiercehoffman (approved):
This seems fine to me

PR #553, opened 2023-06-16
Update SR counting and genotyping

mwalker174:
- Fixes svtk sr-test and count-sr code to allow for `-+` in addition to `+-` strandedness on INS variants. The sr-test strategy also was updated so that it is more likely to find a significant SR signature on each side, and now defaults to the nearest significant position rather than one 50bp away. These changes result in substantially more two-sided support for INS variants, as well as more accurate breakpoint refinement for small variants, particularly insertions.
- Updates SR genotyping model. First, INS are now genotyped separately from other balanced variants, as they exhibit a distinctive SR count distribution. Second, separate model parameters are now used for one-sided and two-sided variants. The two models are also applied based on whether the particular sample exhibits one- or two-sided support, rather than the variant as a whole. Third, the SR GQ model has been simplified to compute a p-value of the left-hand tail based on the count z-score. It therefore scores the confidence that the genotype is not over-called rather than that the genotype is correct (i.e. using a two-sided test).

The combination of these updates substantially improves INS genotyping and alleviates some of the het bias of the previous model in all SR-supported variants.

This branch has been tested on the 1kgp reference panel as well as batch_23 from AoU-PhaseI.

### DIFF
--- a/src/sv-pipeline/04_variant_resolution/scripts/SR_genotype.opt_part2.sh
+++ b/src/sv-pipeline/04_variant_resolution/scripts/SR_genotype.opt_part2.sh
@@ -3,7 +3,7 @@
 # SR_genotype.sh
 #
 
-set -euxo pipefail
+set -euo pipefail
 
 ##Genotype##
 function assign_genotype_quality () {
--- a/src/sv-pipeline/04_variant_resolution/scripts/SR_genotype.opt_part2.sh
+++ b/src/sv-pipeline/04_variant_resolution/scripts/SR_genotype.opt_part2.sh
@@ -3,14 +3,114 @@
 # SR_genotype.sh
 #
 
+set -euxo pipefail
 
-set -euo pipefail
+##Genotype##
+function assign_genotype_quality () {
+
+  # Arguments
+  INFILE=$1
+  VIDS_LIST=$2
+  MEDIAN_HET=$3
+  STDEV=$4
+  GQ_FILE=$5
+  VARQ_FILE=$6
+
+  ##normalization factor##
+  ##Normalize phred scores to 999 if count is the mean  ##
+  normalization=$(Rscript -e "print(((pnorm(0) )))"  \
+                    | tr '\n' '\t' \
+                    | awk '{print 999/($NF)}')
+  ##null genotype get max quality score##
+  zcat $INFILE \
+    | awk '{if ($NF==0 && $3==0) print $0 "\t" 999}' \
+    | gzip \
+    > null.geno.txt.gz
+
+  ##null genotype but has SR reads determined by poisson test##
+  if [ $(zcat $INFILE|awk '{if ($NF==0 && $3>0) print}'|wc -l) -gt 0 ]
+  then
+    zcat $INFILE|awk '{if ($NF==0 && $3>0) print}' \
+      | Rscript -e 'd<-read.table("stdin")' \
+      -e "z<-cbind(d[1],d[,2],d[,3],d[,4],round(matrix(apply(d[,3,drop=F],1, function (x) (1-ppois(0, lambda=x))* $normalization) ,ncol=1)) )" \
+      -e 'write.table(z,"null.wreads.geno.txt",col.names=FALSE,quote=FALSE,row.names=FALSE,sep = "\t")'
+  else
+    echo "" > null.wreads.geno.txt
+  fi
+
+  ##genotype based on z score for observation with genotypes##
+  if [ $(zcat $INFILE|awk '{if ($NF==1) print}'|wc -l) -gt 0 ]
+  then
+    zcat $INFILE|awk '{if ($NF==1) print}' \
+      | Rscript -e 'd<-read.table("stdin")' \
+      -e "z<-apply(d[,3,drop=F]-$MEDIAN_HET,1, function (x) ((pnorm(x/$STDEV) )))" \
+      -e "zfinal<-round(z* $normalization)" \
+      -e 'write.table(cbind(d[1],d[,2],d[,3],d[,4],zfinal),"wreads.geno.het.txt",col.names=FALSE,quote=FALSE,row.names=FALSE,sep = "\t")'
+  else
+    echo "" > wreads.geno.het.txt
+  fi
+  if [ $(zcat $INFILE|awk '{if ($NF>1) print}'|wc -l) -gt 0 ]
+  then
+    zcat $INFILE|awk '{if ($NF>1) print}' \
+      | Rscript -e 'd<-read.table("stdin")' \
+      -e "z<-apply(d[,3,drop=F]-2*$MEDIAN_HET,1, function (x) ((pnorm(x/$STDEV) )))" \
+      -e "zfinal<-round(z* $normalization)" \
+      -e 'write.table(cbind(d[1],d[,2],d[,3],d[,4],zfinal),"wreads.geno.hom.txt",col.names=FALSE,quote=FALSE,row.names=FALSE,sep = "\t")'
+  else
+    echo "" > wreads.geno.hom.txt
+  fi
+
+  cat wreads.geno.het.txt wreads.geno.hom.txt null.wreads.geno.txt <(zcat null.geno.txt.gz) \
+    | sed '/^$/d' \
+    | awk '{if ($NF<0) print $1,$2,$3,$4,1;else if ($NF>999) print $1,$2,$3,$4,999 ;else print}' \
+    | tr ' ' '\t' \
+    | sort -k1,1 -k2,2 \
+    | gzip \
+    > $GQ_FILE
+
+  ##Per variant quality scores##
+  ##Assign a normalization factor to scale variant up to 999 if based on expected het median ##
+  normalization_var=$(Rscript -e "print(-10*log10(ppois(0, lambda=$MEDIAN_HET)))"  \
+                    | tr '\n' '\t' \
+                    | awk '{print 999/($NF)}')
+
+  if [ $(zcat $INFILE|awk '{if ($NF>0) print}'|wc -l) -gt 0 ]
+  then
+    zcat $INFILE|awk '{if ($NF>0) print}' \
+      | Rscript -e 'd<-read.table("stdin")' \
+      -e 'x<-tapply(d[,3],d[,1],median)' \
+      -e 'z<-cbind(names(x),(-10*log10(ppois(0, lambda=x))))' \
+      -e 'write.table(z,"sr.variant.quality.final.txt",col.names=FALSE,quote=FALSE,row.names=FALSE,sep = "\t")'
+  else
+    echo "" > sr.variant.quality.final.txt
+  fi
+
+  ##add back variants with no SR support##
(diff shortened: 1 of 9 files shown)


mwalker174:
Genotype-filtered VCF and QC are available here: `gs://broad-dsde-methods-markw/mw-ins-sr-strands/FilterGenotypes/outputs/`

src/sv-pipeline/04_variant_resolution/scripts/SR_genotype.opt_part2.sh
```diff
@@ -25,188 +125,175 @@ rare_both=$(awk '{if($1=="rare_both") print $2}' $metric_file)
...
+  | awk '{print $4}' \
+  | sort -u \
+  | awk -F'\t' -v OFS='\t' 'ARGIND==1{keys[$1]; next} { if ($1 in keys) print }' pass.srtest.txt - \
+  | awk -F'\t' -v OFS='\t' 'ARGIND==1{keys[$1]; next} { if ($1 in keys) print }' <(cat recover.txt recover.bothsides.txt|awk '{print $1}'|sort -u) - \
+  > background.variant.fail.txt
 
 ##Pull out variants to genotype##
 join -j 1 -a 1 -a 2  -e "." -o 1.1 2.1 1.2 1.3 1.4 1.5 2.2 2.3 2.4 2.5 \
   <(zcat sr.geno.final.txt.gz \
-  |fgrep -wf both.pass.txt \
```
epiercehoffman, line 124: Does this mean that we used to null out SR GTs if the SR background was too high but we are no longer doing so?

src/svtk/svtk/pesr/sr_test.py
```diff
@@ -9,59 +9,61 @@
...
-                rec = rec + 1
+            record.stop, record.info['STRANDS'][1], record.info['CHR2'],
+            called, background, record.stop - self.window, record.stop + self.window)
+
+        posA = int(resultA.pos)
+        posB = int(resultB.pos)
+        log_pval_A = float(resultA.log_pval)
+        log_pval_B = float(resultB.log_pval)
+        if record.info['SVTYPE'] == 'INS':
+            if posA > posB + self.ins_window or posA < posB - self.ins_window:
```
epiercehoffman, line 40: Does this mean that posA and posB must be within self.ins_window of each other for INS? What about INS with a deletion at the insertion site, or a sequence of microhomology > 50bp?

src/svtk/svtk/pesr/sr_test.py
```diff
@@ -9,59 +9,61 @@
...
+                    resultA = self._test_coord(
+                        posB, record.info['STRANDS'][0], record.chrom,
+                        called, background, posB - self.ins_window, posB + self.ins_window)
+        elif record.chrom == record.info['CHR2'] and posA >= posB:
+            # Invalid coordinates, need to re-optimize around the better coordinate
+            if log_pval_A >= log_pval_B:
+                # posA is better, so use posA as anchor and check for best posB in valid window
+                resultB = self._test_coord(
+                    record.stop, record.info['STRANDS'][1], record.info['CHR2'],
+                    called, background, posA, posA + self.window)
```
epiercehoffman, line 58: Does this mean that posA could be returned as the optimal 2nd coordinate because it is in the test window?

src/svtk/svtk/pesr/sr_test.py
```diff
@@ -9,59 +9,61 @@
...
+                if log_pval_A >= log_pval_B:
+                    # posA is better, so use posA as anchor and check for best posB in valid window
+                    resultB = self._test_coord(
+                        posA, record.info['STRANDS'][1], record.info['CHR2'],
+                        called, background, posA - self.ins_window, posA + self.ins_window)
+                else:
+                    # vice versa
+                    resultA = self._test_coord(
+                        posB, record.info['STRANDS'][0], record.chrom,
+                        called, background, posB - self.ins_window, posB + self.ins_window)
```
epiercehoffman, line 51: Could this return a pair of coordinates that don't meet the requirements you are imposing? It seems that after being updated once they are not checked again

epiercehoffman (commented):
Thanks for this fix, this is a massive amount of work and should be a big improvement. The QC is looking good. The new code you've added looks clean, and I especially appreciate the added documentation. 

I wasn't very familiar with a lot of this code so although I spent a good deal of time reviewing, I still have some questions about your changes - hopefully not too hard to answer and then we can move forward!

src/sv-pipeline/04_variant_resolution/scripts/SR_genotype.opt_part2.sh
```diff
@@ -3,14 +3,114 @@
 # SR_genotype.sh
 #
 
+set -euxo pipefail
```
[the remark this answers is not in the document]
mwalker174: Not much of an issue unless there's a bash loop somewhere. I'll still remove it to reduce storage a little though.

src/sv-pipeline/04_variant_resolution/scripts/SR_genotype.opt_part2.sh
```diff
@@ -25,188 +125,175 @@ rare_both=$(awk '{if($1=="rare_both") print $2}' $metric_file)
...
+  | tr ' ' '\t' \
+  | gzip \
+  > sr.geno.final.txt.gz
 
 zcat ${SR_sum} \
   | awk '{print $0 "\t" $1"@"$2}' \
   | cut -f1-3 \
-  | awk -v var=$sr_count -v var1=$median_hom -v var2=$sd_het '{if ($3<var) print $1,$2,$3,0;else if ($3<=var1-var2) print $1,$2,$3,1; else print $1,$2,$3,int($3/(var1/2)+0.5)}' \
-  |tr ' ' '\t' \
+  | awk -v var=$sr_count -v hom_cutoff=$oneside_hom_cutoff -v median_het=$oneside_median_het '{if ($3<var) print $1,$2,$3,0;else if ($3<=hom_cutoff) print $1,$2,$3,1; else print $1,$2,$3,(int(($3/median_het)+0.5) < 2) ? 2 : int(($3/median_het)+0.5)}'  \
```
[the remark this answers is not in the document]
mwalker174: Yes you can have GT>2 for other types, which can happen for example if the event is nested within a duplication. I believe these cases are handled in CleanVcf.

src/sv-pipeline/04_variant_resolution/scripts/SR_genotype.opt_part2.sh
```diff
@@ -25,188 +125,175 @@ rare_both=$(awk '{if($1=="rare_both") print $2}' $metric_file)
...
+  |awk '{print $1"@"$2"\t"$0}'|sort -k1,1) \
+  |awk '{if ($1!="." && $5) print $1,$3,$4,$5,$6,$9,$10; else print $2,$7,$8,$5,$6,$9,$10}' \
+  |awk '{if ($4>0) print $1,$2,$3,$4,$5; else print $1,$2,$3,$6,$7}' \
   |gzip>combine.int.txt.gz
+# Output columns:
+# 1) VID@Sample (passing variants only)
+# 2) VID
+# 3) Sample
+# 4) SR count (prioritizing two-sided if its two-sided genotype is non-ref)
+# 5) Genotype (prioritizing two-sided if its two-sided genotype is non-ref)
```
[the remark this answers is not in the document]
mwalker174: Yeah this is a bit confusing, but a genotype will always be non-ref if it has two-sided support, so you can't have a two-sided hom-ref.

src/sv-pipeline/04_variant_resolution/scripts/SR_genotype.opt_part2.sh
```diff
@@ -25,188 +125,175 @@ rare_both=$(awk '{if($1=="rare_both") print $2}' $metric_file)
...
+
+# Reformat
+zcat combine.int.txt.gz | awk '{print $2,$3,$4,$5}' |tr ' ' '\t'|gzip>sr.geno.withfinal.filter.txt.gz
```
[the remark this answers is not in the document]
mwalker174: Fair point. I guess there's a slight gain by not having to compress extra text and read/write that column though.

src/svtk/svtk/pesr/sr_test.py
mwalker174: Yes they could be larger, but I was attempting to match the intended behavior of the previous implementation which uses 50bp. Larger window sizes could increase fall positives, e.g. in messy regions you'd be more likely to find significant SR signals by chance. The microhomologous sequences are most frequently ~20bp but they can be larger. This is probably a point worth discussing so let's follow up.

src/svtk/svtk/pesr/sr_test.py
mwalker174: It does allow for posB < posA, but that will get fixed later in `rewrite_SR_coords.py`. Is there another case you're thinking of?

epiercehoffman: Yes, I guess it's just posA > posB. To clarify, why is that considered invalid on line 52 if it is fixed in rewrite_SR_coords.py?

src/svtk/svtk/pesr/sr_test.py
mwalker174: Yes I think there's a rare corner case here for intrachromosomal variants with `++` and `--` strandedness (INV and some BNDs) that are small (less than twice the window size). In this case with an overlapping window, if the best SR position is within that overlapping window, the algorithm will choose posA=posB, which shouldn't be valid for this variant class. 

Luckily this seems to be a rare case, at least in the 1kgp reference panel there were only 2, and there were no differences in the variants between runs on main and in this branch.

I'm still going to add a check for this case, specifically by rerunning `_test_coord` for posB but explicitly disallowing it to test at posA.

src/sv-pipeline/04_variant_resolution/scripts/SR_genotype.opt_part2.sh
mwalker174: This has more to do with two-sided support than SR background. 

It is a bit complicated, but basically there are two models for genotyping: a two-sided and a one-sided. Before, it would select the model based on whether the site was deemed one- or two-sided. The methodology for flagging a site as two-sided was fine for the purposes of just assigning a two-side support status to the variant (by checking what fraction of non-ref samples have two-sided support, above). 

However, the method also used to genotype according to the site-level both-side support status. It turns out that within a single site, there can be a mixture of two-side support among samples, and the cutoff method was not good enough. This wasn't an issue before since it happened that the one- and two-sided models were actually identical. But now that the one-sided model has different cutoffs, this approach was causing havoc in a lot of variants.

Here I have just removed the requirement that two-sided genotypes have two-sided status at the site level.

src/svtk/svtk/pesr/sr_test.py
mwalker174: Ah, the difference here is the window size. In the first test we use `self.window` but here we only test over `self.ins_window`, which could be smaller.

epiercehoffman (approved):
Thanks for the updates and explanations! Approved pending decision about the INS breakpoint window

PR #566, opened 2023-07-10
Sharded AnnotateVcf workflow

epiercehoffman:
### Updates

This PR incorporates @xuefzhao's changes to AnnotateVcf to shard the workflow for additional scaling capabilities and add the PAR BED file for AF annotation, as well as additional changes to streamline and clean up the AnnotateVcf workflow. Changes include:
* Shard input VCF(s) for AnnotateVcf for scalability
* Pass PAR BED file to AF annotation task
* **UNDID AFTER REVIEW**: Allow user to provide either a single full input VCF or a list of contig VCFs, and the output will match the input in this regard. **INSTEAD**: Input is a single VCF, and to run separately on subsets of contigs the user should submit multiple workflows with different contig_list inputs.
* **UNDID AFTER REVIEW**: Make index file an optional input. **INSTEAD**: Require presence of index but not input; infer path based on VCF path
* Remove some scatter-gather operations in AnnotateVcf subworkflows that are no longer necessary (there is one scatter at the beginning & one gather at the end of the workflow now)
* Remove some intermediary WDLs in AnnotateVcf subworkflows that are no longer necessary
* Use latest utility tasks for sharding, concatenating, subsetting samples, etc.
* Add set -euo pipefail in some tasks that were missing it
* Change runtime_override to runtime_attr in some tasks to match the rest of the code base

### Testing
* Validated all WDLs and JSONs with womtool & terra validation script
* Tested AnnotateVcf.wdl with ref panel data, with both a full VCF input & contig VCFs
  * The output truncates the floats (to 6 places) for the external AF values compared to the latest main branch, so it is not identical, but that appears to be the only difference
  * The contig VCFs and full VCF output were not comparable because the inputs were from slightly different versions of the reference panel VCF
  * Both the latest main branch & this branch took about 85 mins to run on a single full VCF input, but performance gains are expected for larger input VCFs


inputs/templates/terra_workspaces/cohort_mode/workflow_configurations/AnnotateVcf.SingleBatch.json.tmpl
```diff
@@ -1,6 +1,6 @@
 {
-  "AnnotateVcf.vcf" : "${this.cleaned_vcf}",
-  "AnnotateVcf.vcf_idx" : "${this.cleaned_vcf_index}",
+  "AnnotateVcf.vcf_list" : "${this.cleaned_vcf}",
+  "AnnotateVcf.vcf_idx_list" : "${this.cleaned_vcf_index}",
```
mwalker174, line 3: Are these array types in the workspace?

wdl/AnnotateVcf.wdl
```diff
@@ -1,113 +1,149 @@
...
 workflow AnnotateVcf {
 
   input {
-    File vcf
-    File vcf_idx
+    Array[File] vcf_list  # Must be either single full VCF (array of length 1) or array of VCFs sharded by contig. Outputs will match
+    Array[File]? vcf_idx_list
     File contig_list
-    String prefix
+    Array[String] prefix_list
```
mwalker174, line 14: This design works but it requires some complicated code below. I think it would be better to forego the assumption that it's every sharded by contig. Looking through, the contig is only used for file prefixes, so we could try something a little more general that would be simpler to implement.

I'd suggest dropping the `contig_list` input and using `basename` on the VCFs to generate the `prefix_list`. If it's sharded by contig, we'll still get the contig labels on all the outputs. This way we don't need to worry about whether we're in the sharded or unsharded case nor whether the contig list length matches the array length.

wdl/AnnotateVcf.wdl
```diff
@@ -1,113 +1,149 @@
...
+
     String sv_pipeline_docker
+    String? sv_pipeline_hail_docker
+    String sv_base_mini_docker
     String gatk_docker
 
+    File? NONE_FILE_
+
     RuntimeAttr? runtime_attr_svannotate
     RuntimeAttr? runtime_attr_concat_vcfs
```
mwalker174, line 44: Unused

wdl/TasksMakeCohortVcf.wdl
```diff
@@ -990,6 +990,7 @@ task ScatterVcf {
 
   command <<<
     set -euo pipefail
+    ~{if !defined(vcf_index) then "tabix ~{vcf}" else ""}
```
mwalker174, line 993: This is problematic if the vcf is unsorted. Can you make an explicit input like `Boolean generate_index_if_unavailable` that defaults to false but is also required to be true here?

mwalker174 (changes requested):
Thanks for this. Unless I'm missing something, I think things can be simplified a bit though - see comments below.

inputs/templates/terra_workspaces/cohort_mode/workflow_configurations/AnnotateVcf.SingleBatch.json.tmpl
epiercehoffman: No, but Terra is able to handle conversion from a File to an Array[File] in inputs

wdl/TasksMakeCohortVcf.wdl
epiercehoffman: Removing this line and requiring index in AnnotateVcf as discussed

epiercehoffman:
Addressed review comments and switched to a single input VCF as discussed to simplify the WDL structure. If the user wants to run a subset of contigs they should provide a subsetted contig_list, and they can submit multiple workflows for multiple subsets of contigs. Re-tested: womtool validation & successful run of AnnotateVcf.wdl on the reference panel.

mwalker174 (approved):

PR #569, opened 2023-07-17
Reference panel resources and critical bug fixes

mwalker174:
- Updates some 1KG reference panel to the proper resource bucket
- Add CHR2 to all records and force 'N' ref alleles and symbolic BND alt alleles in gatk-to-svtk svtk formatting script
- Update GATK docker with SVCluster BND allele support
- Increase default memory given to JVM for gCNV tasks

The `GATKSVPipelineBatch` workflows was successfully tested on this branch. The single-sample workflow still fails `UpdateBreakendRepresentation` due to a separate issue with BND `END` fields that will be addressed in an upcoming PR.


wdl/GermlineCNVCase.wdl
```diff
@@ -227,7 +227,7 @@ task DetermineGermlineContigPloidyCaseMode {
     RuntimeAttr runtime_attr = select_first([runtime_attr_override, default_attr])
 
     Float mem_gb = select_first([runtime_attr.mem_gb, default_attr.mem_gb])
-    Int command_mem_mb = ceil(mem_gb * 1000 - 500)
+    Int command_mem_mb = ceil(mem_gb * 1000 * 0.6)
```
VJalili, line 230: You can also do `ceil(mem_gb * 600)`.

src/sv-pipeline/scripts/format_gatk_vcf_for_svtk.py
```diff
@@ -94,14 +94,26 @@ def convert(record: pysam.VariantRecord,
...
     # copy INFO fields
     for key in record.info:
         if key not in remove_infos:
             new_record.info[key] = record.info[key]
-    # fix END, CHR2, SVLEN, STRANDS
+    # svtk generally expects all records to have CHR2 assigned
+    chr2 = record.info.get('CHR2', None)
+    if chr2 is None:
+        new_record.info['CHR2'] = contig
+    # fix END, SVLEN, STRANDS
```
VJalili, line 116: Would it not be more correct to check `new_record.info` than `record.info`?

You can also write the check like the following, but both are functionally equivalent. 

```python
if "CHR2" not in record.info:
    new_record.info["CHR2"] = contig
```

VJalili (approved):
LGTM, thank you, @mwalker174! Hopefully, replacing `- 500` with `* 0.6` will fix the out-of-memory issues you identified. I also checked a few of the updated lists of files, and all those I checked were accessible. I also left a couple of comments, since they will have no functional impact, please feel free to leave your code unchanged.

wdl/GermlineCNVCase.wdl
mwalker174: Yeah that would be shorter, but I'd prefer to leave it this way since IMO it's a little clearer where the factor of 1000 comes from.

src/sv-pipeline/scripts/format_gatk_vcf_for_svtk.py
mwalker174: I did think about whether to use `new_record` or `record` for a moment, but I just thought this was a little safer in case the key-copying code above gets moved/deleted/changed.

I think we have to use the current logic as it's written. In VCFs, the field may exist but still be empty, i.e. if it's `CHR2=.`. In that case I believe `CHR2` in `record.info` will actually still return true but `record.info['CHR2']` will still return `None` in pysam.

VJalili: That makes sense; thanks for elaborating on it!

PR #573, opened 2023-07-26
Remove use of the file command

mwalker174:
A user ran into an error in `gather_cpx_intervals_for_rd_gt.sh` [here](https://github.com/broadinstitute/gatk-sv/blob/12e4562578179ca3eaf0ee8557acb258e2853960/src/sv-pipeline/04_variant_resolution/scripts/gather_cpx_intervals_for_rd_gt.sh#L67) where the `file` command returned `DOS/MBR boot sector, COM/EXE Bootloaderd???v` instead of the expected `gzip`. 

The `file` command is generally not reliable and its usage in the code base to check whether files are gzipped is largely unnecessary. I've removed all instances of it I could find by grepping for `file` and `grep` in the same line.

### DIFF
--- a/src/sv-pipeline/scripts/vcf_qc/compare_callsets.sh
+++ b/src/sv-pipeline/scripts/vcf_qc/compare_callsets.sh
@@ -117,8 +117,7 @@ fi
 ###PREP INPUT FILES
 OVRTMP=`mktemp -d`
 #Unzip SET1, if gzipped, restrict to contigs in $CONTIGS, rewrite non-conventional SVTYPES, and automatically set SVs with size <1 to 1
-if [ $( file ${SET1} | fgrep " gzip " | wc -l ) -gt 0 ] || \
-   [ $( echo ${SET1} | awk -v FS="." '{ if ($NF ~ /gz|bgz/) print "TRUE" }' ) ]; then
+if [ $( echo ${SET1} | awk -v FS="." '{ if ($NF ~ /gz|bgz/) print "TRUE" }' ) ]; then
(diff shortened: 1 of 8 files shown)


VJalili:
If `file` has proven problematic, I think it is safe to remove it as you did here, since it is already mentioned in the usage help that the file needs to be a bgzipped, and also if it receives a non-compliant file, zcat with throw an exception. So, I agree an additional check given the unreliability of `file` can be avoided. Thanks for fixing this!!

src/sv-pipeline/scripts/vcf_qc/compare_callsets.sh
```diff
@@ -141,8 +140,7 @@ if [ ${CARRIER} == 1 ]; then
   mv ${OVRTMP}/set1.bed2 ${OVRTMP}/set1.bed
 fi
 #Unzip & format SET2, if gzipped
-if [ $( file ${SET2} | fgrep " gzip " | wc -l ) -gt 0 ] || \
-   [ $( echo ${SET2} | awk -v FS="." '{ if ($NF ~ /gz|bgz/) print "TRUE" }' ) ]; then
+if [ $( echo ${SET2} | awk -v FS="." '{ if ($NF ~ /gz|bgz/) print "TRUE" }' ) ]; then
```
VJalili, line 143: This check is less "stringent" than using `file` since the updated code will pass if you had only renamed the file and added `.gz` extension to it, while the previous checks the file header which is more reliable. However, given the issues caused by `file`, I think it is still reasonable to check only the extension, but treat with a bit of caution :)

PR #574, opened 2023-07-27
Bug fix building samtools-cloud docker

VJalili:
Fixes https://github.com/broadinstitute/gatk-sv/issues/571
This PR updates GCP SDK installation steps to adhere to the latest recommendations [on GCP docs](https://cloud.google.com/sdk/docs/install#deb). Specifically, using `--keyring` to store/register GCP public keys.

PR #575, opened 2023-07-28
Bug fix Github actions out of disk error

VJalili:
Github runners currently have ~15 GB disk space, and we run out of disk when building some of the base Docker images, as it will result in rebuilding many derived images. This PR adds the `--prune-after-each-image` to `build_docker` invocation to decrease disk usage. This flag was used in the `Test Images Build` job but missing from the `Publish` job, which is a bug.  

The inconsequential change on the `samtools-cloud/Dockerfile` is to trigger building and publishing `samtools-cloud` docker image and all the derived images since the `Publish` job had failed on #574, and the images were not published.


VJalili:
Thanks for the prompt review, @mwalker174.

> Is there any way to increase the runner disk size?

There are [larger runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-larger-runners) under the paid plans; we can discuss how these can be useful moving forward as we currently have little headroom in disk size on current runners.

Issue #394, opened 2022-08-29
Duplicate INV records


mwalker174:
@xuefzhao reported a handful of duplicated INV records in the "cleaned" vcf. I'd speculate that these the duplicates are being derived from BND records in ResolveComplexVariants, but this needs to be confirmed.

mwalker174:
The culprit appears to be `IntegrateResolvedVcfs` in `ResolveComplexVariants.wdl`.

ekarlins:
I'm seeing examples of this for other svTpes as well. CPX, specifically. It seems like the duplicate records will have slight differences in the FILTER field or slightly different EVIDENCE. Here are a few examples:

```
[...]

PR #576, opened 2023-08-01
Deduplicate records and CPX variant IDs in ResolveComplexVariants

mwalker174:
- Addresses #394, which was caused by a bug in the `IntegrateResolvedVcfs` task. This task is now implemented as a python script that is much more efficient and readable.
- Reimplements CPX variant ID generation, which previously used a random number generator to get unique IDs. However, there was a bug where the same seed was used twice for the RNG (once for the first pass and once for the second pass of the algorithm), resulting in CPX variants with duplicated IDs. `svtk resolve` now uses the prefix argument, which in the WDL now includes the contig and shard index. This guarantees unique variant IDs and avoids the use of RNGs.

This was tested against a recent run of `MakeCohortVcf` by comparing the MainVcfQc figures. Note that there will be half as many INV events and 40% fewer CPX due to deduplication.

### DIFF
--- a/src/sv-pipeline/04_variant_resolution/scripts/integrate_resolved_vcfs.py
+++ b/src/sv-pipeline/04_variant_resolution/scripts/integrate_resolved_vcfs.py
@@ -1,0 +1,107 @@
+#!/bin/env python
+
+import argparse
+import sys
+from typing import Optional, List, Text, Dict, Set, Callable
+
+import pysam
+
+MEMBERS_KEY = "MEMBERS"
+UNRESOLVED_KEY = "UNRESOLVED"
+
+
+def update_header(header: pysam.VariantHeader) -> None:
+    header.add_line('##INFO=<ID=UNRESOLVED,Number=0,Type=Flag,Description="Variant is unresolved.">')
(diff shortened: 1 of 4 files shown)


src/sv-pipeline/04_variant_resolution/scripts/integrate_resolved_vcfs.py
```diff
@@ -0,0 +1,107 @@
...
+        if r.id not in all_unresolved_vids_set or r.id not in inv_resolved_members_set:
+            # Resolved in ALL vcf, or unresolved in both VCFs
+            sys.stdout.write(str(r))
+        else:
+            sys.stderr.write(f"all excluded: {r.id}\n")
+    for r in inv_vcf:
+        if r.id in inv_resolved_vids_set:
+            # Resolved variant in the INV vcf
+            members = get_members(r)
+            sys.stderr.write(f"inv considered: {r.id}\n")
```
epiercehoffman, line 62: Do these logging lines create a large stderr, or are INV records rare enough that it's not an issue?

epiercehoffman (approved):
Looks good to me!

src/sv-pipeline/04_variant_resolution/scripts/integrate_resolved_vcfs.py
mwalker174: Accidentally left those in

Issue #578, opened 2023-08-04
Index file conventions in WDLs


mwalker174:
Many files types such as VCFs, CRAMS, BED, and evidence data files have companion index files that are sometimes required for fast retrieval over specific genomic intervals.

### Overview:
Some WDLs require that the file be passed in explicitly. For example:
```
workflow {
  inputs {
    File vcf
    File vcf_index
  }
}
```
In some cases, its path is inferred from the file itself. For example,
```
inputs {
  File vcf
}
…
File vcf_index = vcf + ".tbi"
```
And in other cases, the index file may not be declared at all.

### Considerations:
There are several different scenarios, depending on a number of factors, that affect the decision about how to handle the indexes:

1. Whether declaring workflow inputs or task inputs.
2. Whether or not the index is required for at least one tool or command (e.g. GATK requires VCF indexes, but `zcat` does not)
3. Whether the file is localized or not. If it's not localized, then a tool will be streaming the data directly from cloud storage. In this case, most tools implicitly assume the presence of the index at the path `${FILE_URI}.tbi`, and therefore an explicit declaration of the index is not needed.

Other design factors to consider:

4. Whether the file is a fixed resource (e.g. a reference annotation intervals list), a user-provided file (e.g. CRAM), or an intermediate file generated by the pipeline.
5. Passing around index file paths can bloat the WDLs, input JSONs, and workflow metadata.
7. Explicit index file inputs makes it clear to users that the file must be present (particularly important for user-provided files).
6. Generating an index in some instance is not possible. For example, if the file is not sorted.
8. Generating indexes "on the fly" wastes compute, assumes the file is sorted, and may cause problems with certain backends, such as on shared filesystems.
9. In rare cases (usually for user-provided inputs), the index file will be located in a different path from the main file. In these cases, it is necessary to move both files into the same directory (e.g. using `cp`, `mv`, or `ln`) unless the particular tool being used supports passing in the index path explicitly (most don't). Usually we use `cp` or `ln`, as `mv` can create problems on shared filesystems.

### Proposal:
For the sake of development and UX, we should enforce consistent conventions regarding companion index files in the WDLs.

I'd propose the following rules to balance the above issues:

**Task-level inputs:**
1. If **the index is needed**, then the index should be a required input. This is true regardless of whether the file is localized, as file with `localization_optional=true` will still localize files if unsupported by the backend. 
[...]

PR #579, opened 2023-08-04
Add required index files to hg38 resources

mwalker174:
Lists all required index files for fixed pipeline resource files. Partially addresses #578.


VJalili (approved):
This will indeed improve the UX, in particular when mirroring the required resources on different cloud platforms or different zones. Thank you, @mwalker174!

PR #580, opened 2023-08-04
Resolve new linting errors

mwalker174:
There's possibly been an update to the linting libraries that is causing errors. See #579 for an example.


VJalili:
I saw this in two PRs, thanks for fixing it!!

VJalili (approved):

PR #588, opened 2023-08-22
Update JoinRawCalls formatter arg in template

mwalker174:
JoinRawCalls has been failing due to improper assignment of `END2` during GATK-style formatting. This PR includes 2 small updates:
- Adds a needed "--fix-end" argument to the GATK formatter
- Removes unused parameter `preprocess_args`

### DIFF
--- a/inputs/templates/test/JoinRawCalls/JoinRawCalls.json.tmpl
+++ b/inputs/templates/test/JoinRawCalls/JoinRawCalls.json.tmpl
@@ -15,6 +15,8 @@
   "JoinRawCalls.clustered_melt_vcfs" : [{{ test_batch.merged_melt_vcf | tojson }}],
   "JoinRawCalls.clustered_melt_vcf_indexes" : [{{ test_batch.merged_melt_vcf_index | tojson }}],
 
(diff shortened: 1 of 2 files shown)


inputs/templates/test/JoinRawCalls/JoinRawCalls.json.tmpl
```diff
@@ -15,6 +15,8 @@
   "JoinRawCalls.clustered_melt_vcfs" : [{{ test_batch.merged_melt_vcf | tojson }}],
   "JoinRawCalls.clustered_melt_vcf_indexes" : [{{ test_batch.merged_melt_vcf_index | tojson }}],
 
+  "JoinRawCalls.FormatVcfForGatk.formatter_args": "--fix-end",
```
epiercehoffman, line 18: Does it also need `--scale-down-gq` since these are the raw calls that haven't had ends/GQ fixed in CleanVcf? That's what I see in the filtering instructions document as well

mwalker174: No these are vcfs from `ClusterBatch` so there are no GQs

epiercehoffman (approved):

PR #590, opened 2023-08-25
Filter mCNVs under 5kbp

mwalker174:
Adds a filter to the end of `CleanVcf` that removes multiallelic CNVs under 5kbp, as these calls are generally unreliable.

Tested successfully on the reference panel.


epiercehoffman (approved):
Looks good to me, and I'd like to get this change in before we need it for phase 2

PR #591, opened 2023-08-28
Remove redundant GetPed task in GatherBatchEvidence.CNMOPS

kirtanav98:
The GetPed task should be removed was removed from CNMOPS now that SubsetPedFile is run earlier in the workflow for GatherBatchEvidence. This was validated using womtool and successfully tested on the 1-kg reference data using cromwell.


epiercehoffman (approved):
Looks good to me! Thanks for removing this redundant code - it will reduce the size of the code base that we need to maintain and prevent wasting time on unnecessary compute.

PR #592, opened 2023-08-29
Update melt docker image to current latest build.

VJalili:
Given the license requirements of MELT, this image is not built and pushed to container registries using the GitHub actions. Hence, the latest builds of this image should be manually kept in synch between GCR and ACR. The image currently referenced for ACR is an older build than the image on GCR, and it does not include samtools. This PR updates the Docker image tag of MELT on ACR to the current latest tag on GCR.


mwalker174 (approved):
Thanks, looks like we forgot this in #468.

VJalili:
Thanks @mwalker174!

PR #593, opened 2023-08-30
Make genome_tracks optional in FilterGenotypes

MattWellie:
Current WDLs don't pass genome_tracks through from FilterGenotypes. Calling FilterGenotypes creates a workflow [call to RecalibrateGq](https://github.com/broadinstitute/gatk-sv/blob/main/wdl/FilterGenotypes.wdl#L41). This workflow invocation has neither `genome_tracks` or `recalibrate_gq_args` as attributes.

On the receiving side [only one](https://github.com/broadinstitute/gatk-sv/blob/main/wdl/RecalibrateGq.wdl#L12-L14) of those arguments is marked as optional, the construction of the command to run implies these are both optional.

---

Secondary issue from FilterGenotypes: MainVcfQc requires `primary_contigs_fai` to be passed through, which is not currently possible. 've added this as an argument.

Both of these arguments are present in the [test template](https://github.com/broadinstitute/gatk-sv/blob/main/inputs/templates/test/FilterGenotypes/FilterGenotypes.fixed_cutoffs.json.tmpl#L8-L18), but there's no way to pass them in at runtime.

Apologies if this is all misguided!

### DIFF
--- a/inputs/templates/test/FilterGenotypes/FilterGenotypes.fixed_cutoffs.json.tmpl
+++ b/inputs/templates/test/FilterGenotypes/FilterGenotypes.fixed_cutoffs.json.tmpl
@@ -5,8 +5,8 @@
   "FilterGenotypes.gq_recalibrator_model_file": {{ reference_resources.aou_recalibrate_gq_model_file | tojson }},
   "FilterGenotypes.sl_filter_args": "--small-del-threshold 93 --medium-del-threshold 150 --small-dup-threshold -51 --medium-dup-threshold -4 --ins-threshold -13 --inv-threshold -19",
 
-  "FilterGenotypes.RecalibrateGq.genome_tracks": {{ reference_resources.recalibrate_gq_genome_tracks | tojson }},
-  "FilterGenotypes.RecalibrateGq.recalibrate_gq_args": [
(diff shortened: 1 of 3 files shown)


mwalker174 (changes requested):
Thanks @MattWellie,

We've been providing those inputs as nested parameters, e.g. `FilterGenotypes.RecalibrateGq.genome_tracks` [here](https://github.com/broadinstitute/gatk-sv/blob/4950ac601f40a8595bd4476dbd0995c5ff25a93d/inputs/templates/test/FilterGenotypes/FilterGenotypes.fixed_cutoffs.json.tmpl#L8).

Perfectly okay with these changes, but please update the corresponding template linked above to provide the fasta index directly to `FilterGenotypes.wdl`.

MattWellie:
Thanks for the update @mwalker174! I wasn't familiar with that nested passing of inputs (I'm a WDL n00b), so I wasn't aware that this was a valid way of passing arguments.

I've since found out that we are able to support it with our internal pipeline's way of wrapping Cromwell workflows, but for this PR I've completed the removal of nested argument passing (`MainVcfQc` and `RecalibrateGq`) to reflect the explicit nature of all the other WDLs.

mwalker174 (approved):
Thanks! Looks good!

MattWellie:
Thanks for approving @mwalker174, I don't have write permission to hit merge, could you please action that when you have a minute?

PR #594, opened 2023-08-30
Remove hgdp_1kgp_ped test batch input from MainVcfQc template

mwalker174:
Fixes a minor error in `inputs/templates/test/MainVcfQc/MainVcfQc.per_sample_benchmark.json.tmpl` where HGDP was referenced instead of using the test batch's name/ped file.


epiercehoffman (approved):
Looks good to me. I noticed that the HGDP PED file exists in two locations - `gs://gatk-sv-hgdp/mw-sv-concordance-update/HGDP_1KGP.ped` and `gs://gatk-sv-resources-public/hg38/v0/sv-resources/resources/v1/HGDP_1KGP.ped`. This could be a good opportunity to pick one and reduce redundancy as well.

PR #595, opened 2023-08-30
Fix melt insert size input type in EvidenceQC

mwalker174:
Ran into an error running on Terra when using Scramble instead of MELT. In this case, `melt_insert_size` isn't populated, but instead of passing a `none` value it passes an empty array. This PR fixes  to `EvidenceQC` to effectively treat the empty array as `none`.

Tested in Terra successfully on a Scramble run-through.


epiercehoffman (approved):
Looks good to me, thanks for the fix!

PR #599, opened 2023-09-07
Fix allosomes_list bug

MattWellie:
This optional argument is parsed incorrectly if the file is supplied


mwalker174 (approved):
Thank you @MattWellie, I was surprised we hadn't caught this but it looks like we have been omitting this input in the past.

PR #600, opened 2023-09-07
Toggle Collect PESR/Counts in GatherSampleEvidence with booleans

epiercehoffman:
### Updates
Setting `collect_coverage` or `collect_pesr` to `false` in GatherSampleEvidence should disable CollectCounts and CollectSVEvidence (except if MELT is enabled, which requires CollectCounts), but this setting was ignored if `run_module_metrics` was set to `true`. This is unnecessary because every task in the metrics workflow is optional, and the metrics workflow should only collect metrics on the desired outputs. This PR removes the requirement to run CollectSVEvidence and CollectCounts just for module metrics.

### Testing
* Validated all WDLs and JSONs with womtool and terra validation script
* Ran GatherSampleEvidence on one sample with `collect_pesr = false` and `collect_coverage = false` and only the `wham_docker` provided and verified that only LocalizeReads, Wham, and Wham metrics tasks ran


mwalker174 (approved):
Looks good!

PR #601, opened 2023-09-11
Update manta to 1.6.0

mwalker174:
- Upgrades to the final release of manta, touting improved parallelization performance (both GCP and Azure dockers)
- Updates raw manta vcfs for the 1kgp reference panel
- Reduces default memory usage for manta

These changes were tested in `GatherSampleEvidenceBatch` (average cost $0.23 per sample). Outputs were highly similar. After running the rest of the pipeline through CleanVcf, final QC and benchmarks were nearly identical to those of the current main branch.


epiercehoffman (approved):
Thanks for the updates. The tests you performed sound thorough enough to me. 

Some questions for my own understanding: 
* Do we expect the ubuntu upgrade to change the behavior at all?
* The memory reduction seems substantial. Do we expect that memory to be sufficient for messier samples?

mwalker174:
> * Do we expect the ubuntu upgrade to change the behavior at all?

Should not affect behavior, but I upgraded since 18.04 is no longer in the standard period of its [support cycle](https://ubuntu.com/about/release-cycle).

> * The memory reduction seems substantial. Do we expect that memory to be sufficient for messier samples?

It may not be, but the memory was more than enough for the reference panel. I'd prefer to reduce the default since data quality has been increasing over time. We can increase again if we see high failure rates, but again there were big balancing improvements in this version.

Issue #365, opened 2022-06-30
Is MELT_SCRIPT published?


c2997108:
I am trying to build a private docker image for MELT.wdl, but I cannot find the MELT_SCRIPT.
I downloaded MEMT v2.2.2 and got MELT.jar, but run_MELT*.sh is not included.
MELT_SCRIPT is described in MELT.wdl as following.
MELT_SCRIPT=$(ls "$MELT_DIR/run_MELT"*.sh)

Is the run_MELT*.sh available somewhere?

c2997108:
Thanks for the quick response.
I was able to build the MELT docker container.
I will test it now.

Changing MELT to MEI is attractive. I look forward to seeing it implemented.

c2997108:
I could run wdl/MELT.wdl, but I got the following error.
`/MELT/run_MELT_2.2.2.sh: line 84: /cromwell_root/transposon_reference.list: No such file or directory`

Then I modified the line 608 as the following and the job seemed to be finished with no error.
```
<     CROMWELL_ROOT="/cromwell_root"
---
>     CROMWELL_ROOT="$PWD" #"/cromwell_root"
```

PR #602, opened 2023-09-12
Make the Cromwell root config in the MELT workflow portable

VJalili:
The root directory of the VM started by Cromwell to run a task is a Cromwell instance configuration. The default value is set to `/cromwell_root` and `/cromwell-executions` on GCP and Azure, respectively.  

[MELT script](https://github.com/broadinstitute/gatk-sv/blob/main/dockerfiles/melt/run_MELT_2.0.5.sh) takes the root directory as an input, and MELT's WDL [hardcodes this input to `/cromwell_root`](https://github.com/broadinstitute/gatk-sv/blob/8c9ddab01ab19b86f33afc01f5149645aeee0fa7/wdl/MELT.wdl#L511-L513), which works as expected on GCP. However, the hardcoded value is not portable. 

This PR extends the hardcoded value to the default Cromwell root directories to GCP and Azure, and uses `$PWD` (discussed in https://github.com/broadinstitute/gatk-sv/issues/365#issuecomment-1172809000) as a fallback option if the Cromwell root was set to a different value. 

The changes are successfully tested on GCP and Azure. The GCP workflow id is: `f450a355-031f-44fc-8289-d199d0fd93e8`.

### DIFF
--- a/wdl/MELT.wdl
+++ b/wdl/MELT.wdl
@@ -510,7 +510,24 @@ task RunMELT {
 
     # these locations should be stable
     MELT_DIR="/MELT"
-    CROMWELL_ROOT="$PWD"
+
+    # `cromwell_root` and `cromwell-executions` are the **default**
+    # root directory for cromwell deployments on GCP and Azure respectively.
+    # The last option, i.e., $PWD, is the fall back option if the
+    # default root directory of the cromwell instance was changed,
+    # however, it is not a reliable option as it fails on GCP.
+    # It does not seem Cromwell sets a runtime environment variable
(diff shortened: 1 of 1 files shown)


mwalker174 (approved):
Thanks @VJalili this looks fine. Would it be sufficient to just use `CROMWELL_ROOT="$PWD"` though?

VJalili:
Thank you, Mark! Setting to `CROMWELL_ROOT="$PWD"` would have been simpler; I tried it, but it fails on GCP on the Cromwell instances I used.

PR #603, opened 2023-09-12
Optimize Scramble

mwalker174:
This PR incorporates several updates to the Scramble tool and workflow that substantially improve performance and reduce costs on GCP.

- The Scramble WDL is now split into two parts: (1) the `cluster_identifier` tool that scans the BAM/CRAM for clusters of split reads, and (2) `cluster_analysis` that calls insertion MEIs from the clusters. Note that (2) requires substantially more compute and memory and is now parallelized (see below).
- A Scramble Dockerfile has been added. This builds the docker from a forked Github repository at [markw/scramble-gatk-sv](https://github.com/mwalker174/scramble-gatk-sv) that implements two new features: (1) the ability to override the reference used for CRAM decoding in `cluster_identifier`, and (2) parallelization of the `cluster_analysis` R script. Importantly, (1) prevents htslib from automatically downloading the reference from the EBI web server, which causes major issues at scale. This Dockerfile also loads updated versions of htslib libraries and bcftools relative to the previous Scramble docker released by the developers.
- Scramble now only calls on primary contigs to reduce computation, as calls on all other contigs are filtered downstream.
- Updates raw Scramble vcfs for the 1KGP reference panel.

Final cost estimates are forthcoming but estimated at <$0.10 per sample. We should experience fewer failures and long-running outliers in the future.

Identical output was confirmed on the 1KGP reference panel (on primary contigs).


epiercehoffman (approved):
This looks good to me. I didn't review your fork of scramble but your explanations and the code in this branch make sense. Really excited for the improved cost and stability!!!

VJalili:
Thank you, @mwalker174! This looks very good, and I'm excited to see we're getting closer to a better Scramble integration that can replace the restrictive alternative.

Out of curiosity, is there any reason the scramble docker was not added to [`build_docker.py`](https://github.com/broadinstitute/gatk-sv/blob/c3153eab94cc0657974c489cbd84300ccbe47eb3/scripts/docker/build_docker.py#L170-L244)?

PR #605, opened 2023-10-04
Setup automatic updates to WDLs on Dockstore

VJalili:
This PR adds configurations necessary to enable automatic updates of the WDLs on Dockstore. 
Accordingly, this PR adds `.dockstore.yml`, which registers the workflows defined on GATK-SV Terra workspace (i.e., `[gnomad-v3-sv/gatk-sv-pipeline-1kgp](https://app.terra.bio/#workspaces/gnomad-v3-sv/gatk-sv-pipeline-1kgp-test-noaddress)`) for an automatic sync with Dockstore. 

Only the root workflows need to be registered for automatic sync, and Dockstore will fetch all the dependencies automatically without a need for explicit registration. Additionally, the dockstore app does not support registering workflows using wildcards; hence, each workflow must be registered separately. 

After this PR is merged, dockstore app will automatically sync the registered workflows with dockstore with every commit pushed to any branch on broadinstitute/gatk-sv that has `.github/.dockstore.yml` file in it.

### DIFF
--- a/.github/.dockstore.yml
+++ b/.github/.dockstore.yml
@@ -1,0 +1,73 @@
+version: 1.2
+workflows:
+  - subclass: WDL
+    name: 01-GatherSampleEvidence
+    primaryDescriptorPath: /wdl/GatherSampleEvidence.wdl
+
+  - subclass: WDL
+    name: 02-EvidenceQC
+    primaryDescriptorPath: /wdl/EvidenceQC.wdl
+
+  - subclass: WDL
+    name: 03-TrainGCNV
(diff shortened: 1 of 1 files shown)


mwalker174:
Thanks for your patience @VJalili. This looks great! I do think we will need to move to a forking development model to avoid spamming Dockstore with pushes, as it already seems we are breaking its limits. Still, IMO the transition is worth it to further automate out deployment process.

mwalker174 (approved):

VJalili:
Thanks, @mwalker174! I agree, hopefully this automation will help with dockstore refresh/sync, and one less manual step in the release process.

PR #607, opened 2023-10-10
Check for empty scatter in GenotypeDepthPart2

epiercehoffman:
### Updates

We observed an inconsistent but reoccurring issue in which GenotypeBatch.GenotypeDepthPart2 failed in Terra with the error message `Workflow is making no progress but has the following unstarted job keys: [...]`. This issue is sometimes solved by re-running the workflow, but not always. It appears to be a cromwell bug but we hypothesize that it is related to the empty scatter over variants <5kb in length. As a workaround, I added a check to see if there were variants <5kb before scattering.

### Testing
* Tested in GREGoR cohort where this issue appeared repeatedly. All batches succeeded

### DIFF
--- a/wdl/GenotypeDepthPart2.wdl
+++ b/wdl/GenotypeDepthPart2.wdl
@@ -43,98 +43,102 @@ workflow GenotypeDepthPart2 {
       runtime_attr_override = runtime_attr_split_variants
   }
 
-  scatter (gt5kb_bed in SplitVariants.gt5kb_beds) {
+  if (length(SplitVariants.gt5kb_beds) > 0) {
+    scatter (gt5kb_bed in SplitVariants.gt5kb_beds) {
 
-    call tasksgenotypebatch.MakeSubsetVcf as MakeSubsetVcfOver5kb {
-      input:
-        vcf = cohort_vcf,
-        bed = gt5kb_bed,
-        sv_base_mini_docker = sv_base_mini_docker,
-        runtime_attr_override = runtime_attr_make_subset_vcf
-    }
+      call tasksgenotypebatch.MakeSubsetVcf as MakeSubsetVcfOver5kb {
+        input:
+          vcf = cohort_vcf,
+          bed = gt5kb_bed,
+          sv_base_mini_docker = sv_base_mini_docker,
+          runtime_attr_override = runtime_attr_make_subset_vcf
+      }
 
-    call tasksgenotypebatch.RDTestGenotype as RDTestGenotypeOver5kb {
-      input:
-        bin_exclude=bin_exclude,
-        bin_exclude_idx=bin_exclude_idx,
-        bed = gt5kb_bed,
-        coveragefile = coveragefile,
-        coveragefile_index = coveragefile_index,
-        medianfile = medianfile,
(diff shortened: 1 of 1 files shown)


wdl/GenotypeDepthPart2.wdl
```diff
@@ -91,50 +91,52 @@ workflow GenotypeDepthPart2 {
...
-    }
-    call tasksgenotypebatch.AddGenotypes as AddGenotypesUnder5kb {
-      input:
-        vcf = MakeSubsetVcfUnder5kb.subset_vcf,
-        genotypes = IntegrateDepthGqUnder5kb.genotypes,
-        varGQ = IntegrateDepthGqUnder5kb.varGQ,
-        prefix = basename(lt5kb_bed, ".bed"),
-        sv_pipeline_docker = sv_pipeline_docker,
-        runtime_attr_override = runtime_attr_add_genotypes
+  if (length(SplitVariants.lt5kb_beds) > 0) {
```
VJalili, line 94: Is not `select_all` you added below sufficient? I think it would be useful to test without this `if`, because if this turns out that this `if` is a requirement, it could happen on other workflows where we have shards like this.

VJalili (changes requested):

wdl/GenotypeDepthPart2.wdl
epiercehoffman: I haven't tested it without the `if` so I can't say for sure. This was my logic: 

- Without the `if`, the output arrays from the scatters are both required/defined - the lt5kb one simply may be empty. `select_all` selects all of the items in the array that are defined - but they were already both defined, so that didn't seem like it would be sufficient.
- It wasn't clear whether the cromwell bug was related to trying to initiate an empty scatter, or if it occurred during `flatten` when attempting to gather the empty scatter. The `if` covers both cases.

It is possible that there are other cases like this. We have observed this cromwell issue most often in GenotypeDepthPart2, but I know it has occurred in at least one other workflow before.

I agree that testing with and without the `if` could provide more insight into this issue. I don't have bandwidth to do that right now so I just wanted to get something that works.

mwalker174: The `if` statement seems to be needed due to a Cromwell issue, and that necessitates the `select_all` below. `select_all` only acts on optional types and would have no effect if the `if` weren't present.

mwalker174 (approved):
Thanks @epiercehoffman, looks good. 

For posteriority, depth-only VCFs should not contain variants under 5kbp (at least with the current version), but we are keeping those tasks in for legacy runs that may contain some. It will not add to workflow cost.

wdl/GenotypeDepthPart2.wdl
VJalili: This is a ubiquitous pattern and it seems we're been very lucky so far that other workflows run with no issues.

VJalili (approved):

wdl/GenotypeDepthPart2.wdl
epiercehoffman: It's not luck - it's rare that an array over which we scatter is empty like this

PR #464, opened 2022-11-28
Add CONTRIBUTING.md.

VJalili:
This PR adds the `contributing.md` that includes guidelines on code contribution and bug reporting.

PR #610, opened 2023-10-17
Reorganize docs on running the pipeline.

VJalili:
This PR improves the organization of the documentation on running and execution modes of the pipeline by improving on the description of the different execution modes and shiting focus of documentation on how to run the pipeline toward more end-user friendly options (i.e., using Terra) and moves the more advanced execution option (i.e., Cromwell) to a separate section under `/docs/advanced/development/cromwell`). 

Some pages are added as placeholders that should be completed in follow-up PRs.


website/docs/run/overview.md
```diff
@@ -0,0 +1,45 @@
...
+     power-users and developers. We provide guidelines for this option in the 
+     [_advanced guides_](/docs/advanced/cromwell) section.
+
+2. **How many samples are you planning to process?**
+    We have developed optimized flavors of GATK-SV allowing you to select the one that best suits 
+    your input size for minimal cost and shorter execution time.
+
+    | Execution Mode       | Ideal input size   | Per-sample execution cost | Per-sample execution time |
+    |----------------------|--------------------|---------------------------| ------------------------- |
+    | Singe-sample         | <100 samples       | $7.00-$10.00              | 24-36 hours               |
```
epiercehoffman, line 26:
```suggestion
    | Single-sample         | <100 samples       | $7.00-$10.00              | 24-36 hours               | 
```

website/docs/run/overview.md
```diff
@@ -0,0 +1,45 @@
...
+
+
+## GATK-SV on Terra
+
+Please refer to the following documentation on running the GATK-SV pipeline within the Terra platform. 
+
+- **Single-sample on Terra**: We have developed a 
+  [Terra workspace](https://app.terra.bio/#workspaces/help-gatk/GATK-Structural-Variants-Single-Sample)
+  for running GATK-SV on a single sample, which contains the related documentation. 
+- [Single-batch on Terra](batch.md);
```
epiercehoffman, line 38: I am hesitant about advertising the cohort mode Terra workspace (and single batch, because they are in a single workspace) on the website while it is not publicly available. I think this should at least be noted so that users don't think it's available. 

I'm also not sure that we want to make it seem like single-batch and cohort mode are two entirely different things. There are some differences but currently they share a workspace and more than half the pipeline is identical.

VJalili: That part is mainly from the docs we have on the Terra workspace. I also think that part can benefit from some simplification.

mwalker174 (commented):
Thanks for this update @VJalili. The new Overview page is laid out nicely. As we've discussed in person, there are some questions about how to present the cost data that we should address, but the figure with example data you have looks great.

In the "Run" section, let's start with the following:

- Overview
- Single-sample
- Joint calling

The "Joint calling" (i.e. cohort mode) can just contain a "Coming soon" message or the like.

Regarding Cromwell, I think we want to limit the scope of documentation for standalone servers, as we are presently phasing it out in favor of Terra. I think it will suffice to provide the information provided on the Advanced->Cromwell page for the occasional power-user but forego any specific sub-sections on the workflows.

As I mentioned in #464, it may be a good idea to add an "Advanced->Development" section that includes information on the input json system, terra configs, and wdl testing. This can be added later.

Minor note: "Quick Start" link on the home page actually takes you to "About" rather than "Quick Start".

website/docs/run/overview.md
VJalili: Yes, I agree. That part is from the docs we have on Terra workspace. We do not need to keep them in sync, we can improve here and change the workspace docs once these docs are thoroughly improved.

VJalili:
Thank you! I agree that presenting the execution modes as a single-sample and joint calling is much more intuitive. Accordingly, as you suggested, I reorganized the `/run/overview` page and left the `/run/single` and `/run/joint` calling pages as placeholders for now.

> As I mentioned in https://github.com/broadinstitute/gatk-sv/pull/464, it may be a good idea to add an "Advanced->Development" section that includes information on the input json system, terra configs, and wdl testing. This can be added later.

Agree!! I added this section; it is currently mostly a placeholder and only contains some overview about Cromwell, but we can update it in follow-up PRs. 

> Minor note: "Quick Start" link on the home page actually takes you to "About" rather than "Quick Start".

Good point, thanks! I fixed the link.

VJalili:
Thanks, @mwalker174!

PR #611, opened 2023-10-18
Add extra information to dashboard with regards to which workflow versions should be run

kirtanav98:
This adds extra information after the dashboard description of module 10. It guides the user to delete the irrelevant workflows based on how many batches are being used. This will prevent any unintended workflows to be run if there is an error or a certain workflow that would need to be run if issues arise further downstream.


epiercehoffman (changes requested):
I think you PR'd the wrong branch - I only see Vahid's recent changes regarding the scramble docker here, not any of your documentation updates to the dashboard

epiercehoffman:
Can you close this pull request since you opened a new one?

PR #612, opened 2023-10-18
Update cohort_mode_workspace_dashboard.md.tmpl

kirtanav98:
Add extra information after the description of Module 10 to guide the user as to which workflow versions to run. This clarifies which versions to use based on the batches to ensure the workflows are run as intended.


epiercehoffman (changes requested):
Thanks for pointing out the confusion surrounding the single-batch workflow versions and making documentation changes to help other users!

I want to make sure we get the wording right - I have a suggestion below but want to get your feedback on what's clear. 

Also, make sure to check the formatting of the dashboard by testing it out in Terra before you merge - I haven't done so for my suggested revision so I want to make sure the spacing is right.

epiercehoffman (approved):
Thanks for your updates and for checking the formatting! You're clear to merge

PR #613, opened 2023-10-19
Add ReshardVcf workflow

mwalker174:
This workflow takes a set of vcfs and shards them by contig. The method uses `bcftools concat` with `--regions` to pull records from all vcfs for each contig.

This workflow will be needed at the end of `ResolveComplexVariants`. This is because the contig-sharded vcfs produced in that workflow may contain a small number of records from other contigs, but downstream `CleanVcf` requires the input vcfs to be strictly contig sharded.

Tested on reference panel outputs from `GenotypeComplexVariants`.

### DIFF
--- a/wdl/ReshardVcf.wdl
+++ b/wdl/ReshardVcf.wdl
@@ -1,0 +1,87 @@
+version 1.0
+
+import "Structs.wdl"
+import "TasksMakeCohortVcf.wdl" as MiniTasks
+
+# Consumes a contig-sharded vcf where some shards may contain a small proportion of records from other
+#  contigs and moves them to their correct shards.
+
+workflow ReshardVcf {
+  input {
+    Array[File] vcfs  # Must be sorted and indexed
+    File contig_list
+    String prefix
+    Boolean? use_ssd
+    String sv_base_mini_docker
+    RuntimeAttr? runtime_override_reshard
+  }
+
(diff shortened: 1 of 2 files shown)


wdl/ReshardVcf.wdl
```diff
@@ -0,0 +1,87 @@
...
+    Boolean? use_ssd
+    String sv_base_mini_docker
+    RuntimeAttr? runtime_override_reshard
+  }
+
+  Array[String] contigs = read_lines(contig_list)
+
+  scatter (i in range(length(vcfs))) {
+    File vcf_indexes = vcfs[i] + ".tbi"
+  }
```
VJalili, line 23: Would it be useful to add an optional input `Array[File]? vcf_indexe` and fall back to this approach if the optional input was not provided? I think that can be useful to hint to a user if indexes are needed, and also provide them with an option of providing the indexes explicitly if they are on a different path.

VJalili (approved):
Thank you, @mwalker174! LGTM! I only added a comment.

wdl/ReshardVcf.wdl
mwalker174: As a convention we always require input VCFs to have indexes for top-level workflows. These are already generated by the pipeline so shouldn't cause any problems as long as the VCFs were not manually moved without their indexes.

PR #619, opened 2023-11-22
Copy or move the files instead of creating symlinks in the LocalizeReads task for compatibility with CoA/TES.

VJalili:
The task `LocalizeReads` creates symlinks for the localized cram and its corresponding index file in the current working directory to simplify referencing them in the task output. Specifically, the files are localized in the following directory, and symlinks are created in the `/cromwell_root/.` 

```
/cromwell_root/…/analysis/filename.cram
```

Symlinks as this are incompatible with GA4GH TES and [Cromwell on Azure](https://github.com/microsoft/CromwellOnAzure). 

Therefore, this PR updates the script to move or copy the files instead of creating symlinks. Moving files (`mv`) is faster and requires less disk space on the VM than copying the files. However, run on HPC, moving files could lead to moving them from their source directory compared to moving them from the localized directory to the root directory on the VM. Therefore, we provide both options for moving and copying to avoid moving files unexpectedly and set the copy as default.

Cromwell submission ID for a successful test of this PR is `bd10381c-53b3-4c67-a396-291135af0f6f`.

The following can be an alternative to moving the files, though it still needs to be tested.  

```
File output_file = basename(reads_path)  --> File output_file = reads_path
```

### DIFF
--- a/wdl/GatherSampleEvidence.wdl
+++ b/wdl/GatherSampleEvidence.wdl
@@ -24,8 +24,13 @@ workflow GatherSampleEvidence {
     Boolean collect_coverage = true
     Boolean collect_pesr = true
 
-    # Google Cloud Platform and Azure users can safely enable and get a slight cost improvement.
-    # Users running shared filesystems should NOT enable.
+    # Google Cloud Platform (GCP) and Azure users can safely enable and get a slight cost improvement.
+    # Users running on shared file systems should NOT enable this feature.
+    # Enabling this option when running the workflow on GCP or Azure will result in moving the **localized**
+    # files from one path to another on the VM, without impacting the files in their source persistent location.
+    # It will lead to using a slightly smaller disk size and running faster, thereby providing a slight cost improvement.
+    # However, when run on shared file systems (e.g., HPC), it will, by default, create a copy of the
+    # input files, and all subsequent operations will run on the deep copy of the input file.
     Boolean move_bam_or_cram_files = false
 
     # Convert ambiguous bases (e.g. K, S, Y, etc.) to N
--- a/wdl/GatherSampleEvidence.wdl
+++ b/wdl/GatherSampleEvidence.wdl
@@ -343,7 +343,7 @@ task LocalizeReads {
     # When this pipeline is run on an HPC, moving files could lead to
     # moving the files from their original source, compared to moving
     # them from one directory of the VM to another when run on Cloud.
-    # Therefore, to avoid moving files unexpectadely, we provide both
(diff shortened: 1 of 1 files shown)


mwalker174 (commented):
Thanks @VJalili! Did you happen to test using `cp` as well? We've had complaints in the past about moving `mv` on input files from folks using shared filesystems. I realize it's slower and would require extra disk, but I would think that the cost would be small.

VJalili:
That is a good point, thank you, @mwalker174. 

Yes, I first tested using `cp,` but it failed with an out-of-disk error, so I tried with `mv.` I think this is moving a file localized to a directory on the runner's VM to another directory on the VM. I think even when using a shared file system, this VM still has its own copy/localized version of the file, so it should not be moving a file on a different file system. Though I have not tested, and am happy to test, I will only need some references to files on the shared file system.

wdl/GatherSampleEvidence.wdl
```diff
@@ -23,6 +23,7 @@ workflow GatherSampleEvidence {
     # Evidence collection flags
     Boolean collect_coverage = true
     Boolean collect_pesr = true
+    Boolean move_bam_or_cram_files = false
```
mwalker174, line 26: Can you add a quick comment here explaining when this should be used? tldr; Google Cloud Platform users can safely enable and get a slight cost improvement. Users running on Azure or shared filesystems should NOT enable.

wdl/GatherSampleEvidence.wdl
```diff
@@ -335,8 +338,24 @@ task LocalizeReads {
...
 
   command {
-    ln -s ~{reads_path}
-    ln -s ~{reads_index}
+    set -exuo pipefail
+
+    # When this pipeline is run on an HPC, moving files could lead to
+    # moving the files from their original source, compared to moving
+    # them from one directory of the VM to another when run on Cloud.
+    # Therefore, to avoid moving files unexpectadely, we provide both
```
mwalker174, line 346:
```suggestion
    # Therefore, to avoid moving files unexpectedly, we provide both
```

mwalker174 (approved):
Looks good! Minor comments

wdl/GatherSampleEvidence.wdl
VJalili: That will be helpful! I added the comment with one modification: both GCP & Azure users can safely enable this, and only users running on shared file systems are _not_ recommended to enable it.

VJalili:
Thank you, @mwalker174!!

Issue #585, opened 2023-08-10
Make LocalizeReads optional


mwalker174:
The `LocalizeReads` task was recently introduced to `GatherSampleEvidence` to simplify cases for CRAMs sourced from requester-pays and multi-regional buckets. This simplified the code quite a bit. However, it would be beneficial in some cases to be able to bypass this for some cost and time savings. Adding an optional boolean skip this would be easy to implement.

epiercehoffman:
Addressed in #620

PR #620, opened 2023-11-28
Make LocalizeReads Optional in GatherSampleEvidence

kirtanav98:
This pull request addresses issue [#585](https://github.com/broadinstitute/gatk-sv/issues/585). The LocalizeReads task was introduced to GatherSampleEvidence to simplify cases for CRAMs sourced from requester-pays and multi-regional buckets to simplify the code, but could be skipped in some cases for some cost and time savings. An optional boolean variable set to a default value of true to run LocalizeReads was added. When set to false, the LocalizeReads task is skipped. 
This passed validation with both womtool and cromshell against the reference panel.

### DIFF
--- a/wdl/GatherSampleEvidence.wdl
+++ b/wdl/GatherSampleEvidence.wdl
@@ -134,14 +134,11 @@ workflow GatherSampleEvidence {
   }
 
 
-  File reads_file_ = select_first([CramToBamReviseBase.bam_file, LocalizeReads.output_file])
-  File reads_index_ = select_first([CramToBamReviseBase.bam_index, LocalizeReads.output_index])
-
   if (revise_base) {
     call rb.CramToBamReviseBase {
       input:
-        cram_file = reads_file_,
-        cram_index = reads_index_,
+        cram_file = select_first([LocalizeReads.output_file, bam_or_cram_file]),
+        cram_index = select_first([LocalizeReads.output_index, bam_or_cram_index]),
         reference_fasta = reference_fasta,
         reference_index = reference_index,
         contiglist = select_first([primary_contigs_fai]),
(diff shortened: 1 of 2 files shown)


wdl/GatherSampleEvidence.wdl
```diff
@@ -76,6 +76,11 @@ workflow GatherSampleEvidence {
     File? baseline_melt_vcf
     File? baseline_scramble_vcf
 
+
+    # Localize reads parameters
+    # set to true on default, skips localize_reads if set to false
+    Boolean run_localize_reads = true
```
epiercehoffman, line 82: I would move this parameter up after revise_base to better match the organization of parameters in this workflow - this new input pertains to actions on the CRAM file so it fits best towards the beginning

wdl/GatherSampleEvidenceBatch.wdl
```diff
@@ -64,6 +64,12 @@ workflow GatherSampleEvidenceBatch {
     File? baseline_melt_vcf
     File? baseline_scramble_vcf
 
+
+    # Localize reads parameters
+    # set to true on default, skips localize_reads if set to false
+    Boolean run_localize_reads = true
```
epiercehoffman, line 70: Similarly would move this input up higher in the inputs list - probably after collect_pesr

epiercehoffman (commented):
This is looking good! I have a minor suggestion on the organization of the new input parameter. Also, can you describe your testing in more detail? (ie. did you test with run_localize_reads = true and false, and did you verify whether LocalizeReads was run/not run accordingly?)

epiercehoffman (approved):
Thanks for running the additional test and for reorganizing the inputs! Looks good to merge

PR #622, opened 2023-12-12
Change Vapor bed preprocessing errors to warnings

mwalker174:
These errors are being triggered on allosome contigs, i.e. chrY in female samples. We can leave them in as warnings but they aren't strictly necessary.

Tested with a docker rebuild on a few samples that then completed successfully.

### DIFF
--- a/src/sv-pipeline/scripts/preprocess_bed_for_vapor.py
+++ b/src/sv-pipeline/scripts/preprocess_bed_for_vapor.py
@@ -78,9 +78,9 @@ def reformat(bed_in, bed_out, contig, sample_to_extract):
                                  fields[columns["name"]], svtype_write]) + "\n")
 
     if not found_contig:
-        raise ValueError(f"Could not find any records for contig {contig} in the provided BED file")
+        logging.warning(f"Could not find any records for contig {contig} in the provided BED file")
     if sample_to_extract is not None and "samples" in columns and not found_sample:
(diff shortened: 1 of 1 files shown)


src/sv-pipeline/scripts/preprocess_bed_for_vapor.py
```diff
@@ -78,9 +78,9 @@ def reformat(bed_in, bed_out, contig, sample_to_extract):
                                  fields[columns["name"]], svtype_write]) + "\n")
 
     if not found_contig:
-        raise ValueError(f"Could not find any records for contig {contig} in the provided BED file")
+        logging.warning(f"Could not find any records for contig {contig} in the provided BED file")
```
VJalili, line 81: Without making it too complicated, I wonder if it would be more useful to add an optional `--sex=male` flag to the argument list and throw the exception w.r.t to the value of this flag.

VJalili (approved):
I left a comment, but logging as waring looks equally good to me. Thank you!

src/sv-pipeline/scripts/preprocess_bed_for_vapor.py
mwalker174: This is a possible improvement but I'm going to leave it this way for now. We sometimes have unassigned sex or may want to run on a subset of contigs, which could make this more of a hindrance than a help.

PR #625, opened 2024-01-08
Find CN field for mCNVs in MainVcfQc without hard-coding order of format fields

epiercehoffman:
### Updates
* The order of the format fields was hard-coded in MainVcfQc, resulting in the CollectQcVcfWide script not using the CN field to evaluate genotypes for mCNVs when the order of the fields was updated after processing with GATK. This resulted in all mCNVs showing up as AF=0.5, creating a weird line in the AF plots. The script was updated to check the order of the format fields and use the CN field regardless of its position
* Memory increased for family QC plots. There are no families in the reference panel and HGDP requires more memory
* Filtered VCF key updated in template JSON

### Testing
* Ran MainVcfQc successfully on HGDP cleaned VCF & filtered (GATK-processed) VCF and verified that the AF varied for mCNVs in intermediate files and the weird line at 0.5 was absent from AF plots. Local testing also confirmed that the calculated AF was equal for the cleaned and filtered VCFs for a handful of example mCNVs.

### DIFF
--- a/inputs/templates/test/MainVcfQc/MainVcfQc.json.tmpl
+++ b/inputs/templates/test/MainVcfQc/MainVcfQc.json.tmpl
@@ -17,10 +17,5 @@
   "MainVcfQc.vcfs": [ {{ test_batch.clean_vcf | tojson }} ],
 
   "MainVcfQc.sv_per_shard": 2500,
-  "MainVcfQc.samples_per_shard": 600,
-
-  "MainVcfQc.runtime_override_plot_qc_per_family": {
-    "mem_gb": 15,
-    "disk_gb": 100
-  }
+  "MainVcfQc.samples_per_shard": 600
 }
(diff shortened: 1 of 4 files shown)


inputs/templates/test/MainVcfQc/MainVcfQc.json.tmpl
```diff
@@ -17,5 +17,10 @@
   "MainVcfQc.vcfs": [ {{ test_batch.clean_vcf | tojson }} ],
 
   "MainVcfQc.sv_per_shard": 2500,
-  "MainVcfQc.samples_per_shard": 600
+  "MainVcfQc.samples_per_shard": 600,
+
+  "MainVcfQc.runtime_override_plot_qc_per_family": {
+    "mem_gb": 15,
+    "disk_gb": 100
+  }
```
mwalker174, line 25: Should probably just make this the default in the WDL

mwalker174 (approved):
Thanks @epiercehoffman let's get this in. One minor comment.

PR #626, opened 2024-01-12
Integrate ReshardVcf into ResolveComplexVariants

mwalker174:
Adds a call to `ReshardVcf` at the end of `ResolveComplexVariants`.  In addition, each of the `BOTHSIDES_PASS` and `HIGH_SR_BACKGROUND` contig-sharded variant tables is concatenated into a single genome-wide table prior to annotating records with these flags in `CleanVcf`. This has a slight cost in memory footprint in `CleanVcf1a` but ensures that shuffled records are properly annotated. Json templates and top-level workflows are also updated.

This branch was successfully tested on `ResolveComplexVariants`, `GenotypeComplexVariants,` and `CleanVcf`.


wdl/ReshardVcf.wdl
```diff
@@ -3,20 +3,19 @@ version 1.0
...
+    Array[File] vcfs  # Order does not matter but must be sorted and indexed
     File contig_list
     String prefix
     Boolean? use_ssd
     String sv_base_mini_docker
     RuntimeAttr? runtime_override_reshard
   }
 
-  Array[String] contigs = read_lines(contig_list)
+  Array[String] contigs = transpose(read_tsv(contig_list))[0]
```
epiercehoffman, line 18: Make sure to update the ReshardVcf JSON template to use the primary_contigs_fai to match this change - it looks like it was initially set up with primary_contigs_list

epiercehoffman (approved):
This looks good, thanks for figuring out how to do the resharding efficiently and getting it integrated into our main workflows. Just one small update to make and then go ahead and merge

wdl/ReshardVcf.wdl
mwalker174: It runs ok with the contigs list too

PR #627, opened 2024-01-17
Update BEDTools version to 2.31.0

VJalili:
Updating to the current latest version with a static binary included in the release artifacts, `v.2.31.9`, for a fix for an error reported by Arthur.


mwalker174:
Thank you @VJalili, I confirmed that the cleaned vcf from your test is bitwise identical to the current main branch.

For posterity, this [bug](https://github.com/arq5x/bedtools2/issues/919) in bedtools caused truncated records greater than 1024 characters long, which could have affected several workflows, particularly where records are annotated with a column containing long sample lists.

mwalker174 (approved):

PR #628, opened 2024-01-18
Release AoU filtering model

mwalker174:
- Moves model file to a public bucket
- Adds Acknowledgements section to the README


epiercehoffman (approved):
Looks good! Thanks for figuring out approvals to share!

PR #629, opened 2024-01-19
Move cloud_sdk_docker to broad-dsde-methods

mwalker174:
Prevents quota errors we've been seeing recently with the public image.


epiercehoffman (approved):
Thank you! Hope this works

mwalker174:
Turns out the errors we were seeing were related to a different image that gets pulled at the start of every task. Google has resolved the issue.

PR #630, opened 2024-01-24
Disable MakeCohortVcfMetrics by default to avoid pesky resource errors

epiercehoffman:
Multiple users have encountered out-of-memory errors in MakeCohortVcfMetrics with larger cohorts. It can be costly to rerun CleanVcf with updated resources for very large cohorts. This metrics collection is meant for development and is not necessary for users to run every time, so it can be safely disabled. An alternative would be to increase the default memory, but it is enough memory for the test set and unnecessary to run for large cohorts so this seems more efficient.

### DIFF
--- a/wdl/CleanVcf.wdl
+++ b/wdl/CleanVcf.wdl
@@ -33,7 +33,7 @@ workflow CleanVcf {
     String? gcs_project
 
     # Module metrics parameters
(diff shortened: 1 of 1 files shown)


mwalker174 (approved):
Looks good. We should do this in other modules as well, as we're discussing in #633.

PR #583, opened 2023-08-09
Remove CHR2 and END2 for CPX at the end of CleanVcf

epiercehoffman:
### Updates
Remove CHR2 and END2 fields (if they exist) for all CPX records at the end of CleanVcf. These fields are redundant with the information in CPX_INTERVALS and are currently inconsistent and confusing (#377). 

### Testing
* Validated all WDLs and JSONs with womtool & terra validation script
* Rebuilt sv_pipeline_docker (us.gcr.io/broad-dsde-methods/eph/sv-pipeline:eph-cpx-remove-chr2-end2-6feed34) and tested CleanVcf with reference panel data. The workflow ran successfully and the resulting VCF did not have CHR2 or END2 fields for CPX records.

### Note
Downstream workflows should be tested and updated if needed to comply with this change.

PR #631, opened 2024-01-29
Remove CHR2 and END2 from INS in CleanVcf

epiercehoffman:
### Updates
* Remove CHR2 and END2 from INS in CleanVcf. In most cases, these fields duplicate information in CHROM and POS. In other cases, they duplicate information in SOURCE. The inconsistent use is confusing and unnecessary.
* Do not require CHR2 in svtk. If CHR2 is not present, use CHR2=CHROM. This was causing issues when running svtk vcfcluster after removing CHR2 from INS.

### Testing
* Built updated docker image, ran CleanVcf, and verified CHR2 and END2 tags were not present in INS records
* Ran reclustering workflow with updated docker image and it resolved the error message about CHR2<CHROM and the subsequent error message about the missing CHR2 field

### DIFF
--- a/src/sv-pipeline/scripts/format_svtk_vcf_for_gatk.py
+++ b/src/sv-pipeline/scripts/format_svtk_vcf_for_gatk.py
@@ -137,7 +137,7 @@ def convert(record: pysam.VariantRecord,
     is_ddup = svtype == 'CPX' and 'dDUP' in record.info.get('CPX_TYPE', '')
     if svtype == 'BND' or svtype == 'INS' or svtype == 'CTX' or is_ddup:
         record.stop = record.start + 1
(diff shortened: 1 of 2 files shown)


mwalker174 (approved):
We've been rolling with this for a while now so I'd assume it's okay still. Did it cause any issues with CPX records for other downstream workflows like ManualReview?

epiercehoffman:
> We've been rolling with this for a while now so I'd assume it's okay still. Did it cause any issues with CPX records for other downstream workflows like ManualReview?

We had to update ManualReview to accommodate #583. I don't recall any additional issues that arose as a result of altering the INS records (this PR). The issues we had with that were related to reclustering, and I addressed those as part of this PR.

PR #633, opened 2024-01-31
Change default behavior of GatherSampleEvidence to skip running module metrics

VJalili:
The `GatherSampleEvidence` workflow is set to run `GatherSampleEvidenceMetrics` by default. However, running that task requires setting values for the optional inputs `primary_contigs_fai` and `sv_pipeline_base_docker`. Therefore, the `GatherSampleEvidence` workflow currently fails if only the required input is provided. 

Since WDL does not support making the optional inputs as required w.r.t to the value of another argument, and Terra UI lacks features to hint to a user about such a chain of requirements, in-line comments in code seem the only option, as it is implemented. However, in-code comments are less user-friendly and should be reserved for developers and power users only.

Skipping to run this task as a default behavior results in the successful execution of the workflow if only the required input is provided seems to be a better design and improved UX.

### DIFF
--- a/wdl/GatherSampleEvidence.wdl
+++ b/wdl/GatherSampleEvidence.wdl
@@ -81,7 +81,7 @@ workflow GatherSampleEvidence {
 
     # Module metrics parameters
     # Run module metrics workflow at the end - on by default
-    Boolean run_module_metrics = true
+    Boolean run_module_metrics = false
     File? primary_contigs_fai # required if run_module_metrics = true
     String? sv_pipeline_base_docker  # required if run_module_metrics = true
     File? baseline_manta_vcf # baseline files are optional for metrics workflow


epiercehoffman:
The default input configuration in our template JSONs and Terra workspace includes all inputs required to run module metrics in the workflows where it is enabled by default. I honestly don't think it's a big issue, because we don't expect users to configure inputs from scratch - the inputs are complicated in other ways too. I'm definitely open to disabling module metrics collection by default, and to any changes that improve the clarity of inputs, but I think for now the real solution in terms of ease of use here is to pre-configure the inputs, which we have done.

VJalili:
It is a better UX to have the workflows run successfully if only the required input is provided.

mwalker174:
We should disable module metrics across the pipeline. We don't use them in production and have just been useful for development in a few rare cases.

PR #634, opened 2024-01-31
Replace `sv_pipeline_base_docker` with `sv_pipeline_docker`

VJalili:
A follow-up from https://github.com/broadinstitute/gatk-sv/pull/633

Since both `sv_pipeline_base_docker` and `sv_pipeline_docker` are references to the same docker image, this PR implements the following changes to consolidate them: 

- [x] If a workflow has both `sv_pipeline_docker` and `sv_pipeline_base_docker`,  remove `sv_pipeline_base_docker` and replaces all runtime references with `sv_pipeline_docker`. 
- [x] If a workflow does not have `sv_pipeline_docker`, then rename `sv_pipeline_base_docker` to `sv_pipeline_docker`.
- [x] Update all the templates according to the above changes.

A few other docker image variables also use the same image as `sv_pipeline_docker`. I will update these variables similarly in follow-up PRs.


mwalker174 (approved):
Thank you! This has been a long time coming. I hope you didn't have to do it all by hand :)

VJalili:
Thank you, @mwalker174!  

> I hope you didn't have to do it all by hand :)

I did! :) I think that was the only option to make sure the changes were correctly applied.

PR #635, opened 2024-02-02
Make manta region bed index required

VJalili:
The manta workflow fails if the index file does not exist; hence, following the discussion in https://github.com/broadinstitute/gatk-sv/issues/578, we are making this input required. Specifically:

The workflow fails to evaluate `vm_disk_size` and results in `EXECUTOR_ERROR` on CoA, which, since it occurs at the TES level, is a more challenging error to debug than task-level errors. 

The alternative approach to the changes implemented in this PR is to adjust the `vm_disk_size` calculation to exclude the index file size from the calculation (which would not impact much since the file is small). The main drawback of this approach is that the `/usr/local/bin/manta/bin/configManta.py` script may fail if the optional file does not exist. Hence, making the file required seems a better option.


mwalker174 (approved):
This looks good - the coding was a unnecessarily awkward. Do you want to do the same thing for the `reference_index` input also?

VJalili:
Thank you, @mwalker174, for the prompt review!! 

> Do you want to do the same thing for the reference_index input also?

Yes, I will update that next.

Issue #636, opened 2024-02-05
Make the ploidy_table output a final output in JoinRawCalls


kirtanav98:
The ploidy table is currently an intermediate output in JoinRawCalls. This file is required for further downstream analysis and is a necessary input for FilterGenotypes. Thus it should be a final output of the workflow. 
The JoinRawCalls wd. has been adjusted to make the ploidy table generated by calling  CreatePloidyTableFromPed from the TasksClusterBatch.wdl a final output of the entire wdl. 
This passed validation with both womtool and cromshell against the reference panel.

kirtanav98:
This issues has been resolved.

PR #637, opened 2024-02-05
Make the ploidy table output a necessary output instead of an intermediate …

kirtanav98:
The ploidy_table is an intermediate output in the JoinRawCalls workflow but is necessary for further downstream analysis. This file is required for further downstream analysis and is a necessary input for FilterGenotypes. Thus it should be a final output of the workflow.
The JoinRawCalls.wdl has been modified to make the ploidy table generated by calling CreatePloidyTableFromPed from the TasksClusterBatch.wdl a final output of the entire wdl.
This passed validation with both womtool and cromshell against the reference panel. This addresses issue #636 [https://github.com/broadinstitute/gatk-sv/issues/636#issue-2119378008].

### DIFF
--- a/wdl/JoinRawCalls.wdl
+++ b/wdl/JoinRawCalls.wdl
@@ -120,6 +120,6 @@ workflow JoinRawCalls {
   output {
     File joined_raw_calls_vcf = ConcatVcfs.concat_vcf
     File joined_raw_calls_vcf_index = ConcatVcfs.concat_vcf_idx
-    File joined_raw_calls_ploidy_table = SVCluster.ploidy_table
+    File joined_raw_calls_ploidy_table = CreatePloidyTableFromPed.out
   }
 }
--- a/wdl/JoinRawCalls.wdl
+++ b/wdl/JoinRawCalls.wdl
@@ -120,6 +120,6 @@ workflow JoinRawCalls {
   output {
     File joined_raw_calls_vcf = ConcatVcfs.concat_vcf
     File joined_raw_calls_vcf_index = ConcatVcfs.concat_vcf_idx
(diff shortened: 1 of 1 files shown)


wdl/JoinRawCalls.wdl
```diff
@@ -120,5 +120,6 @@ workflow JoinRawCalls {
   output {
     File joined_raw_calls_vcf = ConcatVcfs.concat_vcf
     File joined_raw_calls_vcf_index = ConcatVcfs.concat_vcf_idx
+    File joined_raw_calls_ploidy_table = CreatePloidyTableFromPed.out
```
epiercehoffman, line 123:
```suggestion
    File ploidy_table = CreatePloidyTableFromPed.out
```
This ploidy table is not specific to JoinRawCalls, so it makes more sense to just name it ploidy_table. I checked that  there are no other ploidy_table outputs from other workflows so it won't be confusing.

epiercehoffman (commented):
Thanks, this will be helpful for running the filtering workflows to avoid having to regenerate the ploidy table manually! Small change requested around naming and then you'll be good to go.

I want to make a note here that eventually we should move to generating the ploidy table once during ClusterBatch, saving it, and not regenerating it in JoinRawCalls. That will require addressing [this TODO](https://github.com/broadinstitute/gatk-sv/blob/0cad1662db0522dee17ae6d874a51cab6b739cae/wdl/ClusterBatch.wdl#L100). No action needed on that front right now though.

wdl/JoinRawCalls.wdl
kirtanav98: Thank you for that comment. The file name was changed as recommended.

epiercehoffman (approved):
Thanks!

PR #638, opened 2024-02-08
Replace hail docker with sv-pipeline image.

VJalili:
A follow up from #634 & #633 

Similar to #634:

- if a workflow has both `sv_pipeline_hail_docker` and `sv_pipeline_docker`, then remove `sv_pipeline_hail_docker` and replace all runtime docker image references with `sv_pipeline_docker`;
- if the workflow does not have  `sv_pipeline_docker` in its inputs, replace `sv_pipeline_hail_docker` with `sv_pipeline_docker`.


mwalker174 (approved):
Looks good. Unfortunately you're going to have to rebase with some conflicts at this point, but as long as it passes womtool/miniwdl validation you're good to go.

VJalili:
Thanks, @mwalker174!

PR #644, opened 2024-02-12
fix: add sd_locs_vcf_index input to SDtoBAF

### LINKED ISSUE #643: task SDtoBAF calls SiteDepthtoBAF without passing required index file
### Affected module(s) or script(s)
`task SDtoBAF` in `wdl/BatchEvidenceMerging.wdl`

### Affected version(s)
- [X] Latest public release version [0.28.4-beta]

### Description 
Task fails from a GATK User Error linked to missing index. (File paths were abridged).

```
***********************************************************************

A USER ERROR has occurred: An index is required but was not found for file /cromwell-executions/[...]/Homo_sapiens_assembly38.dbsnp138.vcf.gz:/cromwell-executions/[...]/Homo_sapiens_assembly38.dbsnp138.vcf.gz. Support for unindexed block-compressed files has been temporarily disabled. Try running IndexFeatureFile on the input.

***********************************************************************
```

This task calls GATK's `SiteDepthtoBAF` and one of its input, `--baf-sites-vcf`, requires a relevant index file.

#### Steps to reproduce
This workflow was ran with Cromwell on AWS Batch. This setup requires explicit input of index files size it cannot rely on having the index in the same location as the file.

#### Expected behavior
The `task SDtoBAF` should have an extra input field `sd_locs_vcf_index` to allow for the index to be passed (or it could be inferred in place by appending ".tbi"

This issue was solved on my setup by adding the index input file.

simojoe:
Fixes #643 

Adding the index in-place from the main file name.

I have not found any other reference to SDtoBAF that needs to be updated.

### DIFF
--- a/wdl/BatchEvidenceMerging.wdl
+++ b/wdl/BatchEvidenceMerging.wdl
@@ -63,11 +63,13 @@ workflow BatchEvidenceMerging {
     }
   }
   if (!defined(BAF_files)) {
+    File sd_locs_vcf_index = select_first([sd_locs_vcf]) + ".tbi"
+
     call SDtoBAF {
       input:
         SD_files = select_first([SD_files]),
         sd_locs_vcf = select_first([sd_locs_vcf]),
-        sd_locs_vcf_index = select_first([sd_locs_vcf]) + ".tbi",
+        sd_locs_vcf_index = sd_locs_vcf_index,
         batch = batch,
         samples = samples,
         rename_samples=rename_samples,
--- a/wdl/BatchEvidenceMerging.wdl
+++ b/wdl/BatchEvidenceMerging.wdl
@@ -67,6 +67,7 @@ workflow BatchEvidenceMerging {
       input:
         SD_files = select_first([SD_files]),
         sd_locs_vcf = select_first([sd_locs_vcf]),
+        sd_locs_vcf_index = select_first([sd_locs_vcf]) + ".tbi",
         batch = batch,
         samples = samples,
         rename_samples=rename_samples,
@@ -179,6 +180,7 @@ task SDtoBAF {
   input {
     Array[File] SD_files
     File sd_locs_vcf
+    File sd_locs_vcf_index
     String batch
     Array[String] samples
     Boolean rename_samples


VJalili:
@simojoe, thank you for fixing this! Optional index files can be confusing, and your suggested fix fits perfectly with what we discussed in https://github.com/broadinstitute/gatk-sv/issues/578.

I want to suggest an improvement on your approach; can you please add an optional input `sd_locs_vcf_index` to the WDL, such that, if `sv_locs_vcf` is given, the `sd_locs_vcf_index` should also be given, and if `sd_locs_vcf_index` is not given, then infer it from `sv_locs_vcf`.

mwalker174 (changes requested):
Thanks @simojoe. I was at first surprised since we haven't seen this before but then noticed you are using the gzipped version of the dbsnp vcf. We currently reference the uncompressed version, which is why it didn't require an index. Unfortunately, if we ran your change with the uncompressed version it would fail because its index does not have a `.tbi` extension:

```
> gsutil ls "gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.dbsnp138.vcf*"
gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.dbsnp138.vcf
gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.dbsnp138.vcf.gz
gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.dbsnp138.vcf.gz.tbi
gs://gcp-public-data--broad-references/hg38/v0/Homo_sapiens_assembly38.dbsnp138.vcf.idx
```

I think we should be using the compressed one instead, but this will affect the code base in 2 other places: 

- Needs index file input: https://github.com/broadinstitute/gatk-sv/blob/0cad1662db0522dee17ae6d874a51cab6b739cae/wdl/CollectSVEvidence.wdl#L57
- Update to the `.gz` file: https://github.com/broadinstitute/gatk-sv/blob/0cad1662db0522dee17ae6d874a51cab6b739cae/inputs/values/resources_hg38.json#L15

PR #646, opened 2024-02-15
Remove KING and PLINK

mwalker174:
The release of PLINK referenced in the sv-pipeline Dockerfile no longer exists and has been causing Docker builds to fail. This tool is not needed for any supported workflows. The KING library for relatedness inference, while still installing properly, also is not needed by any workflows.

- Removes KING and PLINK from the sv-pipeline Dockerfile
- Deletes parse_KING_results.R script
- Deletes FilterCleanupQualRecalibration workflow, which is a part of MinGq filtering but is no longer supported


mwalker174:
Addressed in #584

PR #650, opened 2024-03-01
Make SplitVariants task in TasksGenotypeBatch.wdl  compatible with downstream analysis

kirtanav98:
The SplitVariants task used to have some lines to switch columns 5 and 6 of the bed file output, which is read in downstream tasks of TrainRDGenotyping.GenotypePESR. This causes the TrainRDGenotyping.GenotypePESR to error out reporting.

Error: WARNING: Incorrect CNV type specified
1: stop("WARNING: Incorrect CNV type specified")
The python script splitvariants.py was modified to switch the columns to the appropriate order to be compatible with downstream analysis requirements. The script passed testing with both womtool and cromshell on the 1kg reference panel with the bca flag set to both True and False.


epiercehoffman (approved):
Thank you for addressing this issue! Based on your testing and mine (described below), this appears to replicate the behavior of the old version of SplitVariants (as of v0.28.4-beta), so it looks good to merge.

Testing: I ran GenotypeBatch on the ref_panel with your branch and with v0.28.4-beta. Both succeeded. I confirmed that the genotyped_depth_vcf, genotyped_pesr_vcf, sr_background_fail, and sr_bothside_pass outputs were identical. I also spot-checked many of the output files directly from the SplitVariants task and found that they were identical other than the lack of a newline character at the end of the new files - but that doesn't appear to impact the final outputs.

PR #653, opened 2024-03-08
Workflow to apply NCR and reference artifact filters and remove zero-carrier sites

epiercehoffman:
### Updates

Create a standalone WDL for NCR and reference artifact filters. Also removes sites with zero carriers. This script pulls code from the existing NCR calculation in apply_sl_filter.py but applies it independently from genotype filtering to allow for intervening steps like outlier removal and CPX filtering which impact the NCR calculation.

Marked as draft while AoU Phase 2 downstream steps development is ongoing.


mwalker174:
We need to have an offline discussion about how to streamline these downstream steps. Not ideal to have two implementations of NCR.

PR #655, opened 2024-03-14
Xz fixes 5

xuefzhao:
updated the AF annotation scripts to :
1. revise "pop_AF" to "AF_pop"
2. added "CN_STATUS"

### DIFF
--- a/src/sv-pipeline/05_annotation/scripts/compute_AFs.py
+++ b/src/sv-pipeline/05_annotation/scripts/compute_AFs.py
@@ -235,6 +235,7 @@ def calc_allele_freq(record, samples, prefix=None, hemi=False):
             nonnull_CNs, nonref_CN_count, nonref_CN_freq = [0] * 3
             CN_dist = (0, )
             CN_freqs = (0, )
+            CN_status = (0, )
         else:
             # Count number of samples per CN and total CNs observed
             CN_counts = dict(Counter(CNs))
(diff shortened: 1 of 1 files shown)


src/sv-pipeline/05_annotation/scripts/compute_AFs.py
```diff
@@ -360,51 +348,54 @@ def main():
         '##INFO=<ID=FREQ_HOMALT,Number=1,Type=Float,Description="Homozygous alternate genotype frequency (biallelic sites only).">',
         '##INFO=<ID=CN_NUMBER,Number=1,Type=Integer,Description="Total number of samples with estimated copy numbers (multiallelic CNVs only).">',
         '##INFO=<ID=CN_COUNT,Number=.,Type=Integer,Description="Number of samples observed at each copy state, starting from CN=0 (multiallelic CNVs only).">',
+        '##INFO=<ID=CN_STATUS,Number=.,Type=Integer,Description="Differnt copy states observed, starting from CN=0 (multiallelic CNVs only).">',
```
epiercehoffman, line 351:
```suggestion
        '##INFO=<ID=CN_STATUS,Number=.,Type=Integer,Description="Different copy states observed, starting from CN=0 (multiallelic CNVs only).">',
```
What is CN_STATUS and why are we adding it? It looks like it's just 0,1,...,maxCN, is that right? But we may not observe every copy state between 0 and maxCN so this description does not seem accurate. What is it used for?

epiercehoffman:
* Updated CN_STATUS description for clarity
* Added 1KG population assignments file for testing population AF annotation
* Tested AnnotateVcf on reference panel and verified that CN_STATUS description and AF_pop fields were as expected in output

epiercehoffman (approved):
Thanks for making these changes that were needed for gnomAD!

PR #665, opened 2024-04-10
Make MELT scripts independent of the execution path.

VJalili:
MELT scripts build logic dependent on the script invocation path. The invocation path on Cromwell is a deployment-specific configuration, and currently, the default settings on GCP and Azure differ, which breaks MELT execution. We have two possible tracks to address this: (1) update the scripts to match the currently configured invocation path on Azure and GCP, or (2) update the scripts to be independent of the invocation path. This PR implements the latter approach as it seems more portable and invariant to deployment-specific configurations of Cromwell. 

Specifically, this PR implements the following: 

- [x] Remove invocation path-dependent logic from MELT script and WDL;
- [ ] Update MELT docker and update `dockers*.json` accordingly.

### DIFF
--- a/dockerfiles/melt/run_MELT_2.0.5.sh
+++ b/dockerfiles/melt/run_MELT_2.0.5.sh
@@ -19,7 +19,7 @@ set -euo pipefail
 ##### Usage statement
 usage(){
 cat <<EOF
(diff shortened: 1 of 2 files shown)


mwalker174 (approved):
Thanks!

VJalili:
Thank you, @mwalker174!

PR #674, opened 2024-05-06
Fix UnboundLocalError

CuriousTim:
The `src/sv-pipeline/scripts/make_evidence_qc_table.py` script attempts to set the columns of a pandas DataFrame before creating it resulting in an `UnboundLocalError`. This commit swaps these steps.


mwalker174 (approved):
Thanks looks good!

PR #675, opened 2024-05-13
Add documentation for the de-novo pipeline

VJalili:
This PR extends the docs in the following areas: 

- [ ] Document input and output of the de-novo pipeline;
- [ ] Document the method used for de-novo variant calling.

### DIFF
--- a/website/docs/modules/denovo.md
+++ b/website/docs/modules/denovo.md
@@ -1,0 +1,126 @@
+---
+title: DeNovoSV
+description: Downstream filtering (work in progress)
+sidebar_position: 15
+slug: denovo
+---
+
+The de-novo workflow operates on the annotated multi-sample VCF file created by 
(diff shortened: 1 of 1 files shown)


website/docs/concepts/denovo.md
```diff
@@ -0,0 +1,44 @@
+---
```
mwalker174, line 1: This file is a lot of detail. IMO, it would be better to reference the scripts themselves and have those be sufficiently organized/commented that the methods are clear.

mwalker174 (commented):
I have some initial suggestions here. I think the inputs need to be greatly simplified / cleaned up for use in Terra before we commit any documentation.

VJalili:
Thank you, @mwalker174, for the feedback! I agree that we need the setup of the inputs polished before we add docs; I will update docs after the inputs are updated.

PR #682, opened 2024-05-29
Set the default disk size in AnnotateIntervals as a function of input files size

VJalili:
The `AnnotateIntervals`, called as part of the TrainGCNV workflow, is running out of disk on ToA. The default disk size is set to a fixed value, `10 GiB`. This PR updates that to be computed as a function of input files size. 

Tested successfully on GCP: `45d20c60-8a80-47c7-8298-c77f2d3d18d1`

### DIFF
--- a/wdl/GermlineCNVTasks.wdl
+++ b/wdl/GermlineCNVTasks.wdl
@@ -18,10 +18,12 @@ task AnnotateIntervals {
       RuntimeAttr? runtime_attr_override
     }
 
+    Float input_size = size(select_all([intervals, ref_fasta, mappability_track_bed, segmental_duplication_track_bed]), "GB")
+
     RuntimeAttr default_attr = object {
       cpu_cores: 1,
       mem_gb: 1.7,
-      disk_gb: 10,
+      disk_gb: ceil(10.0 + input_size),
       boot_disk_gb: 10,
       preemptible_tries: 3,
       max_retries: 1


wdl/GermlineCNVTasks.wdl
```diff
@@ -18,10 +18,12 @@ task AnnotateIntervals {
       RuntimeAttr? runtime_attr_override
     }
 
+    Float input_size = size(select_all([intervals, ref_fasta, mappability_track_bed, segmental_duplication_track_bed]), "GB")
+
     RuntimeAttr default_attr = object {
       cpu_cores: 1,
       mem_gb: 1.7,
-      disk_gb: 10,
+      disk_gb: ceil(10.0 + input_size),
```
MattMcL4475, line 26: Is 10GB enough for the docker image and all outputs?  Might also want to multiply it by 1.15, since Azure Batch requires 10% of the disk as well: `disk_gb: ceil((10.0 + input_size) * 1.15)`

VJalili: That is a good point!

mwalker174: On GCP, the docker image goes on a separate boot disk

mwalker174 (approved):

wdl/GermlineCNVTasks.wdl
VJalili: I merged this before having pushed a commit with @MattMcL4475's suggestion. Hopefully, it will work with the increased disk size; otherwise, I will make a follow-up PR.

PR #684, opened 2024-05-30
Prevent sample ID mangling in WGD computation

### LINKED ISSUE #683: X gets added to some sample IDs in WGD in EvidenceQc
We have observed that X gets prepended to some sample IDs (those that start with `__` or a numeric digit) during processing with R for WGD during EvidenceQc. This messes with the `qc_table` output of EvidenceQc because there are some rows with the original sample IDs and some with the modified sample IDs.

CuriousTim:
When reading a file in R via read.table, the default is to convert all column names into syntactically valid identifiers. Identifiers can only start with a letter or a dot, but a dot cannot be followed by a number. To "fix" these column names, read.table will prepend them with an X. Because the column names in the WGD scoring matrix are sample IDs, which could begin with a non-letter character, sample IDs become mangled when scoring dosage biases. This results in fragmented rows in the final QC table generated by 02-EvidenceQC. This commit changes the way the dosage bias scoring script reads the WGD scoring matrix so that sample IDs are not changed.

One assumption of this commit is that all the sample IDs are unique.

Fixes #683

### DIFF
--- a/src/WGD/bin/scoreDosageBiases.R
+++ b/src/WGD/bin/scoreDosageBiases.R
@@ -15,9 +15,9 @@ options(scipen=1000,stringsAsFactors=F)
 #####Helper function to load coverage matrix
 ############################################
 readMatrix <- function(INFILE){
-  dat <- read.table(INFILE,comment.char="",header=T)
+  dat <- read.table(INFILE,check.names=FALSE,comment.char="",header=T)
   colnames(dat)[1] <- "Chr"
-  dat[,-1] <- t(apply(dat[,-1],1,as.numeric))
+  dat[,-1] <- apply(dat[,-1],2,as.double)
   return(dat)
 }
 


mwalker174:
Thanks @CuriousTim, have you tested this script? I don't think you'd need to do a full docker rebuild, but at least testing in a local environment.

CuriousTim:
I tested the script on a batch with only problematic sample IDs and a batch with both problematic and non-problematic sample IDs. I can do some further testing for only non-problematic sample IDs and more exotic sample IDs. I also have not tested anything downstream.

mwalker174:
Those extra tests would be a good idea to rule out any possible regressions. I don't think you need to test downstream in this case, as long as the output looks correct to you.

CuriousTim:
I tested a batch with all non-problematic IDs and the script works. I also tested sample IDs containing combinations of every single printable ASCII character except space (32) and DEL (127) and the only character that breaks the script is a double quote character, though that can be addressed by passing `quote = ""` to the `read.table()` call. However, given the sample ID restrictions of the pipeline, I don't think the script needs to cover those cases.

mwalker174 (approved):
Sounds good, thanks!

PR #685, opened 2024-06-03
Update docs on building inputs

VJalili:
This PR gathers all the docs related to building inputs in one section, simplifies the descriptions, and improves their formatting. 

As a side update, it also fixes the mermaid version to the same version as Docusarus to avoid version conflicts.


website/docs/advanced/build_inputs.md
```diff
@@ -0,0 +1,137 @@
...
+```
+
+
+### Generating a reference panel
+
+This section only applies to the single-sample mode. 
+New reference panels can be generated from a single run of the 
+`GATKSVPipelineBatch` workflow. 
+If using a Cromwell server, we recommend copying the outputs to a p
+ermanent location by adding the following option to the workflow configuration file:
```
mwalker174, line 75:
```suggestion
If using a Cromwell server, we recommend copying the outputs to a 
permanent location by adding the following option to the workflow configuration file:
```

mwalker174: Link "workflow configuration" to the Cromwell docs so it's clear which file we're referring to: https://cromwell.readthedocs.io/en/latest/wf_options/Overview/

website/docs/advanced/build_inputs.md
```diff
@@ -0,0 +1,137 @@
+---
+title: Default inputs
+description: Building default inputs
```
mwalker174, line 3:
```suggestion
title: Building inputs
description: Building work input json files
```

mwalker174 (approved):
Thanks @VJalili, just a few minor suggestions here.

VJalili:
Thanks, @mwalker174!

PR #689, opened 2024-06-10
Deprecate single-batch Terra configs

epiercehoffman:
### Updates

Terra now supports the creation and use of sample_set_sets containing a single sample_set. This allows us to remove the separate workflow configurations for single-batch cohorts and use a single set of input configurations that assume a sample_set_set, which will be simpler for users and for us as maintainers. This PR removes the single-batch JSON templates and updates the Terra dashboard to reflect those changes.

### Testing
* Creation and use of a sample_set_set containing a single sample_set was tested in [this workspace](https://app.terra.bio/#workspaces/broad-firecloud-dsde-methods/gatk-sv-pipeline-1kgp-copy-manual-rev-test/data) with MergeBatchSites and RegenotypeCNVs
* New dashboard is live in the same workspace to test formatting


mwalker174 (approved):
Awesome!

Issue #586, opened 2023-08-15
Make Vapor plots optional to store


epiercehoffman:
Vapor produces a large amount of storage (~1.5 GiB per sample), 99% of which is the plots. These plots are not necessary to store for every sample, so to reduce storage costs we should add a parameter to the WDL to toggle whether those plots are kept as a workflow output or not.

epiercehoffman:
Addressed in #692

PR #692, opened 2024-06-27
Making the Vapor plots optional to store as a final output

kirtanav98:
This PR addresses issue [586](https://github.com/broadinstitute/gatk-sv/issues/586). Vapor produces a large amount of storage (~1.5 GiB per sample), 99% of which is related to the plots. To reduce storage, a boolean parameter 'save_plots' was added to the WDL to toggle whether those plots are kept as a final workflow output or not. When set to false, the plots are not produced as a final output. When set to true, the plots are produced as a final output. 
The updated wdl and json were tested using womtool with the 1kgp data for both instances, when the parameter is set to false and when it is set to true. Testing was also done by running updated script in Terra with the boolean set to "true" and "false" to ensure the plots were generated or not as per the input.

### DIFF
--- a/wdl/Vapor.wdl
+++ b/wdl/Vapor.wdl
@@ -28,7 +28,7 @@ workflow Vapor {
     RuntimeAttr? runtime_attr_LocalizeCram
 
     Boolean save_plots  # New input to control whether plots are final output
-     File NONE_FILE = write_lines([])  # Create an empty file
+    File? NONE_FILE_ # Create an empty file
   }
 
   scatter (contig in read_lines(contigs)) {
@@ -69,7 +69,7 @@ workflow Vapor {
 
   output {
     File vapor_bed = ConcatVapor.merged_bed_file
-    File vapor_plots = if save_plots then ConcatVapor.merged_bed_plot else NONE_FILE
+    File vapor_plots = if save_plots then select_first(ConcatVapor.merged_bed_plot) else NONE_FILE_
   }
 }
 
--- a/wdl/Vapor.wdl
+++ b/wdl/Vapor.wdl
@@ -69,7 +69,7 @@ workflow Vapor {
 
   output {
     File vapor_bed = ConcatVapor.merged_bed_file
-    Array[File]? vapor_plots = if save_plots then select_all(RunVaporWithCram.vapor_plot) else NONE_FILE_
+    File? vapor_plots = if save_plots then ConcatVapor.merged_bed_plot else NONE_FILE_
   }
 }
 
--- a/wdl/Vapor.wdl
+++ b/wdl/Vapor.wdl
@@ -27,8 +27,8 @@ workflow Vapor {
     RuntimeAttr? runtime_attr_concat_beds
     RuntimeAttr? runtime_attr_LocalizeCram
 
-    Boolean save_plots
-    File NONE_FILE
+    Boolean save_plots  # New input to control whether plots are final output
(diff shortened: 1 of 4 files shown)


wdl/Vapor.wdl
```diff
@@ -20,15 +20,18 @@ workflow Vapor {
...
-    RuntimeAttr? runtime_attr_vapor 
+    RuntimeAttr? runtime_attr_vapor
     RuntimeAttr? runtime_attr_bcf2vcf
     RuntimeAttr? runtime_attr_vcf2bed
     RuntimeAttr? runtime_attr_split_vcf
     RuntimeAttr? runtime_attr_concat_beds
     RuntimeAttr? runtime_attr_LocalizeCram
+
+    Boolean save_plots  # New input to control whether plots are final output
+    File? NONE_FILE_ # Create an empty file
```
epiercehoffman, line 31:
```suggestion
    File? NONE_FILE_  # Create a null file - do not use this input
```

wdl/Vapor.wdl
```diff
@@ -20,15 +20,18 @@ workflow Vapor {
...
 
-    RuntimeAttr? runtime_attr_vapor 
+    RuntimeAttr? runtime_attr_vapor
     RuntimeAttr? runtime_attr_bcf2vcf
     RuntimeAttr? runtime_attr_vcf2bed
     RuntimeAttr? runtime_attr_split_vcf
     RuntimeAttr? runtime_attr_concat_beds
     RuntimeAttr? runtime_attr_LocalizeCram
+
+    Boolean save_plots  # New input to control whether plots are final output
```
epiercehoffman, line 30:
```suggestion
    Boolean save_plots  # Control whether plots are final output
```
You don't need to specify that it's a new input - in a year it won't be new! I would also move this up - probably between sample_id and ref_fasta. We like to group inputs together, and this isn't a runtime attribute so it should be higher up. You can leave NONE_FILE_ where it is though to discourage people from trying to provide it a value.

epiercehoffman (commented):
Nice work on this! I left some minor comments on the new inputs you added but otherwise looking very good. I also checked the Terra tests you ran and it looks like it's working as intended!

wdl/Vapor.wdl
kirtanav98: Thank you for that clarification, the edit was made.

wdl/Vapor.wdl
kirtanav98: Thank you for that clarification and organization suggestion. These changes were made.

wdl/Vapor.wdl
```diff
@@ -11,6 +11,8 @@ workflow Vapor {
     File bed_file
     String sample_id
 
+    Boolean save_plots  # Control whether plots are produced in final output
```
epiercehoffman, line 14:
```suggestion
    Boolean save_plots  # Control whether plots are final output
```
This is really nit-picky, but as we've discussed, the plots are produced either way - this is a question of whether they are designated as final workflow outputs, which can translate to whether they get saved vs. deleted

epiercehoffman (approved):
Thanks for making those changes! I have one really nit-picky comment but after addressing that you're clear to merge

wdl/Vapor.wdl
kirtanav98: Understood, fix was made. I guess it can translate either way but I see how that makes it clearer. Thank you!

PR #693, opened 2024-06-27
Reduce memory usage in GenotypeSRPart1

mwalker174:
Replaces `grep -f` with equivalent `awk` statements after a user reported very high (>30GB) memory usage in this task.

The failing case now runs with under 8GB of memory. Tested on the bwa-melt workspace to ensure identical outputs in GenotypeBatch.


epiercehoffman (approved):
Seems like these changes were both successful and effective in reducing memory when Krithika tried this out, and we should definitely get this change merged to prevent further memory issues. Thanks for the fix!

PR #699, opened 2024-07-24
Update deprecated Pandas append operation for log analysis

kjaisingh:
`analyze_monitoring_logs2.py` currently uses the `append` operation of Pandas, though this has been deprecated since Pandas 1.4.0. 

Updates this reference to instead use Pandas' new `concat` operator.


epiercehoffman (approved):
Looks good, go ahead and merge!

I tested this and it is compatible with earlier versions of pandas (tested 1.0.5) and the grouped.tsv output was identical to the previous version of the script.

PR #712, opened 2024-08-19
Fix file naming in GenotypeBatch.SplitVariants

epiercehoffman:
### Updates

The SplitVariants task in GenotypeBatch was not successfully incrementing file name suffixes; after reaching `aaaaaz` it looped back to `aaaaaa`. This resulted in variants being dropped from sufficiently large VCFs in the output.

This PR updates SplitVariants to use numeric file naming for simplicity, with a default of 9 characters in the suffix for a maximum of 1,000,000,000 files (to ensure that there can be at least as many files as alphabetical naming with 6 characters in the suffix, which allows for 308,915,776 files). 

### Testing

* Tested manually on data that was large enough to trigger the issue with the previous SplitVariants version and verified that 52 files were produced instead of the previous and erroneous 26 files.
* Tested manually on the same data with `-d 1` and verified that the error message was produced after 10 files were written.
* Tested with an updated docker image to verify a successful run on the reference panel. The depth VCF was identical to the one produced by v0.28.5-beta (this test set is too small to observe the issue of overwriting files), although there were small differences in the PESR VCF, likely due to the updates to the way MEIs are treated (the number of records was identical)

### DIFF
--- a/src/sv-pipeline/04_variant_resolution/scripts/split_variants.py
+++ b/src/sv-pipeline/04_variant_resolution/scripts/split_variants.py
@@ -3,15 +3,17 @@ import argparse
 import logging
 
 
-def process_bed_file(input_bed, n_per_split, bca=True):
+def process_bed_file(input_bed, n_per_split, bca=True, digits=9):
     SVTYPE_FIELD = 5
     END_FIELD = 2
     START_FIELD = 1
 
     # Check the conditions to generate prefixes for the output files
     condition_prefixes = {
-        'gt5kb': {'condition': lambda line: (line[SVTYPE_FIELD] == 'DEL' or line[SVTYPE_FIELD] == 'DUP') and (int(line[END_FIELD]) - int(line[START_FIELD]) >= 5000)},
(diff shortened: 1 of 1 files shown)


src/sv-pipeline/04_variant_resolution/scripts/split_variants.py
```diff
@@ -3,74 +3,70 @@
...
-def increment_suffix(suffix):
-    alphabet = 'abcdefghijklmnopqrstuvwxyz'
-    if suffix == 'z' * 6:
-        raise ValueError('All possible files generated.')
-    else:
-        index = alphabet.index(suffix[0])
-        next_char = alphabet[(index + 1) % 26]
-        return next_char + suffix[1:]
+def get_file_name(prefix, suffix, digits):
+    if suffix >= 10 ** digits:
```
mwalker174, line 57: I like that you're checking this but the unbounded exponent scares me a little. A safer check would be to get the actual suffix string with zfill, check if it's all 0's and then confirm that prefix == 0. If prefix > 0 we've overflowed.

mwalker174 (approved):
Looks good, just one suggestion on checking for overflow

src/sv-pipeline/04_variant_resolution/scripts/split_variants.py
epiercehoffman: Updated as discussed offline, and repeated local tests (with -d 9 and -d 1)

PR #719, opened 2024-08-27
Update GATK to 4.6.0.0 in sv_pipeline_docker

mwalker174:
Bumps the version of GATK to 4.6.0.0 in `sv_pipeline_docker`. Note this is not the same as `gatk_docker`, as `gatk` is needed in some tasks that use GATK-SV code (e.g. svtk) and so it is built into the `sv_base_docker` image.

This addresses #81, but also the current commit has rotted with a dependency error, preventing new builds of `sv_base_docker`.

Note that the newer version of `PrintSVEvidence` produces an index file with 0-based indexing. This required changing/moving some `tabix` commands in the WDLs for consistency and to avoid wastefully regenerating the index. The 0-based indexing also caused problems with the `sr-count` call in `GenotypeBatch`, since the queries were off by 1 and this is the only place where tabix-based evidence queries require single-base precision.

Tested `GenotypeBatch` against v0.28.1-beta with the "bwa-melt" test cohort. Discrepancies in the output were minor and could be explained by recent PRs (e.g. #618). Also tested `GatherBatchEvidence`, `GenerateBatchMetrics`, `ResolveComplexVariants`, and `GenotypeComplexVariants` which passed sanity checks.

### DIFF
--- a/dockerfiles/sv-base/Dockerfile
+++ b/dockerfiles/sv-base/Dockerfile
@@ -1,7 +1,7 @@
 # This is the base dockerfile for the GATK SV pipeline that adds R, a few R packages, and GATK
 ARG SAMTOOLS_CLOUD_IMAGE=samtools-cloud:latest
 ARG VIRTUAL_ENV_IMAGE=sv-base-virtual-env:latest
-ARG GATK_COMMIT="a33bf19dd3188af0af1bd17bce015eb20ba73227"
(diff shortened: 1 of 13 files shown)


epiercehoffman (approved):
Thanks for unraveling this puzzle! As we discussed offline, it seems that a better long-term solution would be to ensure the index matches the evidence files, but this makes sense as an interim solution to allow the docker to be built with the new GATK version.

Issue #724, opened 2024-09-16
GATKSVPipelineSingleSample is missing from Dockstore automation


VJalili:
The `GATKSVPipelineSingleSample` workflow is missing from [`.dockstore.yml`](https://github.com/broadinstitute/gatk-sv/blob/main/.github/.dockstore.yml), hence the latest version available on Dockstore is v0.28.4.

epiercehoffman:
Addressed in #726

PR #726, opened 2024-09-23
Add the single-sample pipeline to the Dockstore automation.

VJalili:
Addresses https://github.com/broadinstitute/gatk-sv/issues/724

The latest version of the `GATKSVPipelineSingleSample` on Dockstore is v0.28.4; hence, the latest two releases, i.e., v0.28.5 and v0.29.0 are not published to Dockstore. This is because the workflow is not tracked and published by the Dockstore automation, which this PR fixes. Specifically, this PR makes the following changes:

- Add `GATKSVPipelineSingleSample` to Dockstore automation;
- This workflow was previously published under the `gatk-sv/gatk-sv-single-sample` name, which does not match the naming convention of other workflows. Hence, we are renaming it, and the new versions will publish under the `gatk-sv/SingleSample` name that matches the naming convention of other workflows.

### DIFF
--- a/.github/.dockstore.yml
+++ b/.github/.dockstore.yml
@@ -161,3 +161,12 @@ workflows:
         - main
       tags:
         - /.*/
+
+  - subclass: WDL
+    name: SingleSample
+    primaryDescriptorPath: /wdl/GATKSVPipelineSingleSample.wdl
+    filters:
+      branches:
+        - main
+      tags:
+        - /.*/


.github/.dockstore.yml
```diff
@@ -161,3 +161,12 @@ workflows:
         - main
       tags:
         - /.*/
+
+  - subclass: WDL
+    name: SingleSample
```
mwalker174, line 166:
```suggestion
    name: SingleSamplePipeline
```
How about this? Our workflow names usually contain a verb

mwalker174 (approved):
Looks okay, just a suggestion on the name

.github/.dockstore.yml
VJalili: Agree, that reads even better!

PR #728, opened 2024-09-26
Enhance EvidenceQC outputs to better align with sample QC 

kjaisingh:
### Description
This PR is intended to introduce several enhancements to `qc_matrix` output as part of the `EvidenceQC` workflow that enable it to align better with the sample QC process, which is orchestrated by the QC Notebook. These enhancements include the following:
- X

### Testing
- [This Terra job](todo) shows an example run of the pipeline for the 1KGP cohort with this change.
- [This TSV file](todo) shows contains an example `qc_matrix` output from the 1KGP cohort.
- Validated all WDLs with womtool.

### Pre-Merge Changes Required
Remove automated Dockstore image sync for development branch.

### DIFF
--- a/src/sv-pipeline/scripts/make_evidence_qc_table.py
+++ b/src/sv-pipeline/scripts/make_evidence_qc_table.py
@@ -10,6 +10,7 @@ from pathlib import Path
 import numpy as np
 
 ID_COL = "#ID"
+NEW_ID_COL = "sample_id"
 EMPTY_OUTLIERS = "EMPTY_ROWS_DROP"
 
 
@@ -24,6 +25,20 @@ def read_ploidy(filename: str) -> pd.DataFrame:
     df_ploidy = pd.read_csv(filename, sep="\t")
     df_ploidy.loc[round(df_ploidy["chrX_CopyNumber"]) < 2, "chrX_CopyNumber_rounded"] = 1
     df_ploidy.loc[round(df_ploidy["chrX_CopyNumber"]) >= 2, "chrX_CopyNumber_rounded"] = 2
+    return df_ploidy
+
+
+def read_sex_assignments(filename: str) -> pd.DataFrame:
+    """
+    Args:
+        filename: A tab-delimited file containing estimated copy numbers.
+    Returns:
+        A pandas DataFrame containing the following columns:
+        [id, chr1_CopyNumber, ..., chr22_CopyNumber, chrX_CopyNumber, chrY_CopyNumber, chrX_CopyNumber_rounded].
+    """
+    df_ploidy = pd.read_csv(filename, sep="\t")
+    df_assignments = pd.read_csv(filename, sep="\t")
+    df_assignments = df_assignments[[ID_COL, "Assignment"]]
     return df_ploidy
 
 
@@ -156,6 +171,7 @@ def read_all_outlier(outlier_manta_df: pd.DataFrame, outlier_melt_df: pd.DataFra
 
 def merge_evidence_qc_table(
         filename_estimated_cn: str,
+        filename_sex_assignments: str,
         filename_mediancov: str,
         filename_wgd: str,
         filename_cnv_qvalues: str,
@@ -172,6 +188,7 @@ def merge_evidence_qc_table(
     serialized to the given output filename.
     """
     df_ploidy = read_ploidy(filename_estimated_cn)
+    df_sex_assignments = read_sex_assignments(filename_sex_assignments)
     df_bincov_median = read_bincov_median(filename_mediancov)
     df_wgd_scores = read_wgd_scores(filename_wgd)
     df_non_diploid = read_non_diploid(filename_cnv_qvalues)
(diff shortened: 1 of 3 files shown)


kjaisingh:
> What's the reason for changing `NaN` to 0?

@mwalker174 Some samples have a value of NaN for certain metrics (e.g. `high_overall_outliers`) - though in reality, these are simply reflect a value of 0. In the QC notebook, we plot distributions of various metrics, but still want to include samples for which the metric value is 0. To achieve this, we currently convert these NaN values to 0 in the QC notebook itself, but the thought here is that we could convert these to 0 in the EvidenceQC output table to minimize notebook-specific table processing.

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -186,14 +204,16 @@ def merge_evidence_qc_table(
...
-           df_melt_high_outlier, df_wham_high_outlier, df_total_high_outliers,
+    dfs = [df_ploidy, df_sex_assignments, df_bincov_median, df_wgd_scores, df_non_diploid,
+           df_manta_high_outlier, df_melt_high_outlier, df_wham_high_outlier, df_total_high_outliers,
            df_manta_low_outlier, df_melt_low_outlier, df_wham_low_outlier, df_total_low_outliers,
            df_melt_insert_size]
     for df in dfs:
         df[ID_COL] = df[ID_COL].astype(object)
     output_df = reduce(lambda left, right: pd.merge(left, right, on=ID_COL, how="outer"), dfs)
     output_df = output_df[output_df[ID_COL] != EMPTY_OUTLIERS]
+    output_df = output_df.replace([None, np.nan], 0.0)
```
epiercehoffman, line 215: We do want to set samples who are not outliers in any categories to 0 in the outlier columns, rather than NaN, for interpretability. But I think it makes more sense to do that for just the outlier columns, not the entire dataframe. There may not be other metrics which routinely have NaNs at the moment, but if an NaN did show up in another metric (current or future), it would not necessarily make sense to set it to 0

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -27,6 +28,21 @@ def read_ploidy(filename: str) -> pd.DataFrame:
     return df_ploidy
 
 
+def read_sex_assignments(filename: str) -> pd.DataFrame:
+    """
+    Args:
+        filename: A tab-delimited file containing estimated copy numbers.
+    Returns:
+        A pandas DataFrame containing the following columns:
+        [id, chr1_CopyNumber, ..., chr22_CopyNumber, chrX_CopyNumber, chrY_CopyNumber, chrX_CopyNumber_rounded].
```
epiercehoffman, line 37: I assume this documentation was a copy/paste error and should be updated to refer to the sex assignment data?

src/sv-pipeline/scripts/make_evidence_qc_table.py
```diff
@@ -186,14 +204,16 @@ def merge_evidence_qc_table(
...
+    dfs = [df_ploidy, df_sex_assignments, df_bincov_median, df_wgd_scores, df_non_diploid,
+           df_manta_high_outlier, df_melt_high_outlier, df_wham_high_outlier, df_total_high_outliers,
            df_manta_low_outlier, df_melt_low_outlier, df_wham_low_outlier, df_total_low_outliers,
            df_melt_insert_size]
     for df in dfs:
         df[ID_COL] = df[ID_COL].astype(object)
     output_df = reduce(lambda left, right: pd.merge(left, right, on=ID_COL, how="outer"), dfs)
     output_df = output_df[output_df[ID_COL] != EMPTY_OUTLIERS]
+    output_df = output_df.replace([None, np.nan], 0.0)
+    output_df.rename(columns={ID_COL: NEW_ID_COL}, inplace=True)
```
epiercehoffman, line 216: Is there a reason for making this change after the fact, rather than updating the value of ID_COL?

Side note that it may be worth asking around if people have strong feelings about using # to indicate the header line in this file. I don't think it's necessary for this file, and I think you have the right idea about updating it for concordance with other files and ease of use with pandas, but it is a common convention in the field, which is why the original header had #ID.

epiercehoffman (changes requested):
Thanks for making these changes! I have a few implementation notes

src/sv-pipeline/scripts/make_evidence_qc_table.py
kjaisingh: Yes, this was done mainly so that upstream files did not have to be changed to the new ID naming standard.

Now that you mention that this is a common convention though, I'm wondering if we can/should just leave it as is?

kjaisingh: Updated all upstream tables to use `sample_id` - have hence removed this renaming from the script.

epiercehoffman (commented):
Thanks for making these changes, and for your patience waiting for a second review! This is overall looking good, but while looking at the outputs I noticed there is an extra line at the end of the QC table with `sample_id` and a row of nans/zeroes. Can you look into this?

kjaisingh:
> Thanks for making these changes, and for your patience waiting for a second review! This is overall looking good, but while looking at the outputs I noticed there is an extra line at the end of the QC table with `sample_id` and a row of nans/zeroes. Can you look into this?

@epiercehoffman just triggered a new job that should resolve this - thanks for flagging (the additional row wasn't appearing in my local runs of the script).

epiercehoffman (approved):
Thanks for the fix and your work on this! Good to merge once you revert the dockstore.yml change

PR #738, opened 2024-10-25
Leave empty genotypes in CleanVcf part 5

mwalker174:
Addresses a bug where empty genotypes, corresponding to samples with unassigned sex, are erroneously overwritten in `CleanVcf` part 5.


epiercehoffman (approved):
Thanks for the fix!

PR #747, opened 2024-11-12
Replace relative paths with abs paths to JS components

VJalili:
Relative paths used for referencing components break version docs, so I am replacing them with ABS paths for portability.


mwalker174 (approved):
Looks ok to me!

VJalili:
Thanks, @mwalker174!

PR #757, opened 2024-12-06
added gnomad v4 annotation file, prefix, and additional subpopulations

knathamuni:
Updated the external_af_population list, external_af_ref_bed, and external_af_ref_bed_prefix with the gnomad v4 population list, AF file, and file prefix.


CuriousTim:
@knathamuni You should change the path of the frequency file to a public one.

epiercehoffman (approved):
Looks good! Thanks for putting together the updated gnomAD AF file, and thanks for your patience as we worked through the changes last week.

PR #765, opened 2025-01-06
Update VM spot pricing in monitoring analysis scripts

mwalker174:
Updates cost estimates with VM price increases from last year for custom machine types: https://cloud.google.com/compute/cmt-pricing-announcement


epiercehoffman (commented):
Looks like this needs to move over to gatk-sv-internal. Also, I can't find these numbers on the page you linked, was there another source?

PR #768, opened 2025-01-09
Make functional annotation optional if no GTF/BED provided

epiercehoffman:
### Updates
Toggle functional annotation in AnnotateVcf by not providing a GTF or noncoding BED file. I wanted this for some processing I was doing, and I think Harrison requested this ability as well. 

### Testing
* Validated all WDLs & JSONs with womtool and Terra validation script
* Tested AnnotateVcf with and without GTF and ran as expected

Changes to the dockstore.yml were for testing only and will be reverted before merging

### DIFF
--- a/wdl/ShardedAnnotateVcf.wdl
+++ b/wdl/ShardedAnnotateVcf.wdl
@@ -17,7 +17,7 @@ workflow ShardedAnnotateVcf {
     String prefix
     String contig
 
-    File protein_coding_gtf
+    File? protein_coding_gtf
     File? noncoding_bed
     Int? promoter_window
     Int? max_breakend_as_cnv_length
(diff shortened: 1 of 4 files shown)


wdl/ShardedAnnotateVcf.wdl
```diff
@@ -17,7 +17,7 @@ workflow ShardedAnnotateVcf {
     String prefix
     String contig
 
-    File protein_coding_gtf
+    File? protein_coding_gtf
```
mwalker174, line 20:
```suggestion
   # At least one of protein_coding_gtf or noncoding_bed is required for functional annotation
    File? protein_coding_gtf
```
Or something of the like

mwalker174 (approved):
Looks good, maybe just add a comment describing the logic

PR #771, opened 2025-01-21
Add VCF input option to Vapor WDL

mwalker174:
- Enables VCF inputs in addition to BED in the Vapor workflow.
- Adds Vapor.wdl to Dockstore yaml.
- New utility method `SubsetVcfToSample` for subsetting to a single sample from a `String`. This is nearly identical to `SubsetVcfBySamplesList` which takes a samples list file instead.
- Deletes unnecessary `prefix` input since sample names usually suffice for file naming.

Tested successfully in Terra on 8 samples.

### DIFF
--- a/.github/.dockstore.yml
+++ b/.github/.dockstore.yml
@@ -199,6 +199,15 @@ workflows:
         - /.*/
 
   - subclass: WDL
+    name: Vapor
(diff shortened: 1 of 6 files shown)


epiercehoffman (approved):
All looking good

PR #773, opened 2025-01-30
Small fixes to sample QC notebook

epiercehoffman:
Updates include:
* Small documentation clarifications
* Display sample_set_tbl even if not reloading or editing included batches
* Save new sample_set_tbl in the same cell as editing batches 
* Don't comment cells to be run once, since they should be run the first time
* Only put "(Filtered)" in plot title during filtering
* Print number of differences between PED files
* Collapsed a few more headings by default

Testing:
Tested in clone of public workspace on reference panel data, with and without reference PED file. Cleared outputs and tested re-upload

### DIFF
--- a/scripts/notebooks/SampleQC.ipynb
+++ b/scripts/notebooks/SampleQC.ipynb
@@ -1642,7 +1642,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "<div class=\"alert alert-block alert-success\"><b>Optional</b>: If you wish to only include a subset of batches out of the ones listed above, you can filter <tt>sample_set_tbl</tt> to only include them using the cell below. Ensure that all batches you expect are included.</div>"
+    "<div class=\"alert alert-block alert-success\"><b>Optional</b>: If you wish to only include a subset of batches out of the ones listed above, you can filter <tt>sample_set_tbl</tt> to only include them using the cell below. Ensure that all batches you expect are included by running the following cell to display them.</div>"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


scripts/notebooks/SampleQC.ipynb
```diff
@@ -1631,7 +1652,9 @@
    "outputs": [],
    "source": [
     "# Optionally filter to only include a subset of batches from sample_set_tbl and update the saved file. For example:\n",
-    "sample_set_tbl = sample_set_tbl[sample_set_tbl['entity:sample_set_id'].str.contains('KJ_EvidenceQC_Updates')]"
+    "sample_set_tbl = sample_set_tbl[sample_set_tbl['entity:sample_set_id'].str.contains('KJ_EvidenceQC_Updates')]\n",
+    "file_path = generate_file_path(TLD_PATH, 'artifacts', 'sample_sets_qc.tsv')\n",
+    "save_df(WS_BUCKET, file_path, sample_set_tbl)"
```
kjaisingh, line 1657: Can we remove the reference to `KJ_EvidenceQC_Updates`? This must've been missed in the first pass.

kjaisingh: If this is removed, I'm not sure if we'd still want to move the two lines that now follow it (`generate_file_path` and `save_df`) into the same cell. Fear that folks will simply ignore the comment to optionally filter to only include a subset of batches.

Alternatively, if we wish to keep this in a single cell, maybe you can add 1-2 new lines between that comment and the two lines that now follow it (`generate_file_path` and `save_df`)?

epiercehoffman: Could add a green box and a new variable SUBSTRING and only run the following lines if SUBSTRING is not None?

scripts/notebooks/SampleQC.ipynb
```diff
@@ -26,14 +26,15 @@
...
     "\n",
     "**Execution Tips**:\n",
-    "* The first time you start this notebook (one time only), you will need to uncomment and run the package installation cell under *Imports*.\n",
+    "* The first time you start this notebook (one time only), you will need to run the package installation cell under *Imports*. Then, we recommend commenting out the package installation to skip it next time.\n",
     "* Once the packages are installed, to quickly run all the cells containing helper functions, constants, and imports, skip to the first cell of *Data Ingestion*, click \"Cell\" in the toolbar at the top of the notebook, and select \"Run All Above.\" Then, starting from *Data Ingestion*, proceed step-by-step through the notebook.\n",
     "* The keyboard shortcut to run a cell is `Shift`+`Return`.\n",
-    "* The keyboard shortcut to comment or uncomment an enitre cell is `Command`+`/` or `Control`+`/`."
+    "* To select all the lines of code in a cell, click on the cell and then click `Command`+`A` or `Control`+`A`.\n",
+    "* The keyboard shortcut to comment or uncomment all selected lines is `Command`+`/` or `Control`+`/`."
```
kjaisingh, line 37: Minor - but could be helpful to add a note next to this that indicates it's particularly applicable for blue cells after running them once? 

Primarily for users without much experience with python notebooks etc, so that they can kick off their notebook on repeat occasions without having to run cells individually (i.e. they can just do `Run All`).

kjaisingh (approved):
Looks good to me, thanks for these cleanups - just one minor comment about the formatting relating to subsetting the batches to process.

scripts/notebooks/SampleQC.ipynb
kjaisingh: Sounds good.

kjaisingh:
Looks great now, thanks!

PR #775, opened 2025-02-07
Fix bug in ClusterBatch when SR2POS not present

epiercehoffman:
With the latest version, ClusterBatch failed during GatkToSvtkVcf due to SR2POS being missing. This PR updates the check to one that does not fail if SR2POS is not in the header. Tested on a clone of the joint-calling workspace with an updated docker and the same inputs that failed before.

### DIFF
--- a/src/sv-pipeline/scripts/format_gatk_vcf_for_svtk.py
+++ b/src/sv-pipeline/scripts/format_gatk_vcf_for_svtk.py
@@ -153,7 +153,7 @@ def convert(record: pysam.VariantRecord,
         new_record.info['STRANDS'] = '+-'
         # END information is lost when setting POS=END, so we need to set it to SR2POS if it's available
         # Note that the END position is only important following SR breakpoint refinement in FilterBatch
-        if record.info.get('SR2POS') is not None:
+        if record.info.get('SR2POS', None) is not None:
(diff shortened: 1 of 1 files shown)


src/sv-pipeline/scripts/format_gatk_vcf_for_svtk.py
```diff
@@ -153,7 +153,7 @@ def convert(record: pysam.VariantRecord,
         new_record.info['STRANDS'] = '+-'
         # END information is lost when setting POS=END, so we need to set it to SR2POS if it's available
         # Note that the END position is only important following SR breakpoint refinement in FilterBatch
-        if record.info.get('SR2POS') is not None:
+        if 'SR2POS' in record.info:
```
mwalker174, line 156:
```suggestion
        if 'SR2POS' in record.info and record.info['SR2POS'] is not None:
```
I would check for both that the key doesn't exist but also if the key exists but also that the value is not null to be safe

mwalker174:
Did you test this on CombineBatches as well?

epiercehoffman:
@mwalker174 Updated with your suggestion and retested ClusterBatch & CombineBatches with the new docker

mwalker174 (approved):
Thanks!

PR #779, opened 2025-02-21
Fix coordinates issues for complex SV refinement

CuriousTim:
1. The input BED file to the CPX BED reformatting script has 0-based, half-closed variant coordinates while the SOURCE coordinates are 1-based, fully-closed. However, the reformatting script assumes the coordinate systems are the same. The first fix adds a step to adjust the variant coordinates in the BED file to be 1-based, fully-closed prior to executing the reformatting script.
2. The reformatting script does not handle the case where the origin of the inserted fragment shares a breakpoint with the insertion point. The second fix adds an equality check to the conditional comparing breakpoints.


CuriousTim:
@epiercehoffman @mwalker174 @kjaisingh Could you take a look?

epiercehoffman (approved):
Thanks for implementing these fixes and being responsive about testing! Looks good to merge.

Stylistically, I'd prefer to handle the coordinate conversion from 0-based to 1-based in the script, but given the state of the script and our intentions to port this logic into GATK, it's not worth the trouble.

PR #781, opened 2025-02-27
Update the OS version of the Action runners

VJalili:
We're currently using `ubuntu-20.04` in some of the actions (the rest are set to always use the latest version), which is set for deprecation in April 2025. 

This PR updates the pinned version to the current latest, 24.04.


mwalker174 (approved):
Thanks!

PR #784, opened 2025-03-07
Add the de novo pipeline to Dockstore 

VJalili:
The PR updates Dockstore integration so the _de novo_ pipeline's workflows are auto-synched.


cwhelan (approved):
I think it's fine to put this pipeline in dockstore.

PR #789, opened 2025-03-14
Update Docker build CI/CD action 

VJalili:
This PR is a follow-up from https://github.com/broadinstitute/gatk-sv/pull/781 and https://github.com/broadinstitute/gatk-sv/pull/785 where the OS version of the runners was updated to the current latest version of Ubuntu, and it updates the actions to their current latest version to account for their interface update. 

 Specifically: 
- [x] Update actions to their current version since the old versions did not work properly on the latest Ubuntu;
- [x] GCP login is updated to match the current method. Note that we're using the long-lived tokens method instead of the implicit flow (the recommended approach) for two reasons: (1) The tokens generated in the recommended approach are valid for 15min only, and since the GATK-SV's Docker build can take longer than 15min, the tokens are expired when pushing images; and (2) The recommended flow's GCP/IAM configuration has conflicts with the current setting of the GCP project GATK-SV uses. 
- [x] All the credentials are updated to match the requirements of the set versions of the actions.

### DIFF
--- a/.github/workflows/sv_pipeline_docker.yml
+++ b/.github/workflows/sv_pipeline_docker.yml
@@ -30,7 +30,7 @@ jobs:
       image_tag: ${{ steps.image_tag.outputs.IMAGE_TAG }}
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           # By default, this checks out only the current commit;
           # however, since a diff between the current commit and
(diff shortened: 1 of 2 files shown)


.github/workflows/sv_pipeline_docker.yml
```diff
@@ -155,30 +155,34 @@ jobs:
...
+        uses: azure/docker-login@v2
         with:
           login-server: ${{ secrets.AZ_CR }}
           username: ${{ secrets.AZ_USERNAME }}
           password: ${{ secrets.AZ_PASSWORD }}
 
+      - name: GCP login
+        uses: google-github-actions/auth@v2.0.0
+        with:
+          credentials_json: '${{ secrets.GCP_SA_CREDENTIALS }}'
```
cwhelan, line 180: These are already defined somewhere?

VJalili: Yes, these are defined as environment secrets in the repository's configuration.

PR #795, opened 2025-04-08
Use `communicate()` with Python subprocess

CuriousTim:
Using `poll()` on a Popen object will return `None` if the child process has not terminated and that can cause trouble with the `get_command_output()` function in the Docker building script. The function assumes that `poll()` always returns the exit code of the child process and compares it to 0 to check if the child process executed successfully. If the child process has not terminated by the time `poll()` is called, then the comparison will always be false and the child process will appear to have failed. `communicate()` will wait until the child process terminates if a timeout is not given.

An error from using `poll()` was observed with Python 3.12.3


VJalili:
Sounds good. That image might be deprecated; generally, the `xx.04` images are maintained for an extensive period. That image is also referenced in the [docs](https://broadinstitute.github.io/gatk-sv/docs/advanced/docker/manual); we may need to update the image there as well.

VJalili (approved):

PR #796, opened 2025-04-10
Scale down GQ in CombineBatches

mwalker174:
Addresses a bug introduced in #732 where all genotype qualities (GQ) were clipped at 99 in `CombineBatches` (and subsequently to 9 in `CleanVcf`). Note that GQ is recalibrated during `FilterGenotypes`, so final GQs had appeared normal. However, in a test run we found that the clipping had modest effect on recalibration scores for most variant types (see attached figures).

To resolve the clipping, this PR moves GQ rescaling from `CleanVcf` to a formatting step preceding `gatk SVCluster` in `CombineBatches`.

Screenshot 2025-04-10 at 10 49 45 AM

### DIFF
--- a/wdl/CleanVcfChromosome.wdl
+++ b/wdl/CleanVcfChromosome.wdl
@@ -323,7 +323,6 @@ workflow CleanVcfChromosome {
     input:
       vcf=FinalCleanup.final_cleaned_shard,
       ploidy_table=ploidy_table,
-      args="--scale-down-gq",
       output_prefix="~{prefix}.final_format",
       script=svtk_to_gatk_script,
       sv_pipeline_docker=sv_pipeline_docker,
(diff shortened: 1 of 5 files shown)


epiercehoffman (approved):
Thanks for identifying and fixing this bug. The fix seems elegant. 

I did a quick scan for places between CombineBatches and CleanVcf that might expect a GQ in the range [0,999] and didn't find anything obvious. I did find a few places where we've documented that GQ is [0,999] until CleanVcf so we should update those to refer to CombineBatches instead:

- https://github.com/broadinstitute/gatk-sv/blob/main/wdl/MainVcfQc.wdl#L29
- https://github.com/broadinstitute/gatk-sv/blob/main/website/docs/modules/main_vcf_qc.md?plain=1#L216
- https://github.com/broadinstitute/gatk-sv/blob/main/wdl/MakeCohortVcf.wdl#L74

PR #813, opened 2025-05-20
Update image links to address page scrolling issues

VJalili:
This addresses initially discussed at https://github.com/broadinstitute/gatk-sv/pull/809

Generally, two issues contribute to the scrolling imperfections such as this:

- Use of parentheses in headers.
- Incorrect syntax of using images and hosting images in external repositories (details: https://github.com/facebook/docusaurus/issues/9520 and https://github.com/facebook/docusaurus/issues/3081).

We need a thorough update on the section titles and images used. We will do the update after we have updated the content and removed the images not needed, or pointed to the Terra docs where possible. 

This PR only focused on fixing the link mentioned in https://github.com/broadinstitute/gatk-sv/pull/809.


epiercehoffman (approved):
Thanks for figuring this out, the images look fine and when I tested the batching link it scrolled to the right place.

mwalker174 (approved):

Issue #818, opened 2025-06-07
The stitch_fragmented_CNVs.sh script failed to produce the expected results


zhangting1991:
Hi,

Thank you for providing the GATK-SV pipeline — it's been incredibly helpful in our work.

I encountered a potential issue with the stitch_fragmented_CNVs.sh script during the CNMOPS-Large step. Specifically:

  - The input VCF file does not contain the EVIDENCE field in the INFO column.
  - As a result, during the substep "Subset VCF to biallelic CNVs without PE or SR support", the script generates an empty biallelic_depth_CNVs.vcf.gz file with no variants.
  - Since no error is raised at this step, the pipeline continues execution.
  - However, no stitched VCF is produced, and the final output is effectively identical to the input.

This behavior might be unintended. Please let me know if this is expected, or if there's a recommended fix or workaround.

Thanks again for your great work!

PR #821, opened 2025-06-09
fix no EVIDENCE bug in stitch_fragmented_CNVs.sh (#818)

xiaoli0:
The input VCF lacks the "EVIDENCE" field in the INFO column. As a result, during the step "Subset VCF to biallelic CNVs without PE or SR support", the script will produce an empty biallelic_depth_CNVs.vcf.gz file containing no variants. This step does not raise an error or halt execution, so the pipeline continues; however, no stitched VCF is generated. Therefore, the script's output will be identical to the input.


epiercehoffman:
Thanks for contributing! Per my response to #818, this change will require extensive testing, and we expect defragment_cnvs.py is making up for this, so we may not be able to prioritize this right away.

PR #822, opened 2025-06-09
Keep EV as int when reformatting legacy genotyped vcfs

epiercehoffman:
### Updates

When reformatting legacy genotyped VCFs for CombineBatches 1.0.2+, the reformatting script converted the integer EV field to a string. This conversion was premature, as it is done in CleanVcf, so this resulted in a failure in CleanVcf. This PR removes the portion of the reformatting script that alters EV.

### Testing
* Tested with the updated docker in the cohort where the failure was observed, and CombineBatches --> CleanVcf succeeded
* Validated all WDLs and JSONs with womtool & Terra validation script


mwalker174 (approved):
Thank you @epiercehoffman

PR #862, opened 2025-08-28
Annotate retrotransposon deletions with a filter tag

kjaisingh:
### Description
This PR is intended to annotate retrotransposon deletions with a new `RETRO_DEL` filter tag. These are deletions which have their breakpoints with 8 b.p. of an intron, where the introns are provided through a reference file.

### Testing
- The [following job](https://app.terra.bio/#workspaces/broad-firecloud-dsde-methods/GATK-Structural-Variants-Joint-Calling/submission_history/7ab5b73d-40b9-4609-9e9f-f36cc89f1120) represents a run of the original WDL.
- The [following job](https://app.terra.bio/#workspaces/broad-firecloud-dsde-methods/kj-1kgp-annotate-retro-del/submission_history/47c6275f-273b-48f8-9658-ea47e4901e30) represents a run of the modified WDL with no input parameter passed - as you can see, the output VCF now contains several variants tagged with `RETRO_DEL` in the filter column.
- Validated all WDLs with `womtool`.


src/svtk/svtk/utils/__init__.py
```diff
@@ -1,6 +1,5 @@
 from .utils import *
 from .bgzipfile import BgzipFile
-from .helpers import reciprocal_overlap, overlap_frac
```
mwalker174: Is this necessary? Doesn't seem related to the PR

PR #864, opened 2025-09-02
Modify genotype filtering to read cutoffs from generalizable table 

kjaisingh:
### Description
This PR is intended to enable the use of a generalizable TSV file that contains cutoffs stratified by SV type, minimum size and maximum size when filtering genotypes.

### Testing
- The [following job](X) represents a run of the original WDL.
- The [following job](X) represents a run of the modified WDL with the original cutoffs from the job above passed, albeit in the form of a table rather than named arguments.
- Validated all WDLs with `womtool`.

### Pre-Merge Changes Required
Remove automated sync of the `FilterGenotypes` WDL to Dockstore.

### DIFF
--- a/.github/.dockstore.yml
+++ b/.github/.dockstore.yml
@@ -177,6 +177,7 @@ workflows:
     filters:
       branches:
         - main
+        - kj_modify_sl_filter_input_format
       tags:
         - /.*/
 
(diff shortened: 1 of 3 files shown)


wdl/TrainGenotypeFilteringModel.wdl
```diff
@@ -0,0 +1,176 @@
+version 1.0
+
+import "Structs.wdl"
+import "RecalibrateGq.wdl" as recalibrate_gq
+import "TasksMakeCohortVcf.wdl" as tasks_cohort
+
+workflow TrainGenotypeFilteringModel {
```
mwalker174:
```suggestion
workflow ScoreGenotypes {
```
While there is some "training" functionality built into optimizing cutoffs, the real purpose of this workflow is to assign SL scores

mwalker174 (changes requested):
Just one minor request

PR #937, opened 2026-04-30
Fix Stripy VCF sample ID

mwalker174:
Addresses an assumption about sample ID naming in the Stripy workflow. Stripy now requires `--sample-id` to be provided at the CLI and uses that for setting the sample ID in the output vcf rather than slicing it from the BAM/CRAM name directly.

Tested successfully on a cram with a different filename format.

TODO: revert dockstore.yml and dockers.

### DIFF
--- a/.github/.dockstore.yml
+++ b/.github/.dockstore.yml
@@ -240,6 +240,7 @@ workflows:
     filters:
       branches:
         - main
+        - mw_stripy_sample_id_fix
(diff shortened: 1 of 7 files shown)


VJalili (approved):
Thanks for the fix, @mwalker174. Tested this change in sv-shell using `RGP_1153_3` sample, and it worked as expected.

PR #945, opened 2026-06-24
Add Genomic Disorder CNV calling and VCF integration

mwalker174:
This PR adds Genomic Disorder (GD) CNV calling and VCF integration, including new WDL modules, pipeline wiring, Terra/test templates, and documentation. It introduces a new post-processing path that calls GD CNVs from BAF + read-depth evidence, then integrates those calls into final annotated VCF outputs for both cohort and single-sample workflows.

### 1) New GD calling workflow
- Adds new WDL: `wdl/CallGenomicDisorderCNVs.wdl`
  - Rebins read-depth input with `CondenseReadCounts`
  - Runs `gatk-sv-gd` via `run_gd.sh`
  - Emits `gd_output_tarball` with GD outputs

### 2) New GD-to-VCF integration workflow
- Adds new WDL: `wdl/IntegrateGDVcf.wdl`
  - Extracts/combines `gd_cnv_calls.tsv.gz` across batches
  - Combines ploidy tables
  - Scatters by contig and runs `gatk-sv-gd integrate`
  - Re-concatenates per-contig VCF shards into integrated VCF output

### 3) Single-sample pipeline now runs GD + integrates into annotation path
- `wdl/GATKSVPipelineSingleSample.wdl` now:
  - Imports and calls `CallGenomicDisorderCNVs` and `IntegrateGDVcf`
  - Adds `ConcatBaf` task to derive case BAF from SD evidence and merge with ref panel BAF
  - Uses integrated VCF (when available) as input to `AnnotateVcf`
  - Exposes `gd_output_tarball` in workflow outputs

### 4) BAF generation utility added
- New script: `src/sv-pipeline/scripts/baf_from_sd.py`
  - Merge-joins SD and loci VCF
  - Computes ALT fraction BAF per site
  - Filters homozygous/uninformative sites (0 or 1)
  - Outputs tabix-friendly BAF table

### 5) `CondenseReadCounts` generalized for matrix inputs
- `wdl/CollectCoverage.wdl` updates:
  - Handles both Picard-like counts and `.rd.txt.gz` matrix formats
  - Supports optional `output_prefix`, `sequence_dictionary`
  - Produces matrix-style or Picard-style output appropriately
  - Improves runtime/disk sizing and runtime blocks
  - Adds `CondenseReadCountsMatrix` task variant

### 6) Terra/workspace/test templates expanded for GD
- New workflow config templates:
  - `CallGenomicDisorderCNVs.json.tmpl`
  - `IntegrateGDVcf.json.tmpl`
[...]

PR #958, opened 2026-09-09
Use HTTPS CRAN mirrors in sv-base-virtual-env (fixes image-build CI)

mwalker174:
## Problem
Image-build CI currently fails whenever the `sv-base-virtual-env` / `sv-pipeline-virtual-env` R layers are rebuilt, e.g. [run 34387428118](https://github.com/broadinstitute/gatk-sv/actions/runs/34387428118?pr=945) on #945:

```
Error: (converted from warning) unable to access index for repository http://lib.stat.cmu.edu/R/CRAN/src/contrib:
  cannot open URL 'http://lib.stat.cmu.edu/R/CRAN/src/contrib/PACKAGES'
```

The plain-HTTP mirror `http://lib.stat.cmu.edu/R/CRAN/` now 301-redirects to HTTPS (and 403s from some networks). `install_R_packages.R` / `install_bioconductor_packages.R` run with `options(warn = 2)`, so R's normally-tolerable "unreachable first repo, trying fallback" warning becomes fatal — the listed `cran.rstudio.com` fallback never gets used. This is infrastructure bit-rot: main is green only because no image rebuild has run since the mirror changed behavior.

## Fix
Point all three CRAN repo references in `dockerfiles/sv-base-virtual-env/` at `https://cloud.r-project.org` (CDN) with `https://cran.rstudio.com` as fallback — the same remedy already documented and used in `dockerfiles/sv-shell/Dockerfile` (see its comment about cran.us.r-project.org redirecting to CMU/403).

Note: the failing CI layer executes the install scripts baked into the **pinned** `sv-base-virtual-env:2026-05-30-v1.1-0639be96` image. Because this PR changes `dockerfiles/sv-base-virtual-env/`, `build_docker.py` will rebuild `sv-base-virtual-env` from source in CI and downstream layers will use the fixed scripts; a regular release build will republish the pin.

## Tests
- `curl -I https://cloud.r-project.org/src/contrib/PACKAGES` → HTTP/2 200
- Verified the three edited locations are the only live `lib.stat.cmu.edu` repo references in `dockerfiles/` on `main`
- Full validation is the image-build CI workflow on this PR (it will rebuild `sv-base-virtual-env` + dependents)

Related: unblocks the docker rebuild currently failing on #945 (that PR's own change is unrelated: it adds `pyro-ppl` to the conda env.

### DIFF
--- a/dockerfiles/sv-base-virtual-env/Dockerfile
+++ b/dockerfiles/sv-base-virtual-env/Dockerfile
@@ -56,7 +56,7 @@ COPY dockerfiles/sv-base-virtual-env/union_lists.sh $R_INSTALL_BIN/
 ENV PATH=$R_INSTALL_BIN:/opt/bin:$PATH
 
 # update the installed R packages
-RUN Rscript -e "update.packages(contriburl=contrib.url(repos=c('http://lib.stat.cmu.edu/R/CRAN/', 'https://cran.rstudio.com')), clean=TRUE, quiet=TRUE, ask=FALSE)"
+RUN Rscript -e "update.packages(contriburl=contrib.url(repos=c('https://cloud.r-project.org', 'https://cran.rstudio.com')), clean=TRUE, quiet=TRUE, ask=FALSE)"
 # install the required R packages for this stage
 ARG SV_BASE_R_PKGS="optparse BiocManager"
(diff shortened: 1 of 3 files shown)


VJalili (approved):
LGTM, thank you for fixing this!

PR #961, opened 2026-09-16
Train batch genotyping on the batch's own calls and gate on cutoff quality

### LINKED ISSUE #965: Single-sample GD BAF step cannot run since #945: task ConcatBaf calls /gatk/gatk inside the sv_pipeline_docker image
### The single-sample GD BAF step cannot run: `task ConcatBaf` invokes `/gatk/gatk` inside the `sv_pipeline_docker` image

Found while running `GATKSVPipelineSingleSample` on Terra (NA12878) on 2026-09-17/18. It is not a
silent edge case — the workflow dies in `call-ConcatBafCase` with `rc=127`:

```
/mnt/disks/cromwell_root/script: line 46: /gatk/gatk: No such file or directory
```

The step immediately before it succeeds (`baf_from_sd.py`: `53085495 loci matched, 3473862 BAF values
kept`), so inputs/access are fine; only the second step's binary is missing.

### Cause

`wdl/GATKSVPipelineSingleSample.wdl` — `task ConcatBaf` runs the GATK tool `PrintSVEvidence` via the
**gatk-image** launcher while declaring the **sv-pipeline** image:

```wdl
    /gatk/gatk --java-options "-Xmx2g" PrintSVEvidence \
      -F evidence.list \
      --sample-names samples.list \
      -O "~{batch}.baf.txt.gz"
    ...
    runtime {
      docker: sv_pipeline_docker      # <- this image has no /gatk/gatk
```

`PrintSVEvidence` is a GATK tool (`gatk/src/main/java/org/broadinstitute/hellbender/tools/sv/PrintSVEvidence.java`),
and the `sv-base` → `sv-pipeline` chain ships GATK as a **jar**, not as a `gatk` launcher:

```dockerfile
# dockerfiles/sv-base/Dockerfile
ARG GATK_JAR="/opt/gatk.jar"
```

I checked `dockerfiles/sv-base/Dockerfile` and `dockerfiles/sv-pipeline/Dockerfile` at `v1.1`, `v1.1.1`,
`main` and `mw_genotype_scale`: **no** rule creates `/gatk` or a `gatk` launcher in any of them. So this
call has never been executable in any published image, rather than having been broken later by an image change.

Every other `PrintSVEvidence` call site in the repo is consistent about this split:

| call site | how it invokes the tool | `docker:` |
|---|---|---|
| `wdl/BatchEvidenceMerging.wdl:164` (`MergeEvidence`, `SDtoBAF`) | `/gatk/gatk --java-options …` | `gatk_docker` |
| `wdl/CollectSVEvidence.wdl` (`RunCollectSVEvidence`) | `/gatk/gatk …` | `gatk_docker` |
| `wdl/CNMOPS.wdl:290`, `wdl/MatrixQC.wdl:153`, `wdl/GenotypeCpxCnvsPerBatch.wdl:245` | `java -Xmx… -jar ${GATK_JAR} …` | sv-pipeline family |
| **`wdl/GATKSVPipelineSingleSample.wdl:1668` (`ConcatBaf`)** | `/gatk/gatk …` | **`sv_pipeline_docker`** ← the odd one out |

### Introduced by

`31d3df3c` — *Add Genomic Disorder CNV calling and VCF integration (#945)*, 2026-09-09. The task, the
`/gatk/gatk` command line and the `docker: sv_pipeline_docker` runtime all land as additions in that one
commit, i.e. the mismatch is original rather than a later refactor. It has been latent on `main` for 9 days.
`git log -S"/gatk/gatk" -- wdl/GATKSVPipelineSingleSample.wdl` returns exactly that commit.

Scope: **single-sample mode only.** `ConcatBaf` exists solely in `GATKSVPipelineSingleSample.wdl`; batch /
cohort mode merges BAF through `BatchEvidenceMerging.wdl`, which runs under `gatk_docker` and is fine.

Why CI does not catch it: the failure is between a command string and the filesystem of the image it is
scheduled in. `miniwdl` and `womtool validate` only see WDL validity, so the repo's WDL gates stay green.

### Suggested fix — one line, using this file's own convention

```diff
-    /gatk/gatk --java-options "-Xmx2g" PrintSVEvidence \
+    java -Xmx2g -jar ${GATK_JAR} PrintSVEvidence \
```

`GATK_JAR` is already an env var in the sv-pipeline image (`sv-base/Dockerfile` `ARG GATK_JAR`) and this is
what `CNMOPS.wdl` / `MatrixQC.wdl` / `GenotypeCpxCnvsPerBatch.wdl` do for the same tool in the same image
family. (The alternative — switching the runtime to `gatk_docker` — would also work, but it changes which
image the merge step runs in and would need the `python`/`tabix` steps of the same command block to exist
there too, which they do not.)

### Reproduction

Workspace `broad-firecloud-dsde-methods/GATK-Structural-Variants-Single-Sample-mw-ploidy` (private),
`GATKSVPipelineSingleSample@main` via Dockstore, `sample/NA12878`; submission
`ad6c8a60-07b7-4162-9fe9-dff9160ff838`, workflow `97006eb7-44a5-45f7-bee2-abc6c48ee2b1`, failed in
`call-ConcatBafCase` attempt-2. Execution directory:
`gs://fc-39abe4ea-f41d-41be-bdd7-24a936018339/submissions/ad6c8a60-07b7-4162-9fe9-dff9160ff838/`.
The same WDL text is present on `mw_genotype_scale`, so a branch run fails identically.

mwalker174:
Train batch genotyping on the batch's own calls and gate on cutoff quality

`TrainSVGenotyping` trained the PE/SR cutoffs over the site list being genotyped,
and fitted RD at a different bin resolution than `GenotypeSVs` applies it at. On a
516-sample AoU batch that produced all-zero SR cutoffs -- which does not loosen SR
background filtering, it disables it outright -- and an RD model whose state-0
spread was an order of magnitude off the reference behaviour.

Training inputs
- New `GenotypeBatch` input `training_vcf`: the batch's own filtered PESR calls
  (pipeline: `GATKSVPipelinePhase1.filtered_pesr_vcf`; cohort mode:
  `${this.outlier_filtered_pesr_vcf}`). Phase-2b site count drops from 4,305,762
  cohort-wide sites (2.8% pass rate, 122 `first_pass_variants`) to 103,483 sites,
  6,574 SR-trainable, 6,065 `first_pass_variants`.
- `n_RD_genotype_bins` (default 100000) is passed as `--num-bins` to **both**
  `TrainSVGenotyping` (fit) and `GenotypeSVs` (apply), so depth is described at one
  resolution on both sides. Before, the fit ran at the Java tool default of 10 while
  apply ran at its default of 100000; v1.1.1 passed `-i` to fit and apply alike.
- `GATKSVPipelineBatch` declares and binds `training_intervals`,
  `depth_exclusion_intervals` and `pesr_exclusion_intervals` instead of leaving
  them unbound and relying on dotted keys in the input JSON, and takes
  `GenotypeBatch.ploidy_table` from the existing `CreatePloidyTableFromPed` call.
  Required inputs become womtool-visible (WDL gives File inputs no default, so an
  unbound one is only discoverable at submission time); the now-redundant dotted
  keys are dropped from the GATKSVPipelineBatch test templates.

Failing loudly instead of emitting zeros
- `TrainSVGenotyping` writes `genotyping_sr_cutoff_diagnostics` and
  `ValidateSRCutoffs` gates the genotyping scatter on it, so a degenerate grid
  fails the workflow instead of applying frequency cutoffs of 0.0 with no warning.
  The gate requires the `rare_selection_status` and `common_selection_status`
  records to be *present*, not merely free of a non-OK value: the training task
  pre-creates the diagnostics file so that a diagnostics write cannot fail the task,
  and an empty or half-written file would otherwise leave nothing to reject.

RD params are split into depth-side and PESR-side tables
- `genotyping_rd_table` -> `genotyping_rd_depth_table` + `genotyping_rd_pesr_table`
  through `RegenotypeCNVs`, `GATKSVPipelineSingleSample`, `SVShell`,
  `sv_shell/genotype_svs.sh` and `sv_shell/single_sample_pipeline.sh` (including its
  later `MakeCohortVcf` block, which takes the depth table, as
  `GATKSVPipelineSingleSample` already did), the `sample_inputs/` fixtures including
  the DRAGEN one, and the ref-panel values (`inputs/values/ref_panel_1kg.json`).

Metric column names aligned with what the Java tools emit
- `BAF_HET_RATIO`, `BAF_DEL_LOGLIK`, `BAF_KS_STAT`, `BAF_KS_Q`,
  `RD_MEDIAN_SEPARATION`, `svsize` in `svtest/cli/metrics_file.py`,
  02e `aggregate.py` and 03 `build_training_set.py` (the last still read the
  v1.1-era `BAF_del_loglik` / `RD_Median_Separation` / `size` names), and
  `adjudicate_sv.py` gains `BAF_DEL_LOGLIK` as a dependent feature.

WHAM deletions and legacy BND END fields
- `GenotypeBatch` gains a `FilterWhamDeletions` task applied to the site VCF *and*
  to the training VCF (wham DELs are needed for adjudication but are enriched for
  false positives, so they should not set cutoffs), and `FilterBatchSites` gains
  the same `bcftools view -e 'ALGORITHMS=="wham" && SVTYPE=="DEL"'` filter.
- `FilterBatchSites` repairs `END2`/`CHR2` on `SVTYPE==BND` records, taking the true
  END from the raw VCF text because pysam silently clamps `END` to `POS` for
  interchromosomal breakends. Legacy-input support, marked TODO for removal. The
  raw-text scan also collects `SVTYPE==CTX`, but no rewrite is applied to CTX: on
  the cohort measured below this step carries 0 CTX records (54,678 manta / 14,684
  scramble / 9,105 wham records, both pipelines), so the BND-only condition is
  unexercised rather than wrong -- say the word if you would rather it were lifted
  to cover CTX than leave the asymmetry.

Copilot review (2026-09-17): 5 findings, 4 fixed, 1 measured
Every claim was checked against the tree and, where testable, against data; all five
were substantively correct, so they are all worth recording:
1. RD table keys were not propagated -- **fixed** (see the split section above). The
   rename left two readers behind, and `jq` turns a missing key into the string
   `null`, so `GenotypeSVs` would have been handed `--rd-depth-table null`.
2. `ValidateSRCutoffs` accepted a diagnostics file with no status records --
   **fixed**, and the fix was tested by extracting the task's real command body from
   the WDL and running it over six fixtures: healthy (passes), empty (blocks), empty
   with `fail_on_degenerate_sr_cutoffs=false` (warns and continues), explicit
   `REJECTED_DEGENERATE_GRID` (blocks, unchanged), one status record absent (blocks),
   truncated write (blocks).
3. `n_RD_genotype_bins` reached the fit but not the apply -- **fixed**; at the default
   it is inert (100000 equals the Java default), so this is a latent-footgun fix.
4. The PR description over-claimed the `END2`/`CHR2` repair (see above) -- wording
   corrected instead of code changed, with the measurement attached.
5. The PR description said the `.dockstore.yml` branch filters were dropped --
   corrected below; they are in the diff.

Branch publishing (`.github/.dockstore.yml`)
`mw_genotype_scale` is added to the `filters.branches` of the 5 workflows in this
file. Entries are `mode: DOCKSTORE_YML`, and Dockstore publishes a branch ref only if
a filter matches it, so without these lines no `@mw_genotype_scale` version exists to
reference. That is not cosmetic: it is how this branch's `GenotypeBatch` was published
for the Terra rerun below. (Related, and already merged separately in #962: `main`'s
copy of this file was invalid YAML, which made Dockstore abort the whole file and stop
publishing all 32 workflows for every ref -- silently, with the only trace in the
GitHub App event logs.)

Image pins (`inputs/values/dockers.json`) -- 3 images, all on the production path
- `gatk_docker` -> `gatk-sv/gatk:mw-gatk-sv-36c8568` (`mw_gatk_sv@36c8568c3`,
  `sha256:8bc36754c257...`), replacing the pre-change `mw-gatk-sv-53d5c2d`. The
  genotyping work lives on a long-lived gatk fork, so no CI-built gatk image contains
  it; tag convention matches the value it replaces (`<branch>-<sha7>`).
- `sv_pipeline_docker` and `sv_pipeline_qc_docker` ->
  `gatk-sv/sv-pipeline:mw-genotype-scale-9b9809` (`sha256:eb5492bb882a...`). This PR
  edits python that the image bakes (`dockerfiles/sv-pipeline/Dockerfile:33-38` copies
  `src/sv-pipeline`, `src/WGD`, `src/RdTest`, `src/svtk`, `src/svtest`, `src/svqc`),
  and the new metric names above are only consistent if the image and the WDLs move
  together. No rebuild was needed for the review fixes: `git diff 9b98092b HEAD --
  src/sv-pipeline src/svtest src/svtk` is empty, so the image built at `9b98092b` is
  current for every subtree that image carries.
- `sv-shell` -> - `sv-shell` -> `gatk-sv/sv-shell:mw-genotype-scale-8748f0` (`sha256:cfe0a28f50f4d4d02af857bfb891488a5803945b6dd3934868a4181ccc70ad45`): rebuilt for the RD-table/fixture fix, then again to move the **GATK pin inside the image** (see the follow-up at the end). Built at `8748f0bb`; later commits touch only `inputs/values/dockers.json` / `.github/.dockstore.yml`, which are copied into no image.
  `src/sv_shell` is baked only into this image (`dockerfiles/sv-shell/Dockerfile:217`),
  which is why finding 1 needed an image at all; the pin it replaces was built from
  `31d3df3c`, whose `src/sv_shell` is identical to `main`'s, so the delta a rebuild
  carries is exactly this PR's 6 files / +10 -6.
- Deliberately unchanged: `sv-pipeline-virtual-env` (conda/pip env only, no `src/`
  COPY) and every non-genotyping image.
- Note on self-reference: an image cannot embed the commit that names it. Each tag
  above is built from the branch commit with that sha prefix, whose tree differs from
  the final commit only in `inputs/values/dockers.json`, which no image copies.

Verification
- Local CI trio, after the image bump: `build_default_inputs.sh -d .` (no tracked
  input JSON changes; generated JSONs are gitignored),
  `miniwdl_validation.py --imports-dir wdl wdl/*.wdl` (all pass),
  `validate.sh -t -d . -j womtool-84.jar` -> **29/29** Terra input JSONs valid.
  Plus `bash -n` on the edited shell script, JSON parse check on the edited fixture,
  and a confirmation that `GenotypeSVs` really exposes `--num-bins`
  (`AggregateDepthEvidence.NUM_BINS_LONG_NAME = "num-bins"`).
- End-to-end gate on Terra: step 10 (`GenotypeBatch`) ran on the frozen 156-sample
  1000G cohort at the tested commit of this branch (`archive/mw_genotype_scale-pre-review-fixes`,
  content-identical to `f19319fd`) -- Succeeded, 75.8 min wall, 316.5 nested VM-min /
  33 jobs. Its `genotyping_{sr,pe,rd_depth,rd_pesr}_geno_params.tsv`,
  `sr_cutoff_diagnostics` are byte-identical (md5) to the previous full-chain run, and
  both `genotyped_pesr_vcf` / `genotyped_depth_vcf` record bodies are md5-identical
  (object sizes differ by 3 B / 1 B, header only). Genotype concordance therefore
  carries over: PESR 0.9895/0.9905 (exact 0.990600), depth 0.9972/0.9997 (exact
  0.999921), 5,386/5,386 baseline depth sites matched.
  The rerun predates fixes 1-3 above; 1 is outside the GenotypeBatch path, 3 is inert
  at the default bin count that was used, and 2 only adds an assertion -- all three
  re-verified locally as described above.
- Known gap, not introduced here: `GenotypeBatch` has no `GenotypeBatchMetrics` task
  wired (the workflow file exists, unwired), so step 10 emits no metrics file.

**The `Verify` check failure is expected and is to be bypassed by admin merge.** That
job declares `readonly_files=(".github/.dockstore.yml" "inputs/values/dockers.json")`
and exits 1 with `Readonly file modified: inputs/values/dockers.json`, which is the
standard way this repo ships an image bump; everything else is green (Linting, CodeQL,
Determine Build Args).

---

**Follow-up: the `sv-shell` image was shipping a jar that could not run this branch's own scripts.**

Found by asking "could a single-sample user actually run this branch?"

1. The image built `/opt/gatk.jar` from GATK `672d8557d`, where `GenotypeSVs` declares only `--rd-table` — but `src/sv_shell/genotype_svs.sh` on this branch passes `--rd-depth-table`/`--rd-pesr-table`. The branch's SV-shell driver could not have run against its own image. Fixed by pinning production `mw_gatk_sv` = `36c8568c37b382fa58e3e2a057849978b7ff7fae` in `dockerfiles/sv-shell/Dockerfile` and rebuilding; the build log confirms it (`#8 50.05 HEAD is now at 36c8568c3 TrainSVGenotyping batch train…`).
2. The base image was wrong too: the previously pinned `sv-shell` was layered on the **v1.1.1 release** `sv-pipeline` (old python under new shell scripts). The rebuild is `FROM gatk-sv/sv-pipeline:mw-genotype-scale-9b9809` pinned **by digest** (`sha256:eb5492bb…`, 68 references in the log).
3. `.github/.dockstore.yml` now publishes `SingleSamplePipeline` from this branch as well as `main`, which is what makes a single-sample run from this branch possible on Terra. The baseline-vs-branch single-sample comparison is the PR gate.

**A correction I owe, because a reviewer cannot see it from the diff.** I first reported `GATKSVPipelineSingleSample.wdl` on `main` as *unlaunchable* (`miniwdl: No such input rd_table`) and this PR as repairing it. That was wrong, and it was my own methodology bug: I loaded main's copy against a scratch directory that already held this branch's other WDLs. `main` is self-consistent — `rd_table` at `:1211` is an argument of `call genotypebatch.GenotypeSVs`, whose **task** declares `rd_table` on main. With imports materialised at each ref, **both** refs load clean. So this file's change is a **companion change forced by the Java argument split** (the task declares `rd_depth_table`/`rd_pesr_table` here, so callers must follow), not a fix for a pre-existing defect. For anyone re-running this: a WDL launchability gate is worthless unless every import comes from the *same* ref — `git archive <ref> wdl | tar -x`, never a reused scratch dir.

**Two self-inflicted defects, named because they are the class that survives review.** An edit to `.github/.dockstore.yml` silently deleted `- main` from one workflow entry and duplicated a `tags:` key — the same invalid-YAML family that stopped Dockstore publishing all 32 workflows for five days (#962). My first check for it was also worthless: the one-liner interpolated the expected branch name into its own output, so it "passed" regardless of the file. What catches it is parsing the file and printing parsed filters (32 entries, `any entry missing main: []`). Separately, `wdl/SVShell.wdl` is not launchable from a bare input JSON on any ref (~90 required `RunSVShell` task inputs are never bound in-repo; it is driven from a method config), which is why the comparison targets `GATKSVPipelineSingleSample`.

Off-PR verification of the baked `sv_shell` code: `testkit/svshell_jq_plumbing_scan.py` executes all 15 `jq -n` blocks of the driver over a full valued fixture and diffs `null`/empty keys against `origin/main` — 0 new (the 7 pre-existing are inert; `MakeCohortVcf` consumes `-r` = the depth table through `Rscript /opt/RdTest/RdTest.R`, which is the right semantic for `genotyping_rd_depth_table`, and `n_rd_test_bins=100000` in both `genotype_complex_variants.sh:46` and `GenotypeComplexVariants.wdl:88` still matches the tables these commits produce). `testkit/svshell-image-check/` compares the shipped image's `/opt/sv_shell` bytes to the branch (driver md5 `259b0b4c…`, fixture `e18605c1…`) and reruns the scan on the extracted bytes.

---

**Two further commits (`d3d79c83`, `f021e85c`), and what the single-sample A/B actually found.**

Both arms of the `SVShell` single-sample A/B (`main` image vs this branch's) ran ~44 h and failed
identically — same call, same attempt, exit 2 — in module 04:

```
add_retro_del_filters.py: error: the following arguments are required: --contig
```

`76bc5526` (#953) made `--contig` required and updated the WDL caller (`CleanVcfChromosome.wdl:802`)
but not `src/sv_shell/clean_vcf.sh`, which predates it. **This is not a regression from this PR — the
caller is byte-identical on `main`** — but it means no `sv_shell` single-sample run can complete on
any input on either ref, and nothing could have noticed because cohort/WDL mode was already correct.
Same family as #965: a command-string-vs-caller mismatch that miniwdl, womtool and `validate.sh` all
cannot see. Carrying the fix here is a deliberate call by the PR owner, not this module's author.

The fix is `--all-contigs`, not "call `--contig` once per contig": `find_matching_introns` compares
coordinates only, so looping a per-contig call over a whole-genome VCF re-applies each contig's
introns to every other contig's deletions — exactly the false-positive class #953 exists to remove.
That was verified rather than assumed (`testkit/simplify-harness/retro_del_contig_parity.py`):
`--all-contigs` reproduces the WDL's per-contig scatter record-for-record; the expected tagging
pattern is asserted so the two are proven correct and not merely equal; and **both** negative
controls — the pre-#953 loader (`76bc5526^`) and the whole-VCF per-contig loop — do mis-tag, so the
test would catch a regression. The `--contig` path is untouched, and the shared-loader refactor
returns identical per-contig lists on a randomized reference with reversed and malformed rows.

`d3d79c83` collapses the four duplicated `rf_cutoffs` awk programs in `TrainSVGenotyping` into one
pass. It is the code the rerun proved byte-identical, so it was not touched on inspection: the
frozen-baseline harness (`awk_extract_parity.py`) compares old vs new over 131 inputs (3 real cutoffs
files, 120 randomized, 8 targeted at the two selection semantics that must survive — first-match for
PEQ/SRQ, and `sort -nr | head -n 1` whose tie-break is GNU sort's whole-line byte comparison), all
131 agree, and on the batch that produced the Track B parameters both yield exactly
`PEQ=21.709999084472656 SRQ=47.68000030517578 PESR_SEP=0.096000000834465 DEPTH_SEP=0.3810000121593475`.
Repo gates after both commits: miniwdl over all `wdl/*.wdl`, `build_default_inputs.sh` (81 JSONs),
`validate.sh -t` **29/29**, `flake8` clean.

**What is tested, and what is not — so the merge decision rests on evidence rather than inference.**

*Tested, on the path production uses (cohort mode, which is what AoU runs):* GenotypeBatch on the
frozen 156-sample 1000G cohort against the `v1.1.1` baseline of the same samples — genotype
concordance **0.9895 / 0.9905** PESR and **0.9972 / 0.9997** depth (all 5,386/5,386 baseline depth
sites matched, both-called agreement ≈0.9999); every trained parameter table and
`sr_cutoff_diagnostics.txt` **byte-identical** between the branch run and a rerun against these
images; drift is non-ref-inflating (`homref→het` 0.37%, `homalt→het` 0.45% at pair level); step-10
cost 316 VM-min vs 1673 on the old pipeline.

*Tested but incomplete:* the `sv_shell` single-sample path now clears the genotyping stage this PR
changes — arm B logged `Successfully finished genotype svs` on the rebuilt image, which is the
evidence that the shipped jar accepts `--rd-depth-table`/`--rd-pesr-table` — and both arms completed
evidence gathering, clustering, filtering, merge-depth, complex-variant resolution and `stripy`. What
is **not** obtained is the end-to-end *output* comparison (final VCF/metrics/QC), because neither arm
reached the end. No divergence was observed across the shared ~44 h, but that is not the same as a
pass on final outputs. As of 2026-09-23 the chain is **running again** at this tip on images that carry the `--contig` fix described above, so that final-output comparison is pending rather than unavailable - it is not claimed here until it lands.

*Not tested, and what changed on 2026-09-23:* the branch was **rebased onto PR #966** (`mw_fix_single_sample_blocking`,
`4419315c`), which fixes #965's `ConcatBaf` by running `PrintSVEvidence` from the GATK jar the `sv-base`
chain actually bakes (`java -jar ${GATK_JAR}`) and adding the `--sequence-dictionary` the call needs — BAF
evidence has no header dictionary, and the missing launcher's exit 127 was hiding that second defect. So the
`GATKSVPipelineSingleSample` path on **this** branch is no longer blocked at that step and a single-sample run
of the branch can in principle reach the end. The **comparison** still cannot: the `main` arm is still blocked
at `ConcatBaf` until #966 merges, so there is still no paired branch-vs-main single-sample result. Also not
A branch-only `GATKSVPipelineSingleSample` run at this tip is likewise in flight as of 2026-09-23. Still not tested: the AoU 516-sample canary, which needs access-controlled data.

*Known gaps this PR does not close (flagged, not hidden):* `GenotypeBatch.wdl` emits **no genotype
metrics file** (`GenotypeBatchMetrics` exists, unwired) and `TrainSVGenotyping` logs nothing, so PE het
statistics are unobtainable from a run; `GenotypeSVs` changes VCF representation relative to v1.1.1
(`INS` with `END==POS`, resolved `REF`, breakend `BND`s) with no in-pipeline consumer reading it.

*Consequence of `f021e85c` (updated 2026-09-23 - now acted on):* it edits two baked subtrees
(`src/sv-pipeline/…` -> `sv_pipeline_docker`, `src/sv_shell/…` -> `sv-shell`), so both pins had to move
before any run could exercise the fix. Images for this tip now exist in the TEST folder
`us.gcr.io/broad-dsde-methods/markw/gatk-sv`: `sv-pipeline:mw-genotype-scale-cac483`
(`sha256:057d386963b48a…`) and `sv-shell:mw-genotype-scale-cac483` (`sha256:0e54b1e5df5a0c…`).
Verification here is **content, not build exit status**: inside each image the shipped
`sv_shell` / `sv-pipeline` files md5-match this branch; `sv-shell` carries the **new** `sv-pipeline`
layer, proven by md5 of `add_retro_del_filters.py` *inside* the `sv-shell` image - a stale `FROM` pin is
exactly how `sv-shell:2026-09-09-v1.1.1-31d3df3c` once passed for new content here; `import svtk`
succeeds in both images, so PR #966's `pkg_resources` removal is live at runtime; and the baked jar
exposes `--rd-depth-table` / `--rd-pesr-table` with `--rd-table` gone. `inputs/values/dockers.json` is
still deliberately **not** bumped: the validation runs point at those digests through method-config
overrides, so the branch keeps a single pin set and the merge does not introduce a personal image
folder into `inputs/values/`. At merge, upstream CI builds release images for the merged tree and the
pins move to those.

---

**Owner acceptance of the measured behavior change (2026-09-23).** The behavioral deltas this PR
produces against the `v1.1.1` baseline were measured and put to the pipeline owner rather than
smoothed over: ~1% of genotype classes shifting (net `het→hom`, and **non-ref-inflating** at pair
level — `homref→het` 0.37%, `homalt→het` 0.45%), slightly more conservative no-calls at the depth
boundary (0.26% of observations), `pe_count` 4 vs 8 decided by a 0.022% margin on step-07 `PEQ`
(21.71 vs 34.74), and the inverted SR cutoff strategy (strict-single/loose-both where v1.1 was
loose-single/strict-both), with the residual `sd_het` MAD 9-vs-11/10 left explicitly unexplained.
**The owner has accepted them.** This is recorded here because a reviewer cannot tell
"measured, then accepted" apart from "unmeasured, unreviewed", and that distinction is the whole
content of this paragraph. What remained open was not a behavior question, and both items are now answered: #965's `ConcatBaf` defect is fixed by PR #966 (which this branch is stacked on) and filed for the record, and the two `FilterBatchSites` items were closed on 2026-09-23 **with no code change** - the CTX asymmetry is moot because zero `CTX` records exist in any step-07 shard feeding the task in either pipeline, and no malformed-line guard was requested.


mwalker174:
### Copilot findings: verified rather than assumed

Every finding was checked against the tree at the reviewed head, against the Java sources, and wherever
testable against real data or executed code. All five were substantively correct; three carried sub-claims
that did not survive checking. Four are fixed in `0750a0cc`, one is documented instead of patched because
its measured impact is zero.

| # | finding | verdict | action |
|---|---|---|---|
| 1 | RD-table keys not propagated after the `genotyping_rd_table` split | **real — a regression introduced here** | fixed: `SVShell` inputs/`--arg`s, the `MakeCohortVcf` block in `single_sample_pipeline.sh`, DRAGEN fixture |
| 2 | `ValidateSRCutoffs` passes an empty diagnostics file | **real** | fixed: both `*_selection_status` records must be *present*; fix tested by execution (below) |
| 3 | `n_RD_genotype_bins` reaches the fit but not the apply | accurate but **latent** | fixed: task input + `--num-bins` + call-site binding |
| 4 | CTX records collected in the raw-text pass but never rewritten | real in code, **0 measured occurrences** | PR wording corrected (it over-claimed); code left alone, see note below |
| 5 | description contradicts the `.dockstore.yml` diff | **correct** | description rewritten |

Notes that matter more than the table:

**1 was worse than "a null table path".** `jq -r` on a missing key yields the string `null`, so the two
callers broke in opposite directions: through `SVShell` (which supplied only the old key) `genotype_svs.sh`
would have run `GenotypeSVs --rd-depth-table null`; through the shipped `sample_inputs/` fixture (which
supplied only the new pair) `MakeCohortVcf` got `genotyping_rd_tables: [null]`. The depth table is the
correct value for that consumer — `GATKSVPipelineSingleSample.wdl:1269` already does
`genotyping_rd_tables=[genotyping_rd_depth_table]`. `SVShell` is not dead code in the sense that matters:
no in-tree WDL calls it, but it is published on Dockstore for `main`, and its pinned image's `src/sv_shell`
is identical to `main`'s, so this path worked before and would have stopped working.

**2 was tested by running the gate, not by reading it.** The `ValidateSRCutoffs` command body was extracted
from the WDL and executed over six fixtures: healthy real diagnostics (passes), empty (blocks), empty with
`fail_on_degenerate_sr_cutoffs=false` (warns, continues), explicit `REJECTED_DEGENERATE_GRID` (blocks,
unchanged), one status record absent (blocks), truncated write (blocks). A healthy file carries exactly two
status rows, so "require both" is precise rather than heuristic.

**3 is exactly the defect this PR set out to fix, one layer down.** Both tools feed `--num-bins` into an
identical `DepthMatrixLoader(depthSource, numBins, …)` call (`TrainSVGenotyping.java:702`,
`GenotypeSVs.java:354`); only the defaults differ — `TrainSVGenotyping` 10, `GenotypeSVs` 100000. The WDL
passed the value to the fit only, so the defaults happened to agree with apply at 100000 and today's results
are unaffected, but any override desynced fit from apply. v1.1.1 passed `-i` to both.

**4 is unexercised, not wrong.** Measured over the step-07 shards of both pipelines: manta 54,678 records
(8,556 BND), scramble 14,684 (0 BND), wham 9,105 (1,037 BND) — **CTX = 0** in every shard, and END2/CHR2
present on 100% of BNDs. Lifting the condition to cover CTX is a one-line change if a reviewer prefers
symmetry over the narrower legacy shim (it carries a removal TODO already); I chose not to change code that
nothing exercises.

**5 also deserved the correction.** The filters are in the diff on purpose: these entries are
`mode: DOCKSTORE_YML` and Dockstore publishes a branch ref only if a filter matches, which is what let this
branch's `GenotypeBatch` be referenced by the Terra rerun described in the description.

Where the review over-reached, so nobody re-adopts its framing whole: the stale read is at
`single_sample_pipeline.sh:880`, not `:799`; `single_sample_pipeline_dragen.json` is referenced by nothing
in-tree (shipped fixture only); "therefore re-creates the mismatch" (#3) holds only under an override; and
the CTX finding implies corrupted data where the measured count is zero.

**Two errors of mine surfaced while checking its proposed fixes** — worth stating plainly:

- I first mapped `src/sv_shell` to the `sv-pipeline` image. It belongs only to `sv-shell`
  (`dockerfiles/sv-shell/Dockerfile:217`), while `sv-pipeline` bakes `src/sv-pipeline`, `WGD`, `RdTest`,
  `svtk`, `svtest`, `svqc` (`:33-38`). So `sv-pipeline` needed no rebuild at all (`git diff 9b98092b HEAD --
  src/sv-pipeline src/svtest src/svtk` is empty) and `sv-shell` did.
- The PR bumped only `gatk_docker`, but it edits python that `sv_pipeline_docker` bakes — the metric-name
  alignment in `metrics_file.py`, 02e `aggregate.py`, 03 `build_training_set.py`, and `adjudicate_sv.py`.
  Merging it with the old pin would have paired new WDLs with v1.1.1-era python, i.e. the same class of
  train/apply skew this PR exists to remove. `sv_pipeline_docker`, `sv_pipeline_qc_docker` and `sv-shell`
  now move with it; `sv-pipeline-virtual-env` does not, because its Dockerfile copies no `src/`.

Re-verified after the changes: miniwdl over all `wdl/*.wdl` passes, `validate.sh -t` is 29/29, `bash -n`
clean, and the edited JSON parses. The Terra end-to-end proof still applies: fixes 1 is outside the
GenotypeBatch path, 3 is inert at the default bin count that run used, and 2 only adds an assertion.

mwalker174:
### Rebased onto #966 (2026-09-23): `f021e85c` → `cac4830c` (force-push)

Stacked this PR on `mw_fix_single_sample_blocking` (`4419315c`), because #965's `ConcatBaf` defect was the
thing preventing this branch's single-sample path from reaching genotyping at all — both arms of my
`SVShell` A/B and both arms of the `GATKSVPipelineSingleSample` A/B died at it. #966 fixes it properly
(`java -jar ${GATK_JAR}` — the jar the `sv-base` chain actually bakes — plus the required
`--sequence-dictionary`), so this branch now carries that fix.

**One conflict, mechanical.** `.github/.dockstore.yml`: both branches add their own branch name to the same
`filters.branches` lists, 5 hunks, resolved as the union. Verified with a YAML parser rather than by eye
(the silent-Dockstore-outage failure mode is well documented here): parses clean, **32** entries, **all 32
still contain `main`**, `mw_genotype_scale` in 7, `mw_fix_single_sample_blocking` in 5, no entry missing `main`.

**Both sides verified present after the rebase**, not assumed: theirs — `java -Xmx2g -jar ${GATK_JAR}`
`PrintSVEvidence`, `--sequence-dictionary`, `pkg_resources` gone from `svtk` except the docstring that
explains the replacement (identical to `4419315c`'s tree); ours — split `rd_depth_table`/`rd_pesr_table`
call site, `--all-contigs` in `clean_vcf.sh`, the consolidated `QUERIES=$(awk` block, `training_vcf`.

**Gates on the rebased tree:** miniwdl over all `wdl/*.wdl` pass · `flake8` clean ·
`build_default_inputs.sh` rebuilt with no tracked-input drift · `validate.sh -t` **29/29**.

**Review anchors.** The old tip is preserved as the pushed tag **`archive/mw_genotype_scale-pre-966-rebase`**
(`f021e85c`). The four commits are the same work; only the `.dockstore.yml` hunk differs, but GitHub may show
inline comments on the pre-rebase commits as outdated.

**What reviewers now see, and one hazard.** The PR reads as 8 commits / 39 files because #966's 4 commits are
not yet in `main`; it shrinks to 4 commits / 33 files the moment #966 merges. The hazard is the mirror image:
**if #966 is ever closed unmerged, this PR would carry its commits into `main`.** Say so and I will rebase
back onto `main` — the operation is trivial either way.

**Consequence for testing.** The branch is no longer blocked at `ConcatBaf`, but the *paired* comparison still
isn't available: the `main` arm remains blocked there until #966 merges. And a branch single-sample run still
needs an `sv-pipeline` image rebuilt at this tip — the currently pinned `mw-genotype-scale-9b9809` predates
#966, so it does not contain the `svtk`/`pkg_resources` fix, which is blocker #2 of that PR (*reasoned from
the build lineage recorded in the pin, not measured against the image*). Dockstore must also republish the
branch before a Terra run can see any of this, and I will verify the published descriptors rather than assume.

